PR #3, opened 2017-10-11
[MIG]11.0 password security

smurf-U:
Migrate Password Security
Improve code and functionality for set password Policy: Boolean Field => Number Field for check specific number of Uppercase, Lowercase, Special, Integer characters in Password


pedrobaeza:
Thanks! You're doing great. Last step is to squash all "OCA Transbot..." commits together following this guide: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

password_security/README.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Known Issues / Roadmap
+======================
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/LasLabs/odoo-base/issues>`_. In case of trouble, please
```
yajo: Wrong repo

password_security/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
```
yajo: Please remove this line on every python file. v11 is now Python 3, which only works with UTF-8 files, so there's no more need to explicitly specify it.

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
-        if company_id.password_special:
-            password_regex.append(r'(?=.*?\W)')
-        password_regex.append('.{%d,}$' % company_id.password_length)
-        if not re.search(''.join(password_regex), password):
-            raise PassError(self.password_match_message())
-        return True
+        message = []
+        if company_id.password_lower and sum(map(str.islower, password)) < \
+                company_id.password_lower:
+            message.append('\n ' + _('Lowercase letter (At least ' +
```
yajo: No need for `\n`, it's joined at the end with it.

Wrong usage of `_()`: https://www.odoo.com/documentation/11.0/reference/translations.html#explicit-exports

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
+                                     str(company_id.password_lower) +
+                                     ' character)')
+                           )
+        if company_id.password_upper and sum(map(str.isupper, password)) < \
+                company_id.password_upper:
+            message.append('\n ' + _('Uppercase letter (At least ' +
```
yajo: Wrong usage of `_()`

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
-        if company_id.password_numeric:
-            password_regex.append(r'(?=.*?\d)')
-        if company_id.password_special:
-            password_regex.append(r'(?=.*?\W)')
-        password_regex.append('.{%d,}$' % company_id.password_length)
-        if not re.search(''.join(password_regex), password):
-            raise PassError(self.password_match_message())
-        return True
+        message = []
+        if company_id.password_lower and sum(map(str.islower, password)) < \
```
yajo: Simplify this condition with `if sum(map(str.islower, password)) < company_id.password_lower`. Empty Integer fields are always `0`.

Also, although summing `True` values is perfectly possible, it's not very straightforward to understand. I'd change the method to `len(filter(str.islower, password))`. This is just a matter of taste, not really a requirement 😉

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
+                company_id.password_lower:
+            message.append('\n ' + _('Lowercase letter (At least ' +
+                                     str(company_id.password_lower) +
+                                     ' character)')
+                           )
+        if company_id.password_upper and sum(map(str.isupper, password)) < \
```
yajo: Simplify

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
+                company_id.password_upper:
+            message.append('\n ' + _('Uppercase letter (At least ' +
+                                     str(company_id.password_upper) +
+                                     ' character)')
+                           )
+        if company_id.password_numeric and sum(map(str.isdigit, password)) < \
```
yajo: Simplify

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
+                company_id.password_numeric:
+            message.append('\n ' + _('Numeric digit (At least ' +
```
yajo: Wrong usage of `_()`

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
+                                     str(company_id.password_numeric) +
+                                     ' numeric)')
+                           )
+        if company_id.password_special and len(set('[~!@#$%^&*()_+{}":;\']+$'
```
yajo: This does not include all special characters. For instance these are missing: `^\/|·,.-`

Simplify and improve with: `if len(re.findall(r"\W", password)) < company_id.password_special`

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
+                                                   ).intersection(
+                password)) < company_id.password_numeric:
+            message.append('\n ' + _('Special character (At least ' +
+                                     str(company_id.password_special) +
+                                     ' character of [ ~ ! @ # $ % ^ & * ( )_+ '
+                                     '{ } " : ; \' ])')
+                           )
+        if company_id.password_length and len(password) < \
```
yajo: Simplify

yajo (changes requested):

smurf-U:
@Yajo all change request are done

pedrobaeza:
Theres' another PR in green in #4 

Should we switch to that one?

smurf-U:
@pedrobaeza  I just want to say that some important improvements in functionality were done by me which is useful to other in this PR.

pedrobaeza:
Then please fix Travis.

password_security/models/res_company.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2015 LasLabs Inc.
+# Copyright 2017 Kaushal Prajapati <kbprajapati@live.com>.
```
lasley: Why did you remove my copyright?

password_security/models/res_company.py
```diff
@@ -18,25 +18,21 @@ class ResCompany(models.Model):
         default=12,
         help='Minimum number of characters',
     )
-    password_lower = fields.Boolean(
+    password_lower = fields.Integer(
         'Lowercase',
-        default=True,
         help='Require lowercase letters',
```
lasley: These help attributes are now incorrect

password_security/models/res_users.py
```diff
@@ -1,9 +1,7 @@
 # -*- coding: utf-8 -*-
-# Copyright 2015 LasLabs Inc.
+# Copyright 2017 Kaushal Prajapati <kbprajapati@live.com>.
```
lasley: I still copyright this file as well. You can only copyright your changes.

password_security/models/res_users.py
```diff
@@ -76,19 +52,41 @@ def _check_password_rules(self, password):
...
-            password_regex.append('(?=.*?[A-Z])')
-        if company_id.password_numeric:
-            password_regex.append(r'(?=.*?\d)')
-        if company_id.password_special:
-            password_regex.append(r'(?=.*?\W)')
-        password_regex.append('.{%d,}$' % company_id.password_length)
-        if not re.search(''.join(password_regex), password):
-            raise PassError(self.password_match_message())
-        return True
+        message = []
```
lasley: IMO this code just gained complexity and reduced efficiency. Why can we not continue to use regex for everything? This is a quick regex I wrote illustrating a check for 1 uppercase, 2 lowercase, 3 digits, and 4 special characters

```
/^(?=(.*[^A-Za-z0-9]){4,})(?=(.*[A-Z]){1,})(?=(.*\d){3,})(?=.*[a-z]){2,}).+/
```

Otherwise, we iterate the string 3 times to perform evaluations at each letter, just to regex it anyways 😦

smurf-U:
@pedrobaeza yup work in progress

lasley (changes requested):
Thanks @smurf-U

lasley:
I commented on an outdated diff I guess, but I'm pretty sure my comments still apply.

password_security/models/res_company.py
smurf-U: @lasley sorry remove by mistake

password_security/models/res_users.py
nhomar: +1 More regex, regex, regex.

nhomar (dismissed):
I like what I see but I think the comments from lasley are valid yet, can you check and requests the proper changes? NICE JOB.

password_security/models/res_users.py
yajo: I think @smurf-U wanted to get more granular error messages, which I think is a good thing because, although you're right and possibly performance gets hit, this is really not a performance-critical part of Odoo: you're not changing passwords all the time after all.

So I'm OK on sacrificing performance for boosting UX this time.

It's also true that individual checks could use individual regexps OTOH.

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+        if len(list(filter(str.islower, password))
```
yajo: If you remove the `list()` part, you save a bunch of loops.

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+               ) < company_id.password_lower:
+            message.append('\n ' + _('Lowercase letter (At least ' +
+                                     str(company_id.password_lower) +
+                                     ' character)')
+                           )
+        if len(list(filter(str.isupper, password))
```
yajo: No `list()`

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+               ) < company_id.password_upper:
+            message.append('\n ' + 'Uppercase letter (At least ' +
+                           str(company_id.password_upper) + ' character)')
+
+        if len(list(filter(str.isdigit, password))) < \
```
yajo: No `list()`

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+        if len(re.findall(r"\W", password)) < company_id.password_special:
+            message.append('\n ' + 'Special character (At least ' +
+                           str(company_id.password_special) +
+                           ' special character')
+
+        if len(password) < company_id.password_length:
+            message = ['Password must be %d characters or more.' %
+                       company_id.password_length] + message
+
+        if len(message) > 0:
```
yajo: `if message`

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+            raise PassError('\r'.join(message))
```
yajo: I think you should join it with `\n` instead, just try please.

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+        message = []
+        if len(list(filter(str.islower, password))
+               ) < company_id.password_lower:
+            message.append('\n ' + _('Lowercase letter (At least ' +
+                                     str(company_id.password_lower) +
+                                     ' character)')
+                           )
+        if len(list(filter(str.isupper, password))
+               ) < company_id.password_upper:
+            message.append('\n ' + 'Uppercase letter (At least ' +
```
yajo: No need for `\n`, it's joined at the end with it.

Wrong usage of `_()`: https://www.odoo.com/documentation/11.0/reference/translations.html#explicit-exports

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+                           str(company_id.password_upper) + ' character)')
+
+        if len(list(filter(str.isdigit, password))) < \
+                company_id.password_numeric:
+            message.append('\n ' + 'Numeric digit (At least ' +
```
yajo: No need for `\n`, it's joined at the end with it.

Wrong usage of `_()`: https://www.odoo.com/documentation/11.0/reference/translations.html#explicit-exports

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+                           str(company_id.password_numeric) + ' numeric)')
+
+        if len(re.findall(r"\W", password)) < company_id.password_special:
+            message.append('\n ' + 'Special character (At least ' +
```
yajo: No need for `\n`, it's joined at the end with it.

Wrong usage of `_()`: https://www.odoo.com/documentation/11.0/reference/translations.html#explicit-exports

password_security/models/res_users.py
```diff
@@ -76,19 +55,36 @@ def _check_password_rules(self, password):
...
+                           str(company_id.password_special) +
+                           ' special character')
+
+        if len(password) < company_id.password_length:
+            message = ['Password must be %d characters or more.' %
```
yajo: Why don't you put this check at the top and just use the `.append()` system as everywhere. Just to keep it consistent.

Besides, you're missing the `_()` part here.

yajo (changes requested):

password_security/models/res_users.py
smurf-U: >len(filter(str.islower, password)) 

is not work in python3

password_security/models/res_users.py
smurf-U: >len(filter(str.isupper, password))

is not work in python3

password_security/models/res_users.py
smurf-U: >len(filter(str.isdigit, password))

is not work in python3

password_security/models/res_users.py
lasley: The flip side of this implementation is that there is now no way to display all of the password requirements at once. We directly put the output of the `password_match_message` in our template. IMO being able to communicate all of the rules at once is more valuable than communicating the one that failed.

password_security/models/res_users.py
```diff
@@ -55,36 +81,23 @@ def _check_password_rules(self, password):
...
-        if len(password) < company_id.password_length:
-            message = ['Password must be %d characters or more.' %
-                       company_id.password_length] + message
-
-        if len(message) > 0:
-            raise PassError('\r'.join(message))
-        else:
-            return True
+        password_regex = ['^']
+        if company_id.password_lower:
```
lasley: Interestingly enough, I just realized you can outright remove all of these conditions. Now that we're using numerics, an empty option is simply checking for zero instances of the condition:

```
(?=.*?[a-z]){0,}
```

password_security/models/res_users.py
```diff
@@ -55,36 +81,23 @@ def _check_password_rules(self, password):
...
+            password_regex.append('(?=.*?[a-z]){' +
+                                  str(company_id.password_lower) + ',}')
```
lasley: `'(?=.*?[a-z]){%d,}' % int(company_id.password_lower)`

Because:
```
>>> int(False)
0
```

password_security/models/res_users.py
pedrobaeza: The ideal is to put the full message but highlighting any way (for example, with an ending `**`) the infringement.

lasley: So something I was playing around with to make this work, but also not increase complexity a bunch, was to use named groups. Each character type (chars, upper, lower, numeric) would be a named group, then you simply take the length of each group and compare against what you want.

We ended up deciding meh (when originally creating). I think this is the best way to go about something like that if it were to be implemented though.

password_security/README.rst
```diff
@@ -65,7 +65,7 @@ Bug Tracker
 ===========
 
 Bugs are tracked on `GitHub Issues
-<https://github.com/LasLabs/odoo-base/issues>`_. In case of trouble, please
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
```
lasley: Excellent catch! We're in `server-auth` now though 😉

password_security/models/res_users.py
```diff
@@ -52,41 +80,17 @@ def _check_password_rules(self, password):
...
-                company_id.password_length:
-            message = [_('Password must be %d characters or more.') %
-                       company_id.password_length] + message
-        if len(message) > 0:
-            raise PassError('\r'.join(message))
-        else:
-            return True
+        password_regex = [
+            '^',
+            '(?=.*?[a-z]){' + str(company_id.password_lower) + ',}',
```
lasley: You're going to want to use `int()` then just interpolate. Reasoning:

```
>>> str(False)
'False'
>>> int(False)
0
```

But:

```
>>> 'test' + int(1)
  File "<stdin>", line 1, in <module>
TypeError: cannot concatenate 'str' and 'int' objects
```

So we're looking for:

```
'(?=.*?[a-z]){%d,}' % int(company_id.password_lower),
```

yajo: Actually you can strip the `int()`:

```python
>>> "%d" % False
'0'
```

password_security/models/res_users.py
```diff
@@ -41,6 +44,31 @@ def write(self, vals):
...
+                company_id.password_lower) + ' character)'))
+        if company_id.password_upper:
+            message.append('\n* ' + _('Uppercase letter (At least ' + str(
+                company_id.password_upper) + ' character)'))
+        if company_id.password_numeric:
+            message.append('\n* ' + _('Numeric digit (At least ' + str(
+                company_id.password_numeric) + ' character)'))
+        if company_id.password_special:
+            message.append('\n* ' + _('Special character (At least ' + str(
+                company_id.password_special) + ' character)'))
```
yajo: You can't use `_()` like this. Check the warning in the docs: https://www.odoo.com/documentation/11.0/reference/translations.html#explicit-exports

yajo (changes requested):

password_security/models/res_users.py
lasley: Oh cool TIL- I didn't know that the `%d` would typecast non-ints. I thought it threw an error like it will with a string:

```
>>> '%d' % '1'
  File "<stdin>", line 1, in <module>
TypeError: %d format: a number is required, not str

```

yajo: TIL it fails with str 😆

lasley: Yeah the `%d` is meant to serve as a validation as well, so it fails with things that don't have a direct int val. I guess `False` does technically have a direct int val though, because 0 or 1

hbrunn: `bool` actually inherits from `int`:

```
>>> issubclass(bool, int)
True
```

lasley: And that's a TIL for me! Today is a good day for learning 😄

StefanRijnhart:
Can you please cherrypick https://github.com/OCA/server-tools/pull/1077?

StefanRijnhart:
Thanks for taking my change! Travis fails because the commit you cherrypicked from my branch needs to be adapted to your refactoring. For one thing, when changing the boolean configuration fields to integers you set the default to 0. I think they should be set to 1 to keep the same default security level. You can then adapt my test to verify the `1` value instead of the `True` value for requiring special characters which it currently tests. And as Travis notes, you need to prefix the \W string as `r'`

lasley:
We just had another bug fix come in on this module. Gotta love security, it's the gift that keeps on giving! https://github.com/OCA/server-tools/pull/1084

lasley (approved):
Thanks @smurf-U

smurf-U:
Welcome @lasley

StefanRijnhart (approved):
Thanks!

lasley:
@nhomar & @Yajo - have your comments been attended to?

yajo (approved):

PR #4, opened 2017-10-29
[MIG] password_security to 11

mart-e:
Move to server-auth
Bump version number
Remove coding utf-8

_As it is the [hacktoberfest](https://hacktoberfest.digitalocean.com/), it is the occasion to help outside projects than Odoo_

Note: while testing this module, I noticed the error message during signup to be badly formed (`('Password must be 12 characters or more.\rMust contain the following:\r\n* Lowercase letter\r\n* Uppercase letter\r\n* Numeric digit\r\n* Special character', '')`) but it looks this is actually a bug of `auth_signup` that `UserError` are not properly displayed in the signup form.
Will check on core how to fix that so you may want to wait that this bug is fixed before merging this one.


pedrobaeza:
@mart-e how good is to see you contributing to OCA! Thanks for the PR.

I see that you have realized about the split of server-tools as it was very over-crowded. The only problem is about commit history. Please see migration guide https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0, but in this case you will need to add as remote server-tools and change slightly the `git am` command for catching the commits from this repo instead of origin, but nothing complicated for a git-fu like you :wink:

lasley (approved):
Thanks @mart-e

pedrobaeza:
There's already another PR for the same module in #3. This one passes all the CIs and the other not, but we should agree what to do, as first PR usually has priority.

mart-e:
Oh I missed that one. Sure, if he fixes all the issues, it should have the priority obviously.

mart-e:
FYI, the issue in the rendering of the error message was fixed upstream at odoo/odoo@c3bf1618bd9

pedrobaeza:
Thanks for that, Martin!

yajo (approved):
code review

yajo:
Oops, I think we should close this in favor of #3, no?

PR #7, opened 2018-01-03
[MIG] 11.0 mig auth from http remote user

TDu:


auth_from_http_remote_user/README.rst
```diff
@@ -0,0 +1,174 @@
...
+====================================
+Authentication From HTTP Remote User
+====================================
+
+This module initialize the session by looking for the field HTTP_REMOTE_USER in
+the HEADER of the HTTP request and trying to bind the given value to a user.
+To be active, the module must be installed in the expected databases and loaded
+at startup; Add the *--load* parameter to the startup command: ::
+
+  --load=web,web_kanban,auth_from_http_remote_user, ...
```
guewen: in v11 `web_kanban` is gone

auth_from_http_remote_user/controllers/main.py
```diff
@@ -0,0 +1,110 @@
...
+                request.params['login_success'] = True
+                # generate a specific key for authentication
+                key = randomString(utils.KEY_LENGTH, '0123456789abcdef')
+                cr.execute('''update res_users
+                                set sso_key=%s
+                                where id=%s''', (key, user_id.id))
+            request.session.authenticate(db_name, login=login,
+                                         password=key, uid=user_id.id)
+        except http.AuthenticationError as e:
+            raise e
```
guewen: The correct way to reraise an expection is only `raise` otherwise you lose where you are in the traceback.
In this case, I don't see why the handler is there.

auth_from_http_remote_user/controllers/main.py
```diff
@@ -0,0 +1,110 @@
...
+        except Exception as e:
+            _logger.error("Error binding Http Remote User session",
+                          exc_info=True)
+            raise e
```
guewen: same note about reraising

auth_from_http_remote_user/models/res_users.py
```diff
@@ -0,0 +1,76 @@
...
+class Users(models.Model):
+    _inherit = 'res.users'
+
+    sso_key = fields.Char('SSO Key', size=utils.KEY_LENGTH,
+                          readonly=True)
+
+    def copy(self, default=None):
+        default = default = {} if default is None else default
+        default['sso_key'] = False
+        return super(Users, self).copy(default)
```
guewen: The override of `copy` should be replaced by the `copy=False` argument in the definition of the `sso_key` field.

auth_from_http_remote_user/models/res_users.py
```diff
@@ -0,0 +1,76 @@
...
+                                FROM res_users
+                               WHERE id=%s
+                                 AND sso_key=%s
+                                 AND active=%s''', (int(uid), passwd, True))
+                if not cr.fetchone()[0]:
+                    raise
+                cls._uid_cache.setdefault(db, {})[uid] = passwd
+
+    @classmethod
+    def find_uid_from_iamuserid(cls, db='', login=''):
```
guewen: I don't think this one ought to be there?

auth_from_http_remote_user/controllers/main.py
```diff
@@ -0,0 +1,110 @@
...
+                    # HTTP_REMOTE_USER login not found in database
+                    request.session.logout(keep_db=True)
+                    raise http.AuthenticationError()
+
+                request.params['login_success'] = True
+                # generate a specific key for authentication
+                key = randomString(utils.KEY_LENGTH, '0123456789abcdef')
+                cr.execute('''update res_users
+                                set sso_key=%s
+                                where id=%s''', (key, user_id.id))
```
guewen: I guess there is a reason to bypass the ORM but could you put a comment explaining why?

guewen (changes requested):
Thanks! A few suggestions below

auth_from_http_remote_user/README.rst
```diff
@@ -0,0 +1,174 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+    :target: https://www.gnu.org/licenses/agpl
+    :alt: License: AGPL-3
+
+====================================
+Authentication From HTTP Remote User
+====================================
+
+This module initialize the session by looking for the field HTTP_REMOTE_USER in
+the HEADER of the HTTP request and trying to bind the given value to a user.
```
simahawk: s/HEADER/HEADERS

auth_from_http_remote_user/README.rst
```diff
@@ -0,0 +1,174 @@
...
+.. [#] Based on a ubuntu 12.04 env
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+    :alt: Try me on Runbot
+    :target: https://runbot.odoo-community.org/runbot/149/11.0
+
+Known Issues / Roadmap
```
simahawk: remove this section if empty

auth_from_http_remote_user/__manifest__.py
```diff
@@ -0,0 +1,36 @@
+##############################################################################
```
simahawk: replace extended headers w/ short one

auth_from_http_remote_user/__manifest__.py
```diff
@@ -0,0 +1,36 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Authenticate via HTTP Remote User',
+    'version': '11.0.1.0.0',
+    'category': 'Tools',
+    'author': "Acsone SA/NV,Odoo Community Association (OCA)",
+    'maintainer': 'ACSONE SA/NV',
+    'website': 'http://www.acsone.eu',
```
simahawk: https://github.com/OCA/server-auth

auth_from_http_remote_user/__manifest__.py
```diff
@@ -0,0 +1,36 @@
...
+    'depends': ['base', 'web', 'base_setup'],
+    "license": "AGPL-3",
+    'data': [],
```
simahawk: remove empty keys

auth_from_http_remote_user/models/res_users.py
```diff
@@ -0,0 +1,76 @@
...
+        default['sso_key'] = False
+        return super(Users, self).copy(default)
+
+    def check_credentials(self, password):
+        res = self.sudo().search([('id', '=', self._uid),
+                                  ('sso_key', '=', password)])
+        if not res:
+            try:
+                return super(Users, self).check_credentials(password)
+            except odoo.exceptions.AccessDenied:
```
simahawk: same as noted above by Guewen: maybe I'm wrong but I don't see a good reason to catch the exception and raise it right after.

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -0,0 +1,93 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.tests import common
```
simahawk: s/openerp/odoo

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -0,0 +1,93 @@
...
+        finally:
+            cr.close = org_close
+    cr.autocommit = org_autocommit
+
+
+class TestResUsers(common.TransactionCase):
+
+    def test_login(self):
+        self.env.cr.execute('update res_users set sso_key = null;')
+        self.env.cr.commit()
```
simahawk: is this commit really needed? can't we use a write on res.users. Also, shouldn't the key be empty already at this stage?

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -0,0 +1,93 @@
...
+        #
+        res_users_obj = self.registry('res.users')
```
simahawk: self.env

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -0,0 +1,93 @@
...
+        res = res_users_obj.authenticate(
+            common.get_db_name(), 'admin', 'admin', None)
+        uid = res
+        self.assertTrue(res, "Basic login must works as expected")
+        token = "123456"
+        res = res_users_obj.authenticate(
+            common.get_db_name(), 'admin', token, None)
+        self.assertFalse(res)
+        # mimic what the new controller do when it find a value in
```
simahawk: s/find/finds

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -0,0 +1,93 @@
...
+        # the http header (HTTP_REMODE_USER)
```
simahawk: s/REMODE/REMOTE

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @unittest.skipIf(os.environ.get('TRAVIS'),
+                     'When run by travis, tests runs on a database with all '
+                     'required addons from server-tools and their dependencies'
+                     ' installed. Even if `auth_from_http_remote_user` does '
+                     'not require the `mail` module, The previous installation'
+                     ' of the mail module has created the column '
+                     '`notification_email_send` as REQUIRED into the table '
+                     'res_partner. BTW, it\'s no more possible to copy a '
+                     'res_user without an intefirty error')
```
simahawk: s/intefirty/integrity

simahawk: just to check: is this still needed?

simahawk (changes requested):

auth_from_http_remote_user/README.rst
TDu: Removed

auth_from_http_remote_user/controllers/main.py
TDu: Thanks for the explanation, done.

auth_from_http_remote_user/controllers/main.py
TDu: Done

auth_from_http_remote_user/models/res_users.py
TDu: Oops, removed

auth_from_http_remote_user/README.rst
TDu: Changed

auth_from_http_remote_user/README.rst
TDu: Done

auth_from_http_remote_user/__manifest__.py
TDu: Done

auth_from_http_remote_user/__manifest__.py
TDu: Done

auth_from_http_remote_user/__manifest__.py
TDu: Done

auth_from_http_remote_user/tests/test_res_users.py
TDu: Changed this file and the others

auth_from_http_remote_user/tests/test_res_users.py
TDu: done

auth_from_http_remote_user/tests/test_res_users.py
TDu: done

auth_from_http_remote_user/tests/test_res_users.py
TDu: Done

auth_from_http_remote_user/controllers/main.py
TDu: Done

auth_from_http_remote_user/models/res_users.py
TDu: Changed

auth_from_http_remote_user/tests/test_res_users.py
TDu: Not sure but are we not bypassing the ORM because the authentication methods uses a different env ?

simahawk: that's not my concern :) My concern is that if you commit you are going to commit any records that might be in the pool and when you reuse a test db (like when you are in dev mode) you can have strange behaviors/errors because of persistent test records.

TDu: Well tried removing it and the test failed.

guewen: Isn't because of the cache? Could you try `self.env.clear()`?

TDu: Nope tests failed as well with `self.env.clear()`
Maybe because the authentication process by pass the cache ?

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -24,12 +26,25 @@ def mock_cursor(cr):
...
+        self.assertEqual(self.user.sso_key, key)
 
     def test_login(self):
-        self.env.cr.execute('update res_users set sso_key = null;')
-        self.env.cr.commit()
-        #
+        reg = registry(self.env.cr.dbname)
+        with api.Environment.manage():
+            with reg.cursor() as cr:
+                env = api.Environment(cr, 1, {})
```
simahawk: you are already admin here, you can use self._uid or self.env.user.id

auth_from_http_remote_user/README.rst
```diff
@@ -0,0 +1,171 @@
...
+
+Finally reload the configuration: ::
+
+   $ sudo service apache2 reload
+
+Open your browser and go to MY_VHOST.com. If everything is well configured, you
+are prompted for a login and password outside Odoo and are automatically
+logged in the system.
+
+.. [#] Shibolleth, Tivoli access manager, ..
```
simahawk: *Shibboleth

auth_from_http_remote_user/README.rst
```diff
@@ -0,0 +1,171 @@
...
+.. [#] Based on a ubuntu 12.04 env
```
simahawk: still 12.04? or 16.04 now?

auth_from_http_remote_user/controllers/main.py
```diff
@@ -0,0 +1,62 @@
...
+            return
+        request_login = request.session.login
+        if request_login:
+            if request_login == login:
+                # Already authenticated
+                return
+            else:
+                request.session.logout(keep_db=True)
+        try:
+            user = utils.get_user(request.env['res.users'], login)
```
simahawk: any benefit in moving this to an util? Why not an instance method and `self.get_user`? It's easier to override it if needed too.

auth_from_http_remote_user/controllers/main.py
```diff
@@ -0,0 +1,62 @@
...
+            if not user:
+                # HTTP_REMOTE_USER login not found in database
+                request.session.logout(keep_db=True)
+                raise http.AuthenticationError()
+            # Logging SSO user using separate environment as the authentication
+            # later on is done in a specific environment as well
+            with api.Environment.manage():
+                with request.env.registry.cursor() as cr:
+                    env = api.Environment(cr, 1, {})
```
simahawk: user odoo.SUPERUSER_ID

auth_from_http_remote_user/models/res_users.py
```diff
@@ -0,0 +1,34 @@
...
+
+        Generate a key for authentication and update the user
+        """
+        key = utils.randomString(utils.KEY_LENGTH, '0123456789abcdef')
+        user.with_env(env).sudo().write({'sso_key': key})
+        return key
+
+    def check_credentials(self, password):
+        """Check credentials for SSO user"""
+        res = self.sudo().search([('id', '=', self._uid),
```
simahawk: self.env.user.id

auth_from_http_remote_user/tests/test_res_users.py
```diff
@@ -0,0 +1,92 @@
...
+                common.get_db_name(), 'admin', token, None)
+            self.assertTrue(res)
+
+    @unittest.skipIf(os.environ.get('TRAVIS'),
+                     'When run by travis, tests runs on a database with all '
+                     'required addons from server-tools and their dependencies'
+                     ' installed. Even if `auth_from_http_remote_user` does '
+                     'not require the `mail` module, The previous installation'
+                     ' of the mail module has created the column '
+                     '`notification_email_send` as REQUIRED into the table '
```
simahawk: just asking: is this still true? I don't see this field on v11

auth_from_http_remote_user/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    n = len(chrs)
+    return ''.join([chrs[randrange(n)] for _ in range(length)])
+
+
+def get_user(users, login):
+    """Search for an active user by login name"""
+    user = users.sudo().search([
+        ('login', '=', login),
+        ('active', '=', True)
+        ])
```
simahawk: limit=1

simahawk (changes requested):

auth_from_http_remote_user/README.rst
TDu: Well this doc is based and tested on 12.04 (I guess) and not by me. So would it not be misleading to change the version number ?

auth_from_http_remote_user/models/res_users.py
TDu: This how it is done in the overridden method
https://github.com/odoo/odoo/blob/11dff2b05c72adb01f4094aa5077218a29315dec/odoo/addons/base/res/res_users.py#L441

auth_from_http_remote_user/controllers/main.py
TDu: I just did not think it belongs here and it would be needed to be overriden.
But if you think it is a bad idea I will change it

TDu:
@guewen @simahawk This one is in need of another look. I do think it is ready.

auth_from_http_remote_user/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+{
+    'name': 'Authenticate via HTTP Remote User',
+    'version': '11.0.1.0.0',
+    'category': 'Tools',
+    'author': "Acsone SA/NV,Odoo Community Association (OCA)",
+    'maintainer': 'ACSONE SA/NV',
+    'website': 'https://github.com/OCA/server-auth',
+    'depends': ['base', 'web', 'base_setup'],
+    "license": "AGPL-3",
+    "installable": True,
```
simahawk: you can drop these flags

auth_from_http_remote_user/controllers/main.py
```diff
@@ -0,0 +1,83 @@
...
+        user = users.sudo().search([
+            ('login', '=', login),
+            ('active', '=', True)],
+            limit=1
+        )
+        if user:
+            return user[0]
+        return None
+
+    def logging_http_remote_user(self, env, user):
```
simahawk: logging or login? not clear IMO :)

simahawk (changes requested):
LGTM, minor comments

simahawk:
@lmignon would you mind passing by? :)

auth_from_http_remote_user/__manifest__.py
TDu: All three last ones  (installable, auto_install and application) ?

TDu: Well that is what I did. I am curious where I can find the default values for those in the code ?

auth_from_http_remote_user/controllers/main.py
TDu: Indeed 'logging' is more about cutting trees, which I want nothing to do with :smile: 
Changed, thanks for your review.

auth_from_http_remote_user/__manifest__.py
TDu: Default values are here https://github.com/odoo/odoo/blob/0050019390daa759ac5882ded79a6ee50ab4749e/odoo/modules/module.py#L310
So indeed they can all go.

guewen (approved):

PR #8, opened 2018-01-22
[MIG][11] auth_signup_verify_email WIP

simahawk:
TODO
- [x] finish migration steps
- [x] include fix https://github.com/OCA/server-tools/pull/602
- [x] address commit issue as described here https://github.com/odoo/odoo/issues/12237


simahawk:
@Yajo ping :)
A note regarding tests: I tried in many ways to avoid the commit. No way to achieve that w/ the cursor / env trick (which in any case sounds like is not needed in v11 as the env is already tied to the same cr).

I've kind of bypassed the issue by patching get_param. See https://github.com/OCA/server-auth/pull/8/commits/7cec1101329bb39b2a7db21062b3527b2065d49b

Regarding my fix for user lang, I tried to add a test but once again I got stuck on the crapy HTTP case behavior and even if I set cookies, languages in requests, whatever: the lang does not land on the user (which in any case is not even retrievable). My last attempt (damn, so many attempts...) is [here](https://github.com/simahawk/server-auth/commit/0acc7d809bf90f5d4873f5e93c5f7d52529ffd49).

I don't think this is a must have test so is not a blocker for this PR now. Still, maybe you have an idea on how to fix it :wink:

auth_signup_verify_email/controllers/main.py
```diff
@@ -1,12 +1,12 @@
...
-# © 2015 Antiun Ingeniería, S.L.
+# Copyright 2015 Antiun Ingeniería, S.L.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import logging
 from odoo import _, http
 from odoo.addons.auth_signup.controllers.main import AuthSignupHome
 
-_logger = logging.getLogger(__name__)
+request = http.request
```
yajo: `from odoo.http import request` seems more idiomatic

auth_signup_verify_email/controllers/main.py
```diff
@@ -1,12 +1,12 @@
...
+_logger = logging.getLogger('signup_verify_email.controller')
```
yajo: `__name__` is most standard IMHO

auth_signup_verify_email/controllers/main.py
```diff
@@ -15,41 +15,71 @@
...
+        return _("Check your email to activate your account!")
+
     def passwordless_signup(self, values):
         qcontext = self.get_auth_signup_qcontext()
 
         # Check good format of e-mail
         if not validate_email(values.get("login", "")):
-            qcontext["error"] = _("That does not seem to be an email address.")
-            return http.request.render("auth_signup.signup", qcontext)
+            qcontext["error"] = self.msg_invalid_email_error
```
yajo: This change makes not much sense. 😕 

All it is doing is getting the message from a property method instead of just getting it directly. It's easier as it was before.

If you have a customization addon that requires a different translation for this text, you can add dependency on this addon and add a `l18n_extra/it.po` file where you overwrite other addons' translations. It's the standard method to do these things, and again way easier than declaring your own controller, overriding those property methods and adding a new .po file.

The same applies for other similar changes.

auth_signup_verify_email/tests/test_verify_email.py
```diff
@@ -1,47 +1,54 @@
...
-        super(UICase, self).tearDown()
-        self.icp.set_param(
-            "auth_signup.allow_uninvited", self.old_allow_uninvited)
-        self.cr.commit()
+    def patch(self, obj, key, val):
+        """Overidden to keep old method."""
+        old = getattr(obj, key)
+        setattr(obj, key, val)
+        setattr(obj, 'old_' + key, old)
+        self.addCleanup(setattr, obj, key, old)
```
yajo: Isn't it better to just use `mock.patch`?

yajo (changes requested):
Your tests are fixed in https://github.com/simahawk/server-auth/pull/1 😉

auth_signup_verify_email/controllers/main.py
simahawk: it sucks if you have to mute loggers, see tests :wink:

auth_signup_verify_email/controllers/main.py
simahawk: fine

auth_signup_verify_email/controllers/main.py
simahawk: and this makes your "change that message" task easier? :smile: 

In any case if you are customizing stuff in you own addons you'll get translations for it especially if you have a multilang website (you have all your translations in one place).
Plus, if you have only one lang, it's even faster do handle it like this then handling extra translations. 

Personally I don't like to change a message via translations, sounds like a dirt&quick fix. But it's my idea :stuck_out_tongue_winking_eye:

auth_signup_verify_email/tests/test_verify_email.py
simahawk: well, there's a `patch` method in odoo tests common https://github.com/odoo/odoo/blob/11.0/odoo/tests/common.py#L192 
I was using it then I realized that I needed the old method inside the patch and I overridden. 
That's it :wink:

simahawk: just spotted your PR, I'll replace it anyway if it gets green (was not green in my local tests :cry: )

auth_signup_verify_email/controllers/main.py
yajo: If there's a standarized way of doing things, while working in community I still think it's better to follow the standard than to fight it for the sake of saving some keystrokes.

auth_signup_verify_email/controllers/main.py
yajo: Downloading a `.pot` of a couple of addons and putting translations in a folders might not be what you desire, but at the end of the day, it's a very easy task to do (it can be even done through GUI by your translation team) and if we follow that logic, we'd have to create a method for every single translation in Odoo. But it turns out that method already exists: `_()`

Besides, while that system exists, now if I have to create a sub-addon of this one, I have 2 ways of doing the same thing: subclassing and doing the `i18_extra` thing. Which one to use? Maybe subclassing for this addon and `i18n_extra` for all other addons? Why the difference? It's the typical case of reinventing the wheel. You'd possibly end up just using `i18n_extra` since it works the same all across Odoo. [The Zen of Python](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python) says:

> There should be one-- and preferably only one --obvious way to do it.

I think that principle applies here.

So, again, if there's a standarized way of doing things, while working in community I still think it's better to follow the standard than to fight it.

auth_signup_verify_email/controllers/main.py
simahawk: I think that's just laziness, as you type more to give proper naming :smile: 
anyway: rolling back

auth_signup_verify_email/controllers/main.py
simahawk: I strongly disagree here: mocking a translation instead of changing the original according to your needs is not to follow the zen of python. Plus, you can still do that via UI, Plus 2, this change does not make the module more complex, the contrary.
But it's my personal POV, the module is yours and I don't want to spend more time on this: rolling back :stuck_out_tongue_winking_eye:

simahawk:
@Yajo rebased, your fix integrated, all comments - hopefully - attended.

auth_signup_verify_email/controllers/main.py
```diff
@@ -49,7 +53,22 @@ def passwordless_signup(self, values):
...
+        """Retrieve current user language."""
+        langs = request.env['res.lang'].search_read([], ['code'])
+        supported_langs = [
+            lang['code'] for lang in langs
+        ]
+        lang = 'en_US'  # fallback
+        if hasattr(request, 'website'):
+            # default to site lang
+            lang = request.website.default_lang_code
+        if request.lang in supported_langs:
```
yajo: Is it even possible that a request that comes from the website has a lang that is not among the langs accepted by the website? 🤔  I think that in such case, Odoo will redirect the request to the website's default language and then process it. In such case, all of this method makes no sense and you can just use `request.lang`.

auth_signup_verify_email/tests/test_verify_email.py
```diff
@@ -54,19 +42,22 @@ def search_text(self, doc, text):
 
     def test_bad_email(self):
         """Test rejection of bad emails."""
-        self.data["login"] = "bad email"
-        doc = self.html_doc(data=self.data)
+        data = self.data.copy()
```
yajo: You don't need to change this. `self.setUp()` is ran before each test, and it redeclares `self.data`.

auth_signup_verify_email/tests/test_verify_email.py
```diff
@@ -54,19 +42,22 @@ def search_text(self, doc, text):
...
     def test_good_email(self):
         """Test acceptance of good emails.
 
         This test could lead to success if your SMTP settings are correct, or
         to failure otherwise. Any case is expected, since tests usually run
         under unconfigured demo instances.
         """
-        self.data["login"] = "good@example.com"
-        doc = self.html_doc(data=self.data)
+        data = self.data.copy()
```
yajo: Again, no need for this change.

yajo (changes requested):
I think there are some more changes that we don't really need, or maybe I don't understand them. 🤔  Could you check these please?

auth_signup_verify_email/tests/test_verify_email.py
simahawk: I know, it was supposed to go away during rebase/cleanup. Guess I'm doing too many things at the same time :stuck_out_tongue_closed_eyes:

auth_signup_verify_email/controllers/main.py
simahawk: hmm, I remember I've done this for a good reason but I was so idiot to not put proper comments :sob: 
Nevertheless, I did this on v9 and years ago, so: let's drop it. I can put it back if needed later on :+1:

tschanzt (approved):
Looks good to me.

yajo (approved):

PR #9, opened 2018-01-26
[ADD] auth_from_http_remote_user_role

TDu:
A new module to assign user role defined in the http header
It is based on auth_from_http_remote_user  #7


auth_from_http_remote_user_role/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    'name': 'Assign roles via HTTP Header',
+    'summary': 'Assign roles found in the HTTP header, to the connected user',
+    'version': '11.0.1.0.0',
+    'category': 'Tools',
+    'website': 'https://github.com/OCA/server-auth',
+    'author': 'Thierry Ducrest, Odoo Community Association (OCA)',
```
guewen: Author should be `Camptocamp, Odoo Community Association (OCA)`

auth_from_http_remote_user_role/controllers/main.py
```diff
@@ -0,0 +1,49 @@
...
+        header using the field and separator predefined.
+        """
+        roles = []
+        headers = http.request.httprequest.headers.environ
+        roles_in_header = headers.get(self._REMOTE_USER_ROLE, None)
+        if roles_in_header:
+            roles = roles_in_header.split(self._REMOTE_USER_ROLE_SEPARATOR)
+        return roles
+
+    def _authenticate_user(self, user, cr):
```
guewen: What is the reason to use direct SQL rather than the ORM? if there is a reason, could you document it as comments? Besides, cache must be invalidated after the executes.

auth_from_http_remote_user_role/controllers/main.py
```diff
@@ -0,0 +1,49 @@
...
+        """Update roles assigned to user """
+        cr.execute('''DELETE FROM res_users_role_line
+                        WHERE user_id=%s''', [user.id])
```
guewen: I don't think we should delete the roles every time if they don't change. It increases the risk of concurrency errors and it's doing useless delete/writes. Most of the time, the roles won't change.

guewen (dismissed):

auth_from_http_remote_user_role/controllers/main.py
sbidoul: Be careful with performance: changing roles implies writing to user groups which implies cache invalidation.

TDu:
Back at it, thank you for the reviews.
I removed all the direct sql queries and made sure that the roles where updated only when needed.
Please have another look.

auth_from_http_remote_user_role/README.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
+
+Contributors
+------------
+
+* Thierry Ducrest <thierry.ducrest@camptocamp.com
```
asaunier: Missing trailing ``>``

auth_from_http_remote_user_role/tests/tests_user_role_change.py
```diff
@@ -0,0 +1,33 @@
...
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo.tests.common import TransactionCase
+
+
+class TestUserRoleChange(TransactionCase):
+
+    def setUp(self):
+        super(TestUserRoleChange, self).setUp()
```
asaunier: Could be written ``super().setUp()``

auth_from_http_remote_user_role/tests/tests_user_role_change.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
asaunier: No longer needed in Python 3

asaunier (commented):
I have spotted only very minor points. Else LGTM :)

TDu:
Thanks made the changes

TDu:
Waiting on #7 to be merged to finish this one

p-tombez:
@TDu When testing this module, I found a bug when you try to remove user roles: the server will return a 500 error. Here is a fix : https://github.com/TDu/server-auth/pull/1

auth_from_http_remote_user_role/tests/tests_user_role_change.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo.tests.common import TransactionCase
+
+
+class TestUserRoleChange(TransactionCase):
+
+    def setUp(self):
```
simahawk, line 9: any chance to see ir.http tests here? :innocent:

simahawk (commented):
Looks mostly good. Would be nice to see some more test cov.
Not sure is related to this PR but tests are failing on auth_brute_force

auth_from_http_remote_user/controllers/main.py
```diff
@@ -0,0 +1,83 @@
...
+        user = users.sudo().search([
+            ('login', '=', login),
+            ('active', '=', True)],
+            limit=1
+        )
+        if user:
+            return user[0]
+        return None
+
+    def logging_http_remote_user(self, env, user):
```
simahawk: @TDu @guewen I think this meth is never used since it has been renamed https://github.com/OCA/server-auth/pull/7/files#diff-00f6cb1d0bf2f276bf85d1ccf7615dfbR43

grindtildeath (approved):
LGTM :+1:

TDu:
@guewen  :+1:  Thanks for your help and the great improvements on this.
https://github.com/OCA/server-auth/pull/9/files#diff-d8f62dd3485592027fe4a40b3db0b97dR55 -> That is quiet enlightening :zap:

auth_from_http_remote_user_role/models/ir_http.py
```diff
@@ -0,0 +1,131 @@
...
+            _logger.debug(
+                'user %s received header %s and previously had %s',
+                user.id, roles_in_header, user.last_http_header_roles
+            )
+
+            user = user.with_env(env)
+            new_roles = set()
+            role_codes = cls._get_http_role_codes(roles_in_header)
+            if role_codes:
+                new_roles = set(env['res.users.role'].search(
```
simahawk, line 125: small detail: I don't think you need to call set() on .ids as they should be already unique

simahawk (approved):
LG, would be nice to see more test cov :pray:

PR #14, opened 2018-04-20
[IMP] password_security: disallow password/login/name similarity

samek-h:
Hi,
I've added some additional functionality to the "password_security" module that provides one more criterion for creating secure passwords - disallows similarity with user's login (whole login or the name part of e-mail address used as a login name) and with user's display name.

If you have any additional suggestions, please, let me know.
Please note that this contribution has been created as a part of my job.

Thank you.

Regards,
Jan Samek, Orgis IT, https://www.orgis.cz/

### DIFF
--- a/password_security/README.rst
+++ b/password_security/README.rst
@@ -17,6 +17,7 @@ It contains features such as
 * Password minimum number of uppercase letters
 * Password minimum number of numbers
 * Password minimum number of special characters
+* Password similarity to user's name or login
 
 Configuration
 =============
@@ -34,18 +35,19 @@ Settings & Defaults
 
 These are defined at the company level:
 
-=====================  =======   ===================================================
- Name                  Default   Description                             
-=====================  =======   ===================================================
- password_expiration   60        Days until passwords expire
- password_length       12        Minimum number of characters in password
- password_lower        0         Minimum number of lowercase letter in password
- password_upper        0         Minimum number of uppercase letters in password
- password_numeric      0         Minimum number of number in password
- password_special      0         Minimum number of unique special character in password
- password_history      30        Disallow reuse of this many previous passwords
- password_minimum      24        Amount of hours that must pass until another reset
-=====================  =======   ===================================================
+=================================   =======   =======================================================
+ Name                               Default   Description
+=================================   =======   =======================================================
+ password_expiration                60        Days until passwords expire
+ password_length                    12        Minimum number of characters in password
+ password_lower                     0         Minimum number of lowercase letter in password
+ password_upper                     0         Minimum number of uppercase letters in password
+ password_numeric                   0         Minimum number of number in password
+ password_special                   0         Minimum number of unique special character in password
+ password_history                   30        Disallow reuse of this many previous passwords
+ password_minimum                   24        Amount of hours that must pass until another reset
+ password_not_similar_name_login    True      Disallow password similar to user's name or login
+=================================   =======   =======================================================
 
 Usage
 =====
--- a/password_security/models/res_company.py
+++ b/password_security/models/res_company.py
@@ -49,3 +49,9 @@ class ResCompany(models.Model):
         default=24,
         help='Amount of hours until a user may change password again',
     )
+    password_not_similar_name_login = fields.Boolean(
+        'Disallow passwords similar to name or login',
+        default=True,
+        help='When enabled, users\' passwords can\'t match closely to their '
+             'names or logins.'
+    )
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -63,7 +63,7 @@ class ResUsers(models.Model):
                 company_id.password_special) + ' character)')
         if company_id.password_not_similar_name_login:
             message.append(
-                '\n* ' + 'Must not be similar to the user\'s login or name.')
+                '\n* Must not be similar user\'s login or name.')
         if message:
             message = [_('Must contain the following:')] + message
         if company_id.password_length:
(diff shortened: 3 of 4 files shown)


password_security/models/res_users.py
```diff
@@ -72,6 +76,31 @@ def password_match_message(self):
...
+        self._check_password_similar_login(password, self.login)
+        return True
+
+    @api.multi
+    def _check_password_similar_name(self, password, name):
+        self.ensure_one()
+        if self.company_id.password_not_similar_name_login:
+            # remove non-alnums from user's name for comparison
+            name_alnum = re.sub(r'[\W_]', '', name)
+            if name_alnum.lower() == password.lower():
```
MiquelRForgeFlow: But that is checking equality, not similarity 🤔

password_security/models/res_users.py
```diff
@@ -72,6 +76,31 @@ def password_match_message(self):
...
+        return True
+
+    @api.multi
+    def _check_password_similar_login(self, password, login):
+        self.ensure_one()
+        if self.company_id.password_not_similar_name_login:
+            match = re.match(r'([^@]+)(?:@[^@]+)?', login)
+            login_name_only = match.group(1)
+            login_whole = match.group(0)
+            if login_name_only.lower() == password.lower():
```
MiquelRForgeFlow: But that is checking equality, not similarity :thinking:

MiquelRForgeFlow (changes requested):

password_security/models/res_users.py
samek-h: OK, case- and non-alnums-ignoring equality... Maybe some common subsequence or substring matching would do better.

MiquelRForgeFlow: check https://docs.python.org/3/library/re.html

samek-h: For LCS i'd prefer [difflib.SequenceMatcher](https://docs.python.org/3/library/difflib.html#difflib.SequenceMatcher) . Or did you mean that I've done something wrong with [re](https://docs.python.org/3/library/re.html)?

MiquelRForgeFlow: nothing wrong

password_security/models/res_users.py
```diff
@@ -72,6 +76,31 @@ def password_match_message(self):
...
+            name_alnum = re.sub(r'[\W_]', '', name)
+            if name_alnum.lower() == password.lower():
+                raise PassError(self.password_match_message())
+        return True
+
+    @api.multi
+    def _check_password_similar_login(self, password, login):
+        self.ensure_one()
+        if self.company_id.password_not_similar_name_login:
+            match = re.match(r'([^@]+)(?:@[^@]+)?', login)
```
yajo: ~~Missing `^` and `$`?~~ Unneeded, see above.

password_security/models/res_users.py
```diff
@@ -60,6 +61,9 @@ def password_match_message(self):
         if company_id.password_special:
             message.append('\n* ' + 'Special character (At least ' + str(
                 company_id.password_special) + ' character)')
+        if company_id.password_not_similar_name_login:
+            message.append(
+                '\n* ' + 'Must not be similar to the user\'s login or name.')
```
yajo: Yikes! Why not `"\n* Must not be similar to the user's login or name."`?

Translate it please.

password_security/models/res_users.py
yajo: This is unneededly overly complex IMHO. Simpler:

```python
password = password.lower()
name = name.lower()
if password in name or name in password:
    raise PassError(self.password_match_message())
```

password_security/models/res_users.py
```diff
@@ -72,6 +76,31 @@ def password_match_message(self):
     def _check_password(self, password):
         self._check_password_rules(password)
         self._check_password_history(password)
+        self._check_password_similar_name(password, self.name)
+        self._check_password_similar_login(password, self.login)
```
yajo: Why not just 1 method passed with both params?

password_security/models/res_users.py
yajo: I guess equality is a good 1st step towards similarity.

yajo (changes requested):

mreficent:
Travis fails in auth_brute_force 🤔

password_security/models/res_company.py
```diff
@@ -49,3 +49,9 @@ class ResCompany(models.Model):
         default=24,
         help='Amount of hours until a user may change password again',
     )
+    password_not_similar_name_login = fields.Boolean(
+        'Disallow passwords similar to name or login',
+        default=True,
+        help='When enabled, users\' passwords can\'t match closely to their '
```
yajo, line 55: Isn't it easier to use double quotes?

password_security/models/res_users.py
```diff
@@ -60,6 +61,9 @@ def password_match_message(self):
         if company_id.password_special:
             message.append('\n* ' + 'Special character (At least ' + str(
                 company_id.password_special) + ' character)')
+        if company_id.password_not_similar_name_login:
+            message.append(
+                '\n* Must not be similar user\'s login or name.')
```
yajo, line 66: Same here

password_security/README.rst
```diff
@@ -34,18 +35,19 @@ Settings & Defaults
...
+=================================   =======   =======================================================
+ password_expiration                60        Days until passwords expire
+ password_length                    12        Minimum number of characters in password
+ password_lower                     0         Minimum number of lowercase letter in password
+ password_upper                     0         Minimum number of uppercase letters in password
+ password_numeric                   0         Minimum number of number in password
+ password_special                   0         Minimum number of unique special character in password
+ password_history                   30        Disallow reuse of this many previous passwords
+ password_minimum                   24        Amount of hours that must pass until another reset
+ password_not_similar_name_login    True      Disallow password similar to user's name or login
```
yajo, line 49: I never liked those fields where you have to set `True` to say *not*. :thinking: Can't we phrase it in a positive way? Something like `password_login_similar_forbidden`?

MiquelRForgeFlow:
I think the problem with travis is that the PR needs to be updated with a rebase.

PR #17, opened 2018-05-09
[11.0][MIG] auth_totp

MiquelRForgeFlow:
Standard migration.


pedrobaeza:
Check Travis

MiquelRForgeFlow:
@pedrobaeza GREEN :green_apple: :green_apple: :green_apple:

auth_totp/README.rst
```diff
@@ -0,0 +1,102 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
+
+Contributors
+------------
+
+* Oleg Bulkin <obulkin@laslabs.com>
```
simahawk, line 87: Add "Do not contact contributors directly about support or help with technical issues."
See https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

auth_totp/models/res_users_authenticator.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    user_id = fields.Many2one(
+        comodel_name='res.users',
+        ondelete='cascade',
+    )
+
+    @api.multi
+    @api.constrains('user_id')
+    def _check_has_user(self):
+        self.filtered(lambda r: not r.user_id).unlink()
```
simahawk, line 45: a constrain that deletes?

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,117 @@
...
+
+            totp = pyotp.TOTP(record.secret_key)
+            provisioning_uri = totp.provisioning_uri(
+                record.user_id.display_name.encode('utf-8'),
+                issuer_name=record.user_id.company_id.display_name,
+            )
+            provisioning_uri = urls.url_quote(provisioning_uri)
+
+            qr_width = qr_height = 300
+            tag_base = '<img src="/report/barcode/?type=QR&amp;'
```
simahawk, line 80: not mandatory but I'd use a dict + urlencode

auth_totp/wizards/res_users_authenticator_create.py
```diff
@@ -0,0 +1,117 @@
...
+class ResUsersAuthenticatorCreate(models.TransientModel):
+    _name = 'res.users.authenticator.create'
+    _description = 'MFA App/Device Creation Wizard'
+
+    name = fields.Char(
+        string='Authentication App/Device Name',
+        help='A name that will help you remember this authentication'
+             ' app/device',
+        required=True,
+        index=True,
```
simahawk: why all these indexes? Isn't this a wizard? And I don't see any search :)

requirements.txt
```diff
@@ -3,3 +3,4 @@ Pygments
 validate_email
 # auth_saml
 lasso
+pyotp
```
simahawk: can you add the name of the module that requires it on top of it?

simahawk (changes requested):

auth_totp/models/res_users_authenticator.py
MiquelRForgeFlow: That's from previous version. What should I do?

simahawk: I dunno... :) How is it supposed to work? Who's resetting user_id? Or who's creating a record w/ no user_id?

JordiBForgeFlow (approved):
:+1: Functional review.

auth_totp/models/res_users_authenticator.py
JordiBForgeFlow: @obulkin Perhaps you know better. But to me, user_id should be a required field in 'res.users.authenticator'. Then this @api.constrains would not be required at all.

obulkin: Existing authenticators are only exposed through an o2m field in a user view, which doesn't support deletion, so this logic is here to clean up orphaned authenticators when they're unmapped from a user via the view. You could make the field required and drop the `constrains`, but you would need to change the way that users manage their authenticators through the UI.

PR #23, opened 2018-05-28
[11.0][auth_brute_force][IMP] Set CIDR configuration & some extra improvements

### LINKED ISSUE #22: [11.0][auth_brute_force] Express whitelisted IPs in CIDR notation
Often companies will want to maintain a whitelist of IP address expressed by ranges of IP's instead of having to express them one by one.

See https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation
And a sample implementation in python of logic to check if an IP address is whithin a range expressed by a CIDR expression: https://diego.assencio.com/?index=85e407d6c771ba2bc5f02b17714241e2

etobella:
Sets ICDR validation as expected from #22 

Edition and deletion of lines is disallowed and items can be set to unbanned in order to simplify the process. Now, you can set to unbanned in order to reset the counters and you can also maintain all the history.

Fix #22

### DIFF
--- a/auth_brute_force/models/res_authentication_attempt.py
+++ b/auth_brute_force/models/res_authentication_attempt.py
@@ -37,15 +37,6 @@ class ResAuthenticationAttempt(models.Model):
     whitelisted = fields.Boolean(
         compute="_compute_whitelisted",
     )
-    banned = fields.Boolean(
-        compute="_compute_banned"
-    )
-
-    @api.multi
-    @api.depends('remote', 'login')
-    def _compute_banned(self):
-        for item in self:
-            item.banned = not self._trusted(item.remote, item.login)
 
     @api.multi
     @api.depends('remote')
@@ -193,7 +184,7 @@ class ResAuthenticationAttempt(models.Model):
     @api.multi
     def action_unban(self):
         self.ensure_one()
-        if self.banned:
+        if self.result == 'banned':
             self.write({
                 'result': 'unbanned',
             })
--- a/auth_brute_force/models/res_authentication_attempt.py
+++ b/auth_brute_force/models/res_authentication_attempt.py
@@ -191,11 +191,8 @@ class ResAuthenticationAttempt(models.Model):
     def action_unban(self):
         self.ensure_one()
         if self.banned:
-            self.create({
-                'remote': self.remote,
-                'login': self.login,
+            self.write({
                 'result': 'unbanned',
-
             })
 
     @api.multi
--- a/auth_brute_force/models/res_authentication_attempt.py
+++ b/auth_brute_force/models/res_authentication_attempt.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import json
+from distutils import util
 import logging
 import ipaddress
 from urllib.request import urlopen
@@ -24,6 +25,7 @@ class ResAuthenticationAttempt(models.Model):
             ('successful', 'Successful'),
             ('failed', 'Failed'),
             ('banned', 'Banned'),
+            ('unbanned', 'Unbanned')
         ],
         index=True,
     )
@@ -35,13 +37,22 @@ class ResAuthenticationAttempt(models.Model):
     whitelisted = fields.Boolean(
         compute="_compute_whitelisted",
     )
+    banned = fields.Boolean(
+        compute="_compute_banned"
+    )
+
+    @api.multi
+    @api.depends('remote', 'login')
+    def _compute_banned(self):
+        for item in self:
+            item.banned = not self._trusted(item.remote, item.login)
 
     @api.multi
     @api.depends('remote')
     def _compute_metadata(self):
-        if self.env["ir.config_parameter"].sudo().get_param(
-            'auth_brute_force.check_remote', True
-        ):
+        if util.strtobool(self.env["ir.config_parameter"].sudo().get_param(
+            'auth_brute_force.check_remote', 'True'
+        )):
             for item in self:
                 url = GEOLOCALISATION_URL.format(item.remote)
                 try:
@@ -56,7 +67,8 @@ class ResAuthenticationAttempt(models.Model):
                     item.remote_metadata = "\n".join(
                         '%s: %s' % pair for pair in res.items())
 
-    def check_whitelist(self, ip):
+    @api.model
+    def _check_whitelist(self, ip):
         for whitelist in self._whitelist_remotes():
             try:
                 if (
@@ -71,7 +83,7 @@ class ResAuthenticationAttempt(models.Model):
     @api.multi
     def _compute_whitelisted(self):
         for one in self:
-            one.whitelisted = self.check_whitelist(one.remote)
+            one.whitelisted = self._check_whitelist(one.remote)
 
     @api.model
     def _hits_limit(self, limit, remote, login=None):
@@ -94,7 +106,7 @@ class ResAuthenticationAttempt(models.Model):
             domain += [("login", "=", login)]
         # Find last successful login
         last_ok = self.search(
-            domain + [("result", "=", "successful")],
+            domain + [("result", "in", ["successful", "unbanned"])],
             order='create_date desc',
             limit=1,
         )
@@ -125,7 +137,7 @@ class ResAuthenticationAttempt(models.Model):
             return True
         get_param = self.env["ir.config_parameter"].sudo().get_param
         # Whitelisted remotes always pass
-        if self.check_whitelist(remote):
+        if self._check_whitelist(remote):
             return True
         # Check if remote is banned
         limit = int(get_param("auth_brute_force.max_by_ip", 50))
@@ -173,6 +185,17 @@ class ResAuthenticationAttempt(models.Model):
             "auth_brute_force.whitelist_remotes",
             ",".join(whitelist),
         )
+
+    @api.multi
+    def action_unbanned(self):
+        self.ensure_one()
+        if self.banned:
+            self.create({
+                'remote': self.remote,
+                'login': self.login,
+                'result': 'unbanned',
+
+            })
 
     @api.multi
     def action_whitelist_remove(self):
--- a/auth_brute_force/models/res_authentication_attempt.py
+++ b/auth_brute_force/models/res_authentication_attempt.py
@@ -111,12 +111,15 @@ class ResAuthenticationAttempt(models.Model):
             order='create_date desc',
             limit=1,
         )
+
         if last_ok:
             domain += [("create_date", ">", last_ok.create_date)]
         # Count failures since last success, if any
         recent_failures = self.search_count(
-            domain + [("result", "not in",  ["successful", "unbanned"])],
-        )
+            domain + [
+                ("result", "!=", "successful"),
+                ("result", "!=", "unbanned"),
+            ])
         # Did we hit the limit?
         return recent_failures >= limit
 
--- a/auth_brute_force/models/res_authentication_attempt.py
+++ b/auth_brute_force/models/res_authentication_attempt.py
@@ -39,24 +39,30 @@ class ResAuthenticationAttempt(models.Model):
     @api.multi
     @api.depends('remote')
     def _compute_metadata(self):
-        for item in self:
-            url = GEOLOCALISATION_URL.format(item.remote)
-            try:
-                res = json.loads(urlopen(url, timeout=5).read())
-            except Exception:
-                _logger.warning(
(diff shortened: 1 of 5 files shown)


auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -34,29 +37,52 @@ class ResAuthenticationAttempt(models.Model):
...
+                    _logger.warning(
+                        "Couldn't fetch details from %s",
+                        url,
+                        exc_info=True,
+                    )
+                else:
+                    item.remote_metadata = "\n".join(
+                        '%s: %s' % pair for pair in res.items())
+
+    def check_whitelist(self, ip):
```
yajo: `@api.model`. Make it private method, we don't want it to be called through RPC.

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -159,6 +185,17 @@ def action_whitelist_add(self):
...
         )
 
+    @api.multi
+    def action_unbanned(self):
+        self.ensure_one()
+        if self.banned:
+            self.create({
+                'remote': self.remote,
+                'login': self.login,
+                'result': 'unbanned',
```
yajo: IMHO it's better to remove this new state, because:

- Duplicated. We already have the "whitelisted" button. It's confusing the difference between "unbanned" and "whitelisted".
- It's creating an authentication attempt where there was none, so _it's actually lying_. If an attempt was banned, its history must say it was banned, no matter if later the remote is whitelisted. A new attempt must happen for it to pass, and in such case, it must be now "successful".

auth_brute_force/views/view.xml
```diff
@@ -10,6 +10,7 @@
             <tree
                 decoration-warning="result == 'failed'"
                 decoration-danger="result == 'banned'"
+                delete="0" create="0"
```
yajo: Permissions already do this, unless you're an admin, in which case IMHO you should be able to at least delete things.

auth_brute_force/views/view.xml
```diff
@@ -22,8 +23,14 @@
     <record id="view_res_authentication_attempt_form" model="ir.ui.view">
         <field name="model">res.authentication.attempt</field>
         <field name="arch" type="xml">
-            <form>
+            <form create="0" delete="0" edit="0">
```
yajo: Again, permissions already do this, not needed.

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -34,29 +37,52 @@ class ResAuthenticationAttempt(models.Model):
     whitelisted = fields.Boolean(
         compute="_compute_whitelisted",
     )
+    banned = fields.Boolean(
```
yajo: This field is confusing: we already have `result`, which says if it was banned; and we already have `whitelisted`, which says if the remote is whitelisted.

So IMHO you can remove this.

yajo (changes requested):
Hmm by reviewing this, I can see it effectively fixes #22, which is quite cool; but it seems you have added many unneeded features :sweat_smile: 

Please see comments below to keep it DRY. Thanks!

auth_brute_force/views/view.xml
etobella: I think it is a bad idea, because as an admin, you will be deleting accesses to your system. I am not sure if it is right by GDPR.

auth_brute_force/models/res_authentication_attempt.py
etobella: This field shows the current state of the relation login/remote. It is used in order to show the option to unban a login/remote.

auth_brute_force/models/res_authentication_attempt.py
etobella: Unbanned is used in order to clean your history. Then, the counters are set to 0.
If you are whitelisted, you can use brute force in order to attempt access, because you will always be trusted.

auth_brute_force/views/view.xml
yajo: Nothing in the GDPR explicitly forces you to record all logins AFAIK. And you can achieve this better if you use case requires that through permissions.

auth_brute_force/models/res_authentication_attempt.py
yajo: That's the case for whitelisted field also.

etobella: No, whitelisted shows if the remote is whitelisted.
Banned shows if the remote/login is banned. Obviously, it you are whitelisted you cannot be banned, but if you are not whitelisted you can be banned or unbanned.
IMHO, it is not the same.

auth_brute_force/views/view.xml
etobella: Well, maybe there is no detail on GDPR, but if you had an attack from a whitelisted remote, you could delete the attack history in order to clean your responsibilities. Do you think that option is GDPR compliant?

yajo: As an admin, you can uninstall the addon, which deletes all of that, and reinstall it, so that's not a valid argument AFAICS.

Besides, proving that you have in place a brute force protection is not something to hide usually.

auth_brute_force/models/res_authentication_attempt.py
yajo: OK I get it now, but then do not create a new attempt; just update current one and set to unbanned, or possibly update all of them which have the same login/remote combo since current record.

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -159,6 +186,17 @@ def action_whitelist_add(self):
             ",".join(whitelist),
         )
 
+    @api.multi
+    def action_unbanned(self):
```
yajo: action_unban

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -79,7 +106,7 @@ def _hits_limit(self, limit, remote, login=None):
             domain += [("login", "=", login)]
         # Find last successful login
         last_ok = self.search(
-            domain + [("result", "=", "successful")],
+            domain + [("result", "in", ["successful", "unbanned"])],
```
yajo, line 101: Don't you need to update also [line 117](https://github.com/OCA/server-auth/pull/23/files#diff-9520d806e75d6e4b88d87e00ff6709f1R117)?

auth_brute_force/models/res_authentication_attempt.py
yajo: Hmm I see. Then the duplication is not with `whitelisted`, but with `result=banned`. You don't need it in any case.

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -34,29 +37,53 @@ class ResAuthenticationAttempt(models.Model):
...
+            item.banned = not self._trusted(item.remote, item.login)
 
     @api.multi
     @api.depends('remote')
     def _compute_metadata(self):
-        for item in self:
-            url = GEOLOCALISATION_URL.format(item.remote)
+        if util.strtobool(self.env["ir.config_parameter"].sudo().get_param(
+            'auth_brute_force.check_remote', 'True'
+        )):
```
yajo: You can reduce the diff if you just `return` in case this is `False` :blush:

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -34,29 +37,53 @@ class ResAuthenticationAttempt(models.Model):
...
+            except ValueError:
+                continue
+        return False
 
     @api.multi
     def _compute_whitelisted(self):
-        whitelist = self._whitelist_remotes()
         for one in self:
-            one.whitelisted = one.remote in whitelist
+            one.whitelisted = self._check_whitelist(one.remote)
```
yajo: I'm thinking on the benefit of having a separate `_check_whitelist` method, and I don't see much of it... :thinking:  You can possibly reduce it with just:

```py
whitelist = self._whitelist_remotes()  # Make this method return the prefiltered list, pretransformed through ipaddress.ip_network
for one in self:
    ip = ipaddress.ip_address(one.remote)
    one.whitelisted = any(ip in net for net in whitelist)
```

yajo (changes requested):
Thanks, much clear now, although there are still changes to be made :blush: 

(BTW please push new commits instead of ammending, so it's easier to review changes; we'll squash later at merge)

auth_brute_force/models/res_authentication_attempt.py
etobella: Mmh, no, because it will not be shown if you are currently unbanned.
Maybe, you are banned, once you are unbanned, you should not be unbanned again.

auth_brute_force/views/view.xml
etobella: True, I will remove the options.

auth_brute_force/models/res_authentication_attempt.py
etobella: Well you are using it in line 140 and you don't have the record. For that reason, I made the method.

auth_brute_force/models/res_authentication_attempt.py
etobella: True! It is modified now.

auth_brute_force/models/res_authentication_attempt.py
yajo: Ah I see; OK, then at least provide a better naming for the method. `_check` is the standard prefix for constraints. Something like `_is_whitelisted` becomes more clear.

auth_brute_force/models/res_authentication_attempt.py
yajo: True, but [if you don't create a new record](https://github.com/OCA/server-auth/pull/23/files#r191369394), then you'd be getting `result=unbanned`, which can then be used to apply [the domain to show or not the button](https://github.com/OCA/server-auth/pull/23/files#diff-d66823c3bc47d3ae32e26197db149004R31).

etobella:
I think all changes were applied.
@Yajo, can you review again?

auth_brute_force/models/res_authentication_attempt.py
```diff
@@ -79,16 +98,19 @@ def _hits_limit(self, limit, remote, login=None):
...
+
         if last_ok:
             domain += [("create_date", ">", last_ok.create_date)]
         # Count failures since last success, if any
         recent_failures = self.search_count(
-            domain + [("result", "!=", "successful")],
-        )
+            domain + [
+                ("result", "!=", "successful"),
+                ("result", "!=", "unbanned"),
```
yajo: Use `not in`, just like above.

auth_brute_force/views/view.xml
```diff
@@ -24,6 +24,12 @@
         <field name="arch" type="xml">
             <form>
                 <header>
+                    <button
+                        name="action_unban"
+                        type="object"
+                        string="Set to unbanned"
+                        attrs="{'invisible': [('result', '=', 'banned')]}"
```
yajo: I think you need to use `!=` here.

yajo (changes requested):

auth_brute_force/models/res_authentication_attempt.py
etobella: `res.authentication.attempt` has null value on `result` when it is created. In order to not read this record, I added the value False in the not in. This behaviour is related to [odoo not in management](https://github.com/odoo/odoo/blob/11.0/odoo/osv/expression.py#L1209).

yajo (changes requested):
Neat! Code looks good. However, it seems your test is not being executed. Could you check why? I suspect on the decorator...

etobella:
I checked the test, and the problem is related to the decorator, I have made a version in order to solve the issue, but I have a problem.
The create_date is stored with milliseconds, but odoo removes them when I read the field. So, when I search, some extra records could be showed. It is not a problem on real life, because there is more than a second between attempts, but it creates problems on testing.
So, I think there are two options:
- Leave it untested
- Create a new field with the log_date

Which one do you prefer?

etobella:
The other option is to change the option domain += [("create_date", ">", last_ok.create_date)]` by `domain += [("id", ">", last_ok.id)]`. It will do the same, but I don't know if it is totally clean...
@Yajo

yajo:
Yes, that last option seems legit.

etobella:
@Yajo It is :green_apple:. Can you review again?

yajo (approved):

JordiBForgeFlow (approved):
:+1:

PR #25, opened 2018-06-08
remove group no one from saml provider

vincent-hatakeyama:


pedrobaeza:
Why removing it?

vincent-hatakeyama:
When you configure odoo to add SAML authentification, you shouldn’t need to activate debug/technical mode to access a very important menu.
For comparison, to access ldap settings with auth_ldap, you don’t need to activate debug.

pedrobaeza:
OK, I let original authors to decide about this PR.

alexey-pelykh:
Should this module & PR be considered abandoned due to a fact that module and PR has not been revisited for 5 months?

pedrobaeza:
Yes, it can be considered that way.

alexey-pelykh:
What would be your advice on this module? Since we do need it, yet waiting for original author to get back the module is not an option and some of changes from this PR seem valid

vincent-hatakeyama:
I think this should be closed in favor of  #107 .

PR #26, opened 2018-06-13
[FIX] auth_brute_force: Method names shouldn't lie

yajo:
If the method says it raises exceptions, it must do it.

@Tecnativa


pedrobaeza:
Are you sure about this? I remember it was silenced due to a side effect.

yajo:
Yes, sorry, this is wrong.

PR #31, opened 2018-06-22
[FIX] password_security: Allow mini-admin to create users

yajo:
In a normal Odoo deployment, somebody in group *Administration / Access Rights* should be able to create users; but if this addon is installed, it gets this error:

    The requested operation cannot be completed due to security restrictions. Please contact your system administrator.

    (Document type: Res Users Password History, Operation: create)

This is now tested and fixed.

@Tecnativa


pedrobaeza (approved):
Fast-tracking this when CIs finish

PR #44, opened 2018-10-18
[12.0][ADD] auth_ldap_attributes_sync

alexey-pelykh:
An enhanced replacement for `users_ldap_mail`


pedrobaeza:
Can you please remove the plural from the module name as you are now changing it and check CIs status?

alexey-pelykh:
you mean `attribute` instead of `attributes`?

pedrobaeza:
Yes, you can check at https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules that it's encouraged to do that for avoiding problems on the spelling, although historical modules that have been migrated version per version still continue with some plurals (or Odoo itself doesn't follow this rule).

alexey-pelykh:
Moved to #45

pedrobaeza:
As a note, you don't need to open a new PR. Just force pushing this one you can replace entire content.

alexey-pelykh:
Thanks for a hint! I've failed to find a way to change branch (which was renamed) of the PR, thus had to create a new one

pedrobaeza:
Oh, yeah, in that case, you can't do it, but hey, I don't rename branches in that cases. Only change the name of the module, commit with amend and forced push.

PR #45, opened 2018-10-18
[12.0][ADD] auth_ldap_attribute_sync

alexey-pelykh:
Supersedes #44


hbrunn:
this looks pretty much like the inverse of https://github.com/OCA/server-tools/tree/8.0/users_ldap_push

I think you can simplify your code quite a bit and at the same time make the module much more powerful by not hardcoding fields, but providing a model like https://github.com/OCA/server-tools/blob/8.0/users_ldap_push/models/res_company_ldap_field_mapping.py where users can configure arbitrary fields for arbitrary attributes.

alexey-pelykh:
@hbrunn I did think about making list of attributes via model initially, yet one step at a time :) Thanks for a hint and a code snippet!

auth_ldap_attribute_sync/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
```
hbrunn: for new modules, use [split readme files](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst)

auth_ldap_attribute_sync/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'LDAP Attribute Sync',
+    'version': '12.0.1.0.0',
+    'category': 'Tools',
+    'website': 'https://github.com/OCA/server-auth',
+    'author':
+        'Braibean Apps (https://brainbeanapps.com), '
```
hbrunn: this needs to be a python list, and I think you should remove the URL from the company's name

auth_ldap_attribute_sync/models/res_company_ldap.py
```diff
@@ -0,0 +1,127 @@
...
+            ('initial', 'On user creation'),
+            ('always', 'On every login')
+        ]
+
+    def _get_ldap_dicts(self):
+        res = super()._get_ldap_dicts()
+        for rec in res:
+            ldap = self.sudo().browse(rec['id'])
+            for field in self._fields_to_sync():
+                rec['%s_attribute' % field] = getattr(
```
hbrunn: no need for `getattr`, just use the [string index](https://github.com/OCA/OCB/blob/12.0/odoo/models.py#L5025) here too

auth_ldap_attribute_sync/models/res_company_ldap.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+
+        for field in self._fields_to_sync():
+            attribute = conf['%s_attribute' % field]
+            mode = conf['%s_attribute_mode' % field]
+            if not attribute or not mode:
+                continue
+
+            try:
+                fields[field] = ldap_entry[1][attribute][0]
```
hbrunn: this one I don't know, does python_ldap for python3 use unicode strings or byte strings? if the latter, you'll need to convert that

auth_ldap_attribute_sync/models/res_company_ldap.py
```diff
@@ -0,0 +1,127 @@
...
+
+        user = self.env['res.users'].sudo().browse(self.env.user.id)
+        for field in self._fields_to_sync():
+            attribute = conf['%s_attribute' % field]
+            mode = conf['%s_attribute_mode' % field]
+            if not attribute or mode != 'always':
+                continue
+
+            try:
+                setattr(user, field, ldap_entry[1][attribute][0])
```
hbrunn: see above

hbrunn (changes requested):
and we need some unittests

auth_ldap_attribute_sync/models/res_company_ldap.py
alexey-pelykh: As-is in `auth_ldap`, that's not an argument of course. As per [docs](http://www.python-ldap.org/en/latest/bytes_mode.html#the-bytes-mode), it's unicode

auth_ldap_attribute_sync/models/res_company_ldap.py
alexey-pelykh: Done

auth_ldap_attribute_sync/models/res_company_ldap.py
alexey-pelykh: Done

auth_ldap_attribute_sync/__manifest__.py
alexey-pelykh: Done

auth_ldap_attribute_sync/models/res_company_ldap.py
hbrunn: thank for looking this up

auth_ldap_attribute_sync/README.rst
alexey-pelykh: Done

alexey-pelykh:
@hbrunn unfortunately, using list in manifest broke the tools.gen_addon_readme usage, had to use string :(

hbrunn:
ah, it used to be a list, thanks for the update.

hbrunn (approved):
when there are tests

alexey-pelykh:
@hbrunn unrelated to this PR: is there a module that maps LDAP groups membership to Odoo permission groups that needs porting to 12.0, maybe you can give a hint?

alexey-pelykh:
```

**Comment by alexey-pelykh** [comment]
```
_logger.info('_get_ldap_dicts %s', self.env['res.company.ldap']._get_ldap_dicts())
with registry(self.env.cr.dbname).cursor() as cr:
    env = api.Environment(cr, SUPERUSER_ID, {})
    _logger.info('_get_ldap_dicts2 %s', env['res.company.ldap']._get_ldap_dicts())
```
while it is actually working, tests fail for some reason and at the moment I lack knowledge to understand what is going on

**Comment by alexey-pelykh** [comment]
@pedrobaeza I guess I need some hint: how to invalidate cache in all environments when recordset is being updated in a specific env?

**Comment by hbrunn** [comment]
cache: https://github.com/OCA/OCB/blob/12.0/odoo/api.py#L899

**Comment by hbrunn** [comment]
about groups: https://github.com/OCA/server-tools/tree/8.0/users_ldap_groups

**Comment by alexey-pelykh** [comment]
@hbrunn requesting help with tests, I'm yet too junior in Odoo to figure out why test doesn't see the changes made

File: auth_ldap_attribute_sync/tests/test_auth_ldap_attribute_sync.py [hunk 6]
```diff
@@ -0,0 +1,96 @@
...
+                'ldap_filter': '(uid=%s)',
+                'ldap_binddn': 'cn=bind,dc=example,dc=com',
+                'create_user': True,
+                'name_attribute': 'displayName',
+                'name_attribute_mode': 'initial',
+                'email_attribute': 'mail',
+                'email_attribute_mode': 'always',
+            })
+        
+        self.env['res.company.ldap']._patch_method(
```
hbrunn: better use [mock](https://docs.python.org/3/library/unittest.mock.html) here

hbrunn:
the [liniting errors](https://travis-ci.org/OCA/server-auth/jobs/444018658#L573) speak for themselves I guess, and for your test, check if you need to open another transaction given the login stuff opens its own cursor

alexey-pelykh:
Lint errors - yes, those are clear and minor. Regarding the transaction - please a bit more detailed here. Login opens own transaction, and I've tried both amending the user within it and within other cursor - yet from the test browsing the user gives same results - user is unchanged. That is why I've tried clearing cache yet so far I'm stuck there. While on actual instance it works as expected

hbrunn:
ah right, that's the trickery with [TestCursor](https://github.com/OCA/OCB/blob/12.0/odoo/sql_db.py#L426) - carve yourself one that doesn't do a rollback on close, then you should be fine

alexey-pelykh:
Oh, that explains everything! Yet how to disable or override it for the test? (I've googled prior asking)

alexey-pelykh:
Since as far as I've understood, that's for shielding one test from other ones?

hbrunn:
yes. Just inherit the class, override close to release the savepoint instead of rolling back, and pass this to the environment you run the new login in.

alexey-pelykh:
Even in case that I've tried to use
```
        with registry(self.env.cr.dbname).cursor() as cr:
            env = api.Environment(cr, SUPERUSER_ID, {})
            SudoUser = env['res.users'].sudo()
            SudoUser.browse(user.id).write(fields)
```
a TestCursor was used?

alexey-pelykh:
Please disregard, answer "yes" is in [sources](https://github.com/odoo/odoo/blob/12.0/odoo/modules/registry.py#L462) returns TestCursor.

auth_ldap_attribute_sync/tests/test_auth_ldap_attribute_sync.py
alexey-pelykh:
```
        CompanyLDAP = self.env['res.company.ldap']
        User = self.env['res.users']

        user_id = False
        with patch.object(
            CompanyLDAP,
            '_connect',
            lambda self, conf: FakeLdapConnection({
                'dc=example,dc=com': {
                    'cn': ['User Name'],
                    'displayName': ['User Name'],
                    'mail': ['user@example.com'],
                }
            })
        ):
            user_id = User.authenticate(
                self.env.cr.dbname,
                'username',
                'password',
                {}
            )
        user = User.sudo().browse(user_id)
```
does not quite work for me, is that what you meant?

alexey-pelykh:
@hbrunn sorry for bothering, yet the TestCursor extending did not help, and the `patch.object` for some reason didn't work for me as `_patch_method` :|

hbrunn:
post the code of how you tried to modify the test cursor, probably there's something going wrong there

alexey-pelykh:
Current code is in the PR - https://github.com/OCA/server-auth/pull/45/files#diff-3f471e5d67f959fbc51e6790f1328898R37

auth_ldap_attribute_sync/tests/test_auth_ldap_attribute_sync.py
hbrunn: check out this test, there they make a lot of use of the mock package: https://github.com/OCA/social/blob/11.0/mail_tracking_mailgun/tests/test_mailgun.py#L340

alexey-pelykh:
Seems this PR run out of checks? If that's even possible

pedrobaeza:
Today GitHub is being a nightmare, so it's possible a side effect:

https://status.github.com/messages

alexey-pelykh:
@hbrunn turns out TestCursor was not the reason, it was "execute each login within different cursor lifecycle", yet finally we're green to go further!

hbrunn:
great!

alexey-pelykh:
I guess now I should wait for more reviews?

hbrunn:
exactly. It's a problem that there's still runbot's warning, because this halts the CI workflow and many people don't bother looking at non-green PRs (me usually too). Any chance to patch the description into the model somehow?

alexey-pelykh:
Unfortunately, original model in `auth_ldap` is not documented properly, and no way to fix that from here - only a PR to odoo itself. :(

alexey-pelykh:
Runbot warnings are going to be present until https://github.com/odoo/odoo/pull/28047 is merged

hbrunn:
for the time being, try something like

```
from odoo.addons.auth_ldap.models.res_company_ldap import CompanyLDAP
CompanyLDAP._description = 'My desc'
```

in some `__init__` file, and add a comment why you do this, also linking to your patch

alexey-pelykh:
That sounds quite hacky, I think it would be better to wait for https://github.com/odoo/odoo/pull/28047 to be merged, hopefully that won't take ages.

hbrunn:
I can predict it will take ages if it's merged at all into 12

alexey-pelykh:
Injecting this stub into all LDAP-related modules still sounds horrible... Unless that patch get merged, I guess there will be no other option :|

alexey-pelykh:
@pedrobaeza fix was merged into the upstream, when OCB would be updated?

pedrobaeza:
Tonight

pedrobaeza:
@hbrunn don't be so pesimistic about Odoo PRs!

alexey-pelykh:
Super-green :)

alexey-pelykh:
@hbrunn when you've proposed to make the list dynamic - I did think about it again, yet one stopped me - how to get list of all fields.Char from the `res.users` as list?

sebalix:
You can do this:
```python
[f for f, d in self.env['res.users'].fields_get().items() if d['type'] == 'char']
```

alexey-pelykh:
image
@sebalix @hbrunn please review

auth_ldap_attribute_sync/views/res_company_ldap_views.xml
```diff
@@ -0,0 +1,21 @@
...
+    <record id="view_ldap_installer_form" model="ir.ui.view">
+        <field name="name">res.company.ldap.form</field>
+        <field name="model">res.company.ldap</field>
+        <field name="inherit_id" ref="auth_ldap.view_ldap_installer_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//group/field[@name='ldap_server']/../..">
+                <group string="Map User Attributes" >
+                    <field name="user_attributes_mapping" nolabel="1"/>
+                </group>
+            </xpath>
```
sebalix, line 17: Remark/question: there is no `position` keyword, does that mean the default is `after`? Or this code is buggy?

sebalix (approved):
Not tested, only reviewed code (with one question)  :+1:

auth_ldap_attribute_sync/views/res_company_ldap_views.xml
pedrobaeza: Default `position` is `inside`, so it seems correct.

sebalix: @pedrobaeza oh I didn't know, thank you!

alexey-pelykh: Right, `inside` is the default:
"the content of the inheritance spec is appended to the matched node" from [here](https://www.odoo.com/documentation/12.0/reference/views.html#inheritance-specs)

alexey-pelykh:
@pedrobaeza as far as I understand, needed approvals received and it should be merged?

PR #46, opened 2018-10-23
[12.0][MIG] users_ldap_groups

alexey-pelykh:


alexey-pelykh:
@hbrunn if you don't mind, please take a look

alexey-pelykh:
Frankly speaking, not much code left :|

alexey-pelykh:
Requesting help: for some reason using this module produces "Could not serialize access due to concurrent update" yet I'm not sure how to reproduce that in tests

alexey-pelykh:
@pedrobaeza if you don't mind, I'll bump this to your attention

pedrobaeza:
Well, I can't be very helpful here as I don't use this module and don't know about this area. What I can say is about the guidelines, as this module contains plurals, being dragged since long, but I won't force to change it if contributors don't want.

alexey-pelykh:
Frankly speaking, I'm thinking of making a bit revised (an more UX-friendly) one as an alternative

pedrobaeza:
Looking at the description, `auth_ldap_attribute_sync` can't be adapted for this purpose also?

alexey-pelykh:
It may, and actually they do share some code. I think that's up to reviewers if to keep modules lean-and-mean or to make complex ones. Why `auth_ldap_attribute_sync` appeared w/o groups - just because I had to start one-step-at-a-time.

pedrobaeza:
@hbrunn you were the original author of this. What do you think?

hbrunn:
I think it's great that @alexey-pelykh does all the efforts, thanks for that. Group stuff is very different from attributes like whatever is not a group, so I think this surely merits its own module.
I don't think being the original author of something entitles me of anything. So thanks for the consideration, but let's just discuss

alexey-pelykh:
At Brainbean Apps, all our role management is based on LDAP groups, which are matched to groups in Odoo. And takin into account that usually one LDAP group has multiple permissions (groups in Odoo), the resulting table in `users_ldap_groups` is very long and very hard maintain, e.g. 
image and there are 17 other lines like that, and I've mapped only like 20% of what's needed. So there are 2 suggestion where to improve:
1. UX: It seems to me that one LDAP group can safely be granted 1+ Odoo group via many2many - yet I haven't evaluated how to do that
2. Feature: add `contains-regexp` as operation
3. Somehow allow to strip `,cn=groups,cn=accounts,dc=bbapps,dc=org` from every line

pedrobaeza:
Can we rename this at least to `auth_ldap_group_sync` or similar if it can't be added to the other module?

alexey-pelykh:
@pedrobaeza I still plan to roll out my own vision on this feature soon, thus this PR is going to be a just-migration one. (And I plan to use exactly `auth_ldap_group_sync` name)

pedrobaeza:
OK, then I suppose we can dismiss this one on that moment, isn't it?

alexey-pelykh:
Well, I don't think it's the case - since someone will prefer to just use original module, plus I have no idea regarding the schedule - I do hope that it will be within this month. Though we may just keep this PR open, since anyhow the efforts have been invested :)

pedrobaeza:
Yes, of course. I don't think someone has the urgency for implementing this in v12, but well, here it is meanwhile.

alexey-pelykh:
I had 🥇 ))))

pedrobaeza:
Hehe, you are very brave going directly to v12. That's admirable. Integrators that are on Odoo since previous versions (like me) are moving more slowly. We plan to make first v12 implementation in January. We need to adapt localizations, usual OCA modules, and so on for doing this step. But with this version everything is smoother as good modules with tests and so on are very easy to migrate (unless you want to improve them, but most of them are already very good). I have migrated already several modules in Spanish localization in very low spare times: https://github.com/OCA/l10n-spain/issues/922

alexey-pelykh:
In my case, I basically had no choice: we're just trying to use Odoo in our daily operations and migrate from dozens of different services like BambooHR, BreezyHR, Tempo, Toggl to some single place, thus doing that using anything but latest Odoo version won't be beneficial, yet bugs sometimes give some pain.

To boldly go where no man has gone before™

pedrobaeza:
I agree with you and good that you have the correct vision (IMO). Other integrators are still starting projects in 8.0, condemning their customers to the obsolescence since beginning.

All my support and encouragement to you in your trip to v12!

alexey-pelykh:
So true re: 8.0 since integrator we've talked here locally also offered to start with 8.0, and I guess I know why - nobody wants this porting job

pedrobaeza:
But it should be part of the investment the integrator company makes on their business. If they don't do that, they won't be competitive in skills, knowledge and service offer. But well, again this is my PoV of this world.

alexey-pelykh:
@hbrunn so what are your thoughts on approving this PR?

users_ldap_groups/models/res_company_ldap.py
```diff
@@ -1,47 +1,121 @@
...
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 from logging import getLogger
-from odoo import api, fields, models
+from odoo import fields, api, models, registry, tools, SUPERUSER_ID
+
 _logger = getLogger(__name__)
 
+try:
+    import ldap
+    from ldap.filter import filter_format
```
hbrunn: ldap is a [standard requirement](https://github.com/OCA/OCB/blob/12.0/requirements.txt#L27) of odoo, so this is not strictly necessary

users_ldap_groups/models/res_company_ldap.py
```diff
@@ -1,47 +1,121 @@
...
+except (ImportError) as err:
+    _logger.debug(err)
+
 
-class ResCompanyLdap(models.Model):
+class CompanyLDAP(models.Model):
```
hbrunn: your rationale for renaming probably is that core calls the class like that, but at OCA, we have a tendency to use the model name in camel case: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#341odoo-python-classes

users_ldap_groups/models/res_company_ldap.py
```diff
@@ -1,47 +1,121 @@
...
+            ldap = self.sudo().browse(rec['id'])
+            rec['group_mapping_ids'] = ldap.group_mapping_ids or []
+            rec['only_ldap_groups'] = ldap.only_ldap_groups or False
+
+        return res
+
+    def _get_ldap_user(self, conf, login):
+        entry = False
+        try:
+            filter = filter_format(conf['ldap_filter'], (login,))
```
hbrunn: avoid using keywords as variable name

users_ldap_groups/models/res_users.py
```diff
@@ -1,27 +1,18 @@
...
-            # this case to apply ldap groups if necessary
-            ldaps = user.company_id.ldaps
-            if user.active and any(ldaps.mapped('only_ldap_groups')):
-                for conf in ldaps.get_ldap_dicts():
-                    entry = ldaps.authenticate(conf, login, password)
-                    if entry:
-                        ldaps.get_or_create_user(conf, login, entry)
-                        break
-        return user_id
+    def _check_credentials(self, password):
```
hbrunn: this function is called for every request made to odoo, cf https://github.com/OCA/OCB/blob/12.0/odoo/addons/base/models/res_users.py#L610, https://github.com/OCA/OCB/blob/12.0/odoo/service/security.py#L14, https://github.com/OCA/OCB/blob/12.0/odoo/service/model.py#L35

you don't want to do this on every request. `_login` is just fine

users_ldap_groups/models/res_company_ldap_operator.py
```diff
@@ -1,13 +1,14 @@
...
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 from logging import getLogger
 from odoo import api, models
 from string import Template
+
 _logger = getLogger(__name__)
 
 
-class ResCompanyLdapOperator(models.AbstractModel):
+class CompanyLdapOperator(models.AbstractModel):
```
hbrunn: see above

hbrunn (changes requested):
there's one big problem and a bunch of nitpicks.

I'm personally quite allergic to people making themselves authors during migrations. If a migration is effectively a rewrite, I'm okay with that, but in your case, I think this is a very welcome contribution, but doesn't merit authorship.

About your prospective extensions:

1. changing group_id to a many2many with widget many2many_tags should work with very little changes and nearly no UX impact I think
2. regex sounds nice, don't make it contains. Then people can either anchor their regexes or not, thus either checking a substring or the full string.
3. I think it's hard to make assumptions about that for the general case. There are very crazy ldap trees out there. You might add an option to always prepend the ldap base, but further I don't see much possibilities. You could get rid of the editable tree, have some custom summary field for the tree view and do the editing in a tree view

users_ldap_groups/models/res_company_ldap_operator.py
alexey-pelykh: done

users_ldap_groups/models/res_users.py
alexey-pelykh: thanks for a hint, done

users_ldap_groups/models/res_company_ldap.py
alexey-pelykh: done

users_ldap_groups/models/res_company_ldap.py
alexey-pelykh: done

users_ldap_groups/models/res_company_ldap.py
alexey-pelykh: done

hbrunn (approved):
thanks! for me, it's not necessary to reply on the different comments in case it's just done, that i see when re-reviewing the diff anyways ;-)

users_ldap_groups/models/res_company_ldap.py
```diff
@@ -0,0 +1,71 @@
...
+        return list(map(lambda x: (4, x, False), groups))
+
+    def _map_ldap_attributes(self, conf, login, ldap_entry):
+        values = super()._map_ldap_attributes(conf, login, ldap_entry)
+
+        _logger.info(
+            'Setting group membership from LDAP for login "%s"',
+            login
+        )
+        values['groups_id'] = self._get_user_groups(conf, ldap_entry)
```
hbrunn: this is much more elegant than my implementation!

alexey-pelykh:
Regarding authorship, I'd reply with a counter-argument that any amendment to the code effectively makes contributor a co-author, yet not an author per-se. Yet in this particular case the changes are trivial, so whatever comforts you is fine with me 👍

hbrunn:
yes, I totally agree. And another word for a co-author is a contributor :-)

alexey-pelykh:
Right, though technically then OCA is not an author for many modules, if to track the history. Yet I may not use that as an argument since I have no idea why OCA is listed in every module

hbrunn:
there's a practical issue: This way, your module stands out in the app store as being reviewed and approved by the OCA, many people see this as quality seal.
Then technically the OCA is an author in the sense that the CLA assigns some permissions to the code that usually only the author has, like changing the license of the whole.

pedrobaeza:
@alexey-pelykh that's because that was the only way to get all OCA apps classified in Odoo apps store. Odoo refused to add a category or a way to distinguish them, so we used that trick for that. Now that we are deploying our app "store" (still beta), it's more irrelevant:

http://odoo-community.org/apps

alexey-pelykh:
Oh well, it seems this migration conflicts with my own module, since `_map_ldap_attributes` is present in both, and it's conflicting. How to overcome that?

pedrobaeza:
Calling super is not enough?

alexey-pelykh:
It turns out that no, since super is called in both cases:
https://github.com/OCA/server-auth/pull/46/files#diff-4232d7711c0ba9787f714d14399adefaR63
and
https://github.com/OCA/server-auth/blob/7702cf93042fb07bb818355865390efe263ced53/auth_ldap_attribute_sync/models/res_company_ldap.py#L75
From logs I can see that "Initial setting field values from LDAP attributes for login" is printed while "Setting group membership from LDAP for login" is not, while it should

alexey-pelykh:
Any ideas?

pedrobaeza:
Everything in the code seems correct, and Odoo ORM preserves full call chain, although modules are not dependent each other. Are you sure there's no other thing that prevents this? Try several temp tests, like making this module dependent from the other one, or make the other one non installable to see if this still continues to not be called.

alexey-pelykh:
Thanks, will try! https://travis-ci.org/brainbeanapps/server-auth/builds/450777299 separately both modules work as expected, that's all I know as of now

alexey-pelykh:
Indeed was issue with the code and with tests

users_ldap_groups/models/res_company_ldap.py
alexey-pelykh: Unfortunately, this approach works only for newly-created users

alexey-pelykh:
Decided to keep the original behaviour, thus changed the PR

alexey-pelykh:
Good people of 10.0 PR (@sebalix @NL66278), I summon you in order to request a review of this PR as well, since you're aware of the topic 🔥

users_ldap_groups/tests/test_users_ldap_groups.py
```diff
@@ -0,0 +1,109 @@
...
+                    'ldap_binddn': 'cn=bind,dc=example,dc=com',
+                    'create_user': True,
+                    'only_ldap_groups': True,
+                    'group_mapping_ids': [
+                        (0, 0, {
+                            'ldap_attribute': 'name',
+                            'operator': 'contains',
+                            'value': 'hello3',
+                            'group_id': env.ref('base.group_system').id,
+                        }),
```
nikul-serpentcs, line 65: @alexey-pelykh IMO `env.ref('base.group_system')` use multipal time, so use like global variable

nikul-serpentcs (approved):
Minor change Otherwise LGTM

users_ldap_groups/tests/test_users_ldap_groups.py
alexey-pelykh: If it would be a `self.env.ref` call - surely, yet in this case env is unique in each case, thus technically reference _may_ be different

nikul-serpentcs: > If it would be a `self.env.ref` call - surely, yet in this case env is unique in each case, thus technically reference _may_ be different

@alexey-pelykh Then it's Ok

pedrobaeza: It doesn't matter if `env` is different, as the record will be the same, and more accessing only the `ID` field (it might be different from translatable or property fields). You can store the value in a variable and simply retrieve it on each loop.

alexey-pelykh: There are no loops or I didn't get it what do you refer to - the only place where I can place this variable to is `setUp`

pedrobaeza: I have only seen this snippet, and as it was said that it's used several times, I though it was a loop. Any way, the comment applies the same: it doesn't matter if env is different, the ID would remain the same in all the queries.

alexey-pelykh: Done :)

hbrunn:
@alexey-pelykh thanks a lot for this.

I would advise you to ping people as little as possible, we're all doing our best in reviews. In the best case, your ping doesn't create more time for people and is basically a no-op, in the worst case it creates noise.
I for example have a sieve rule that marks mails where I'm mentioned with the Imap flag important, because it's often people asking for help about technical stuff. So if people @mention me, that goes to the look-at-it-at-first-convenience folder. If I then see the same person pinging me repeatedly for reviews that are not on fire, I'll take their ping less seriously, with the extreme of ignoring them.
So I'd advise to just trust people will eventually come back for review, after a round of fixes a message in the veins of 'all done' creates a new (and for people like me less intrusive) notification, that seems fine to me. A reminder after a week of two seems fine to me too.

PR #48, opened 2018-10-25
[10.0][ADD] auth_keycloak

simahawk:
/cc @gurneyalex @guewen


auth_keycloak/models/res_users.py
```diff
@@ -0,0 +1,69 @@
...
+        return validation
+
+    @api.model
+    def _auth_oauth_validate(self, provider, access_token):
+        """Override to use authentication headers.
+
+        The method `_auth_oauth_rpc` is not pluggable
+        as you don't have the provider there.
+        """
+        oauth_provider = self.env['auth.oauth.provider'].browse(provider)
```
grindtildeath, line 44: If you have to browse, means you get an id, so using `provider_id` would make more sense here. Although passing the recordset might avoid the browse...

auth_keycloak/wizard/keycloak_sync_wiz.py
```diff
@@ -0,0 +1,296 @@
...
+    pwd = fields.Char(
+        related='provider_id.superuser_pwd',
+        readonly=True,
+    )
+    # tech field to map unique key from Keycloak to Odoo
+    login_match_key = fields.Selection(
+        selection=[
+            # keycloak:odoo
+            ('username:login', 'username'),
+            ('email:partner_id.email', 'email'),
```
grindtildeath, line 47: As it's a mixin shouldn't you define login and partner_id fields here ?

auth_keycloak/wizard/keycloak_sync_wiz.py
```diff
@@ -0,0 +1,296 @@
...
+            firstname = odoo_user.partner_id.firstname
+            lastname = odoo_user.partner_id.lastname
+        else:
+            # yeah, I know, it's not perfect... you can override it ;)
+            firstname, lastname = odoo_user.name.split(' ')
+        values.update({
+            'firstName': firstname,
+            'lastName': lastname,
+        })
+        logger.debug('CREATE using values %s' % str(values))
```
grindtildeath, line 252: Maybe add "Keycloak : ..." to know what it's about...

grindtildeath (commented):
Few questions looking at the code only

pedrobaeza:
@simahawk out of curiosity. Why do you prefer this over OpenLDAP/Oauth?

auth_keycloak/examples/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+```
+python get_token.py
+```
+
+If everything goes fine, you'll see something like:
+
+```
+$ python examples/get_token.py
+Username [c2c]:
```
tschanzt: maybe remove the c2c mentions in the docs

auth_keycloak/examples/cli.py
```diff
@@ -0,0 +1,164 @@
...
+        params['domain'], USERINFO_PATH.format(realm=params['realm'])
+    )
+    click.echo('Calling %s' % url)
+    headers = {
+        'Authorization': 'Bearer %s' % params['token'],
+    }
+    resp = requests.get(url, headers=headers)
+    if not resp.ok:
+        click.echo(resp.content)
+        click.echo('Something went wrong. Quitting. ')
```
tschanzt: in the other functions you also echo resp.reason. you forgot it here. Maybe even put it in its own function as its the same thing everytime

auth_keycloak/examples/cli.py
```diff
@@ -0,0 +1,164 @@
...
+    click.echo('Calling %s' % url)
+    resp = requests.post(
+        url,
+        data=data,
+        auth=(params['client_id'], params['client_secret']),
+        headers=headers,
+    )
+    if not resp.ok:
+        click.echo(resp.content)
+        click.echo('Something went wrong. Quitting. ')
```
tschanzt: no reason?

auth_keycloak/examples/keycloak-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+services:
+  keycloak:
+    image: jboss/keycloak
+    ports:
+      - 8080:8080
+    # change this if your db container is named differently
+    depends_on:
+      - db
+    environment:
+      # odoodb=# create user keycloak with password 'keycloak';
```
tschanzt, line 13: i probably wouldn't look for this here.

auth_keycloak/models/auth_oauth.py
```diff
@@ -0,0 +1,44 @@
...
+
+    client_secret = fields.Char()
+    users_endpoint = fields.Char(
+        help='User endpoint',
+        placeholder='http://keycloak.mycompany.com'
+                    '/auth/admin/realms/{realm}/users',
+        required=False,
+    )
+    superuser = fields.Char(
+        help='A super power user',
```
tschanzt: These help messages aren't really providing more information

auth_keycloak/models/res_users.py
```diff
@@ -0,0 +1,69 @@
...
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from odoo import api, models, exceptions, _
+import logging
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+class OAuthError(Exception):
```
tschanzt: maybe move this to a exceptions.py or something makes it much clearer for the next guy that it exists. because i wouldn't look for it here.

auth_keycloak/models/res_users.py
simahawk: I agree w/ you 100%. Fact is: I'm respecting odoo's [signature](https://github.com/odoo/odoo/blob/10.0/addons/auth_oauth/models/res_users.py#L39) :sob:

auth_keycloak/wizard/keycloak_sync_wiz.py
simahawk: uh? this is a mapping to match internal fields w/ external fields. What do you mean?

auth_keycloak/wizard/keycloak_sync_wiz.py
simahawk: you get that by the name of the logger ;)

simahawk:
@pedrobaeza in this customer case it's actually LDAP / AD who's feeding Keycloak.
It aggregates everything and provides unified login interface. See their [about page](url). :wink:

auth_keycloak/examples/keycloak-compose.yml
simahawk: it's already in docs, this is just a reminder... you WILL look at it if want to use this compose file as an example

auth_keycloak/examples/cli.py
simahawk: sometimes there's no reason... sometimes KC is not consistent w/ errors. 
Right now I don't remember why I did it like this... I'll double check :stuck_out_tongue_winking_eye:

gurneyalex (approved):

gurneyalex:
you're getting a fail on runbot because 

WARNING openerp_test odoo.addons.base.module.module: module auth_keycloak: description is empty !

could be related to the README generation which only happens later.

pedrobaeza:
Yes, that's it. I prefer to pre-generate README running in my local the script. This way, you also avoid surprises when the README is generated and maybe you get a syntax error in the composition (for example, not using proper title levels).

simahawk:
@pedrobaeza @gurneyalex done.
I've also documented how to get it here https://github.com/OCA/maintainer-tools/pull/392

auth_keycloak/wizard/keycloak_sync_wiz.py
grindtildeath: Actually considering https://github.com/OCA/server-auth/pull/48/files#diff-3ca04c3a0916ac2ee82f250785800931R110 I should have asked if you shouldn't define `username` and `email` here.
But having another look at it, if it's only to be used here, it makes sense.

grindtildeath (approved):

pedrobaeza:
@simahawk thanks for the improvement in the README!

simahawk:
squashed

tschanzt (approved):

auth_keycloak/wizard/keycloak_sync_wiz.py
```diff
@@ -0,0 +1,300 @@
...
+            'username': odoo_user.login,
+            'email': odoo_user.email,
+        }
+        if 'firstname' in odoo_user.partner_id:
+            # partner_firstname installed
+            firstname = odoo_user.partner_id.firstname
+            lastname = odoo_user.partner_id.lastname
+        else:
+            # yeah, I know, it's not perfect... you can override it ;)
+            firstname, lastname = odoo_user.name.split(' ')
```
grindtildeath: We can't guarantee the name includes a space, so we should find a way that it doesn't trigger an error here if name includes :
- no space
- only one space
- multiple spaces

grindtildeath (changes requested):
Considering https://github.com/OCA/server-auth/pull/49#discussion_r230381990 as well, we should fix this.

auth_keycloak/wizard/keycloak_sync_wiz.py
grindtildeath: Actually you provided the solution, so you can also implement it cherry-picking 21edcbf8b1cec8943f90790fb8df7510df796016 :sweat_smile:

simahawk: @grindtildeath done (included and squashed, was already coded in there actually ;))

simahawk:
@grindtildeath I've cherry-picked your doc improvement and improve test compat when you have partner_firstname installed. If you approve we are ready to merge :wink: 
Then you can hard reset and cherry-pick your commits on v9 + v8.

grindtildeath (approved):

PR #51, opened 2018-11-06
[10.0] add auth_api_key

qgroulard:
Authenticate http requests from an API key.
See readme for further information.


qgroulard:
@lmignon

sbidoul:
@qgroulard is covered by ACSONE ECLA.

auth_api_key/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,2 @@
+The api key must be provided by the coniguration parameter named
+'api_key_*.key', and the user under the same option 'api_key_*.user'.
```
sbidoul: An example would be helpful here.

auth_api_key/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Authenticate http requests from an API key.
```
sbidoul: Explain what an API key is and how one can use it.

auth_api_key/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,4 @@
+* Denis Robinet <<denis.robinet@acsone.eu>>
+* Laurent Mignon <<laurent.mignon@acsone.eu>>
+* Quentin Groulard <<quentin.groulard@acsone.eu>>
+* Sébastien Beau <<sebastien.beau@akretion.com>>
```
lmignon: email should be placed between single '<' and '>'

auth_api_key/readme/DESCRIPTION.rst
lmignon: IMO we must specify that the api key is passed via an HTTP request header (API-KEY). We must also warm the user that this kind of authentication mechanism must be used with caution. Indeed, information into HTTP headers are visible in clear. That means that this kind of authentication mechanism can only be used to authenticate requests from know sources (in a private network) and a good practice is to filter out this header at proxy level for unknown sources...

auth_api_key/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+
+{
+    "name": "Auth Api Key",
+    "summary": """
+        Authenticate http requests from an API key""",
+    "version": "10.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "website": "https://acsone.eu/",
+    "depends": ["server_environment"],
```
lmignon, line 14: @qgroulard Can you add the `development_status`key?  https://odoo-community.org/page/development-status @sbidoul can we qualify this addon with `Production/Stable` or `Beta`?

auth_api_key/models/ir_http.py
```diff
@@ -0,0 +1,37 @@
...
+            if uid:
+                # reset _env on the request since we change the uid...
+                # the next call to env will instantiate an new
+                # odoo.api.Environment with the user defined on the
+                # auth.api_key
+                request._env = None
+                request.uid = uid
+                request.auth_api_key = auth_api_key
+                return True
+        _logger.error("Wrong HTTP_API_KEY, access denied")
```
lmignon, line 37: @qgroulard It could be useful to also include the wrong `api_key`into the log message

lmignon (changes requested):
Thank you @qgroulard for this work and welcome into OCA!
The code is clean, with tests and doc. It's a great job and a great first contribution.
I made a few comments to aim for perfection. :smirk:

auth_api_key/__manifest__.py
sbidoul: @lmignon it certainly matches the criteria for Production/Stable (which are more or less the current OCA standards). So up to you and @sebastienbeau to decide if the configuration mechanism is stable enough to mark it stable. I hope it is.

auth_api_key/models/ir_http.py
lmignon: @qgroulard You can forget this comment. It could be dangerous to log the key. This error could occurs if an application try to log in the wrong env with a key valid in an other env...

auth_api_key/__manifest__.py
lmignon: I propose to declare this addon as 'Beta' for the next couple of weeks and change the status once validated in all our deployments.

lmignon (approved):
Thank you @qgroulard

lmignon:
@qgroulard Can you check and fix errors in travis...

auth_api_key/models/ir_http.py
```diff
@@ -0,0 +1,37 @@
...
+class IrHttp(models.AbstractModel):
+    _inherit = "ir.http"
+
+    @classmethod
+    def _auth_method_api_key(cls):
+        headers = request.httprequest.environ
+        api_key = headers.get("HTTP_API_KEY")
+        if api_key:
+            request.uid = 1
+            uid = request.env["auth.api.key"].retrieve_uid_from_api_key(api_key)
```
lmignon: @qgroulard `retrieve_uid_from_api_key`must start with '_'
```suggestion
            uid = request.env["auth.api.key"]._retrieve_uid_from_api_key(api_key)
```

auth_api_key/models/ir_http.py
```diff
@@ -0,0 +1,37 @@
...
+            if uid:
+                # reset _env on the request since we change the uid...
+                # the next call to env will instantiate an new
+                # odoo.api.Environment with the user defined on the
+                # auth.api_key
+                request._env = None
+                request.uid = uid
+                request.auth_api_key = auth_api_key
```
lmignon: @qgroulard `auth_api_key`is never defined.
```suggestion
                request.auth_api_key = api_key
```

lmignon (changes requested):

lmignon:
https://github.com/OCA/server-tools/pull/1425 should fix travis and runbot

sbidoul:
I restarted the build. Now it complains about the missing running_env config entry.

The solution is to [add it in travis-ci.yml](https://github.com/OCA/server-env/blob/b6ae38f48eb90eb6bb3b8e578b2ccdae9e2a11ed/.travis.yml#L34).

lmignon (approved):

lmignon:
@sbidoul done and :green_apple:

auth_api_key/models/ir_http.py
```diff
@@ -0,0 +1,38 @@
...
+
+class IrHttp(models.AbstractModel):
+    _inherit = "ir.http"
+
+    @classmethod
+    def _auth_method_api_key(cls):
+        headers = request.httprequest.environ
+        api_key = headers.get("HTTP_API_KEY")
+        if api_key:
+            request.uid = 1
```
sbidoul, line 25: why this line?

lmignon: We need to sppecify a user so the environment can be properly initialized on first access at the next line. https://github.com/odoo/odoo/blob/10.0/odoo/http.py#L252
This env is reset with the retrieved user just after the call to get the user from the api key.

sbidoul:
@lmignon @sebastienbeau @hparfr would you agree to change the license of this one to LGPL before merging? The reason is it's a low level technical module, LGPL allows to use it more different contexts.

lmignon:
@sbidoul  I agree. :+1:

hparfr:
Ok for me

sebastienbeau:
@sbidoul ok for me too. I also approve the change in this PR

sebastienbeau (approved):

lmignon:
Title changed to WIP since the licence will be changed

sbidoul:
License changed

lmignon:
Thank you @sbidoul

auth_api_key/models/auth_api_key.py
```diff
@@ -0,0 +1,37 @@
...
+    @tools.ormcache("api_key")
+    def _retrieve_uid_from_api_key(self, api_key):
+        if not self.env.user.has_group("base.group_system"):
+            raise AccessError(_("User is not allowed"))
+
+        for section in serv_config.sections():
+            if section.startswith("api_key_") and serv_config.has_option(
+                    section, "key"
+            ):
+                if api_key != serv_config.get(section, "key"):
```
ap-wtioit: the comparison should be done with consteq here
```suggestion
                if not odoo.tools.consteq(api_key, serv_config.get(section, "key")):
```

sbidoul:
@qgroulard can you check the travis errors?

gurneyalex (approved):

PR #56, opened 2018-11-20
[12.0][MIG] auth_api_key from 10 to 12

sbidoul:
Migration of #51 

TODO
- [x] change license to LGPL pending confirmation on #51


lmignon:
@sbidoul the last commits of #51 are missing and should fix travis
* [FIX] add oca dependencies https://github.com/OCA/server-auth/pull/51/commits/1aa57d83f44c2bac4fa5ecd6e7567fa0f9d9a38d
* [FIX] Travis: add [running_env] section into openerp_serverrc: https://github.com/OCA/server-auth/pull/51/commits/b55f6849e1b41964376058413c34a72a0cc14f64

lmignon:
@sbidoul The name of the migrated module is `auth_api_key` not `server_auth` :smirk:

sbidoul:
:green_apple:

lmignon (approved):

auth_api_key/models/auth_api_key.py
```diff
@@ -0,0 +1,37 @@
...
+
+        for section in serv_config.sections():
+            if section.startswith("api_key_") and serv_config.has_option(
+                    section, "key"
+            ):
+                if api_key != serv_config.get(section, "key"):
+                    continue
+
+                login_name = serv_config.get(section, "user")
+                uid = self.env["res.users"].search(
```
RobinetDenisAcsone, line 30: @sbidoul Is it possible to modify to search also the active=False users? 

Note: The suggestion below is 83 char (too long), but suggestion on multi lines isn't possible
```suggestion
                uid = self.env["res.users"].with_context(active_test=False).search(
```

RobinetDenisAcsone (commented):
Seem good to me, just a small request

auth_api_key/models/auth_api_key.py
sbidoul: I'm not sure. I kind of remember people (was it @guewen?) saying running code under an inactive user was a bad idea, but I'm don't remember why.

Cedric-Pigeon: @sbidoul I do not know on further versions but at least on V10.0 I have absolutely no troubles while running code with an inactive technical user....

sbidoul: @lmignon any opinion on this?

lmignon: @sbidoul IMO we must keep the possibility to select an inactive user (it's under the responsibility of the integrator/administrator to select the most appropriate one active or inactive). We must keep in mind that (at least in odoo < 12) active users are paid users. That's why we use inactive users  in such use cases.

gurneyalex:
@sbidoul #51 is merged, was there anything you needed to pick from that PR before this one gets merged?

gurneyalex (approved):

ap-wtioit:
@gurneyalex this addon depends on server_environment which is not yet migrated to 12.0 (or 11.0) in https://github.com/OCA/server-tools

i could add the code suggestion from our private repo where we removed the dependency if that is ok

auth_api_key/models/auth_api_key.py
guewen: > I'm not sure. I kind of remember people (was it @guewen?) saying running code under an inactive user was a bad idea, but I'm don't remember why.

The problem with inactive users is that they will be removed from the group that they are in as soon as you edit the user list on an ir.group. The m2m widget for users on the group will not include them - as they are inactive, so if you changes the list, it will remove it.

guewen: > I do not know on further versions but at least on V10.0 I have absolutely no troubles while running code with an inactive technical user....

Neither on other versions, until you edit the other users on a group and wonder why suddenly it doesn't work anymore ;)

sbidoul:
@ap-wtioit server_environment is merged in https://github.com/OCA/server-env

sbidoul:
@gurneyalex I've cherry-picked the consteq commit of @ap-wtioit, so this is now ready to merge.

lmignon:
Is everyone agreed to merge?

ychirino:
Blunt question: How many sessions are created by this implementation for 1.000.000 requests during a time span of 7 days? - I wonder if it's 1.000.000 or if I'm missing something...

sbidoul:
@ychirino In my experience, there is no sessions issue, as long as the client handles cookies properly. For python clients, using `request.Session()` is easy enough.

It might be worth mentioning in the readme of  `auth_api_key`, though.

ychirino:
@sbidoul Yeah, I wondered. Thanks. My main concern is a bearer token approach for an external API client. I can't guarantee in no way that it would stick to this pretty please recommendation. For more details see [this comment](https://github.com/OCA/rest-framework/issues/12#issuecomment-462967284).

ychirino:
Maybe a k/v map in memory assigning each token to a session? It's not a canonical solution but would reduce that problem by some orders of magnitude.
Assuming that workers power cycle every hour (7d * 24h * 5w * 10apikeys) 8.400 sessions vs 1.000.000 c.p.

sbidoul:
@ychirino multi-db is not really supported by this module at the moment.

Regarding the session, I'm not sure what problem you want to solve. It's easy to ask clients to return cookies.

ychirino:
@sbidoul There is a POC implementation of a bearer token [here](https://github.com/xoe-labs/openapi/blob/7940d625e25dcb95ee0936563bd142407942baca/controllers/pinguin.py#L188-L197).

The problem depends on what "easy" in "easy to ask clients to return cookies" actually means. On a publicized API It assumes cooperative clients. I'm not sure that's a valid assumption.

Or did I miss something here?

sbidoul:
@ychirino api keys and bearer tokens are for different use cases. api keys are meant for cooperating clients indeed. If you need bearer tokens that's for another module.

BTW, this is a migration PR, so please do not pollute it with unrelated comments.

auth_api_key/README.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+   !! This file is generated by oca-gen-addon-readme !!
+   !! changes will be overwritten.                   !!
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
+    :target: https://odoo-community.org/page/development-status
+    :alt: Beta
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
```
lmignon, line 13: I suppose that the licence badge will be fixed once the PR will be merged and the readme re generated

sbidoul: Yes

ychirino:
@lmignon Thanks for the clarification indeed, is a more extensive topic.

PR #61, opened 2018-11-27
[FIX] password_security: Make tests pass if website is installed

yajo:
[The `website` addon returns an aditional redirection][1] that makes these tests fail if ran after installing `website`.

The tests were checking the returned value in a funky way anyways.

Now, instead of checking the final returned value, we check directly the parameters sent to the redirection method.

[1]: https://github.com/odoo/odoo/blob/3b85900fafc9469dca6e7c01fca6dac4f55d20f5/addons/website/controllers/main.py#L85-L89

@Tecnativa

### DIFF
--- a/password_security/tests/test_password_security_home.py
+++ b/password_security/tests/test_password_security_home.py
@@ -206,30 +206,31 @@ class TestPasswordSecurityHome(TransactionCase):
 
 
 @mock.patch("odoo.http.WebRequest.validate_csrf", return_value=True)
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

yajo:
I can confirm integration tests passed. Let's see Travis, that is being quite slow today, I don't know why... :roll_eyes:

yajo:
:heavy_check_mark:

pedrobaeza:
OK, I think we can fast-track this as it's only affecting tests and they are green.

PR #63, opened 2018-12-04
[FIX] auth_signup_verify_email: Improve signup messages errors

zaoral:
* When try to login with a not valid email domain it raised a 500 server
  error. Now we shot the proper message "The domain name <domain> does not
  exist."

   Before | After
   --- | ---
   image | image

* If there is any error we do not have cached when we do the validation then
  we manage it and show the error message to the user instead of 500 server
  error.

* When we try to register with a already exist email it raise "Something went wrong, please try again later or contact us". Instead now show the message as odoo regular signup does: "Another user is already registered using this email address" which is more specific and clear for this case.

### DIFF
--- a/auth_signup_verify_email/controllers/main.py
+++ b/auth_signup_verify_email/controllers/main.py
@@ -9,7 +9,8 @@ from odoo.addons.auth_signup.controllers.main import AuthSignupHome
 _logger = logging.getLogger(__name__)
 
 try:
-    from email_validator import validate_email, EmailSyntaxError
+    from email_validator import validate_email, EmailSyntaxError, \
+        EmailUndeliverableError
 except ImportError:
     # TODO Remove in v12, dropping backwards compatibility with validate_email
     # pragma: no-cover
@@ -17,6 +18,9 @@ except ImportError:
         from validate_email import validate_email as _validate
 
         class EmailSyntaxError(Exception):
+            message = False
+
+        class EmailUndeliverableError(Exception):
             message = False
 
         def validate_email(*args, **kwargs):
@@ -52,6 +56,12 @@ class SignupVerifyEmail(AuthSignupHome):
                 _("That does not seem to be an email address."),
             )
             return request.render("auth_signup.signup", qcontext)
+        except EmailUndeliverableError as error:
+            qcontext["error"] = str(error)
+            return request.render("auth_signup.signup", qcontext)
+        except Exception as error:
+            qcontext["error"] = str(error)
+            return request.render("auth_signup.signup", qcontext)
         if not values.get("email"):
             values["email"] = values.get("login")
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I see several security problems with this. For example, for the point 3: you can get current users of the system trying possible logins until you get the message of e-mail address already in use. That's why the error message is so poor. It's for avoiding any of these problems.

zaoral:
> I see several security problems with this. For example, for the point 3: you can get current users of the system trying possible logins until you get the message of e-mail address already in use. That's why the error message is so poor. It's for avoiding any of these problems.

@pedrobaeza  this is not a custom behavior, it is exactly the way as Odoo manage an already used email as you can see in [these lines](https://github.com/odoo/odoo/blob/11.0/addons/auth_signup/controllers/main.py#L49-L51): this PR only try to maintain the same scheme. IMHO I think this is good info when users will like to know if you already has an account: is useful. 

if you like we can use search_count instead, and also create a Odoo PR to change the way that the main code manage this type of exception?

pedrobaeza:
Uhm, that's weird, as Odoo tends to avoid any information leak. Any way, I let my colleague to say, as he is the expert in this.

auth_signup_verify_email/controllers/main.py
```diff
@@ -9,7 +9,8 @@
 _logger = logging.getLogger(__name__)
 
 try:
-    from email_validator import validate_email, EmailSyntaxError
+    from email_validator import validate_email, EmailSyntaxError, \
+    EmailUndeliverableError
```
yajo: Indent this line and declare a dummy replacement for the cases when `validate_email` is used instead, as you can see below, to add the backwards compatibility support.

auth_signup_verify_email/controllers/main.py
```diff
@@ -70,10 +77,16 @@ def passwordless_signup(self, values):
...
             _logger.exception(error)
-
-            # Agnostic message for security
-            qcontext["error"] = _(
-                "Something went wrong, please try again later or contact us.")
+            if request.env["res.users"].sudo().search(
+               [("login", "=", qcontext.get("login"))]):
+                qcontext["error"] = _(
+                    "Another user is already registered using this email"
+                    " address.")
```
yajo, line 87: About the subject of security through obscurity, I'm generally against it.

You can see, @pedrobaeza, that I just mentioned your GitHub login name, so if I wanted to hack on your account I already know that; when I send you an email, I know your Google login name; when I read your tweets, I know your twitter login; etc.

FTR:

captura de pantalla de 2018-12-05 08-29-34

If you want actual security for your user accounts, there are other modules for that, so this is :+1:  IMHO.

yajo (changes requested):

auth_signup_verify_email/controllers/main.py
zaoral: Hi @Yajo 

Thank you for the review. I have already applied the changes you asked, please review them and let me know if there is something else needed.

Best Regards.

jjscarafia (approved):

auth_signup_verify_email/controllers/main.py
```diff
@@ -52,6 +56,12 @@ def passwordless_signup(self, values):
                 _("That does not seem to be an email address."),
             )
             return request.render("auth_signup.signup", qcontext)
+        except EmailUndeliverableError as error:
+            qcontext["error"] = _(str(error))
```
moylop260: You don't need translate the variables
I mean, you can delete the underscore method here `_()`

moylop260 (approved):
Just small comment
👍

auth_signup_verify_email/controllers/main.py
zaoral: Good catch.
Thanks @moylop260

zaoral:
Hi @Yajo @moylop260 

The requested changes have been applied.
Please let me know if there is something else missing in order to merge this PR.

Best Regards

yajo (approved):

PR #65, opened 2018-12-13
[FIX] auth_saml: method called with wrong number of arguments

gurneyalex:


gurneyalex:
I encountered the error while testing the 12.0 migration. Maybe something changed between 11.0 and 12.0 in the models classes. In any case https://github.com/OCA/server-auth/blob/11.0/auth_saml/models/auth_saml.py#L51 is expected to be called on a recordset. 

NOTE I don't have access to an infrastructure to test this -> if everything is fine, feel free to close this PR.

moylop260 (approved):

tfossoul:
Someone is still using this module ? Authentic2 is no more maintained still 2015, it is still in python 2.7.
And impossible to install Lasso lib which is needed to make the module work..

gurneyalex:
@tfossoul I've never been able to make this module run, and this is why I abandonned the PR....

gurneyalex:
as I said earlier I stopped trying to make the module work, and no longer have access to an infra for testing this. Installing lasso is quite painful...

tfossoul:
Mmm. Good to read ^^. So do we have alternative to enable sso with an active directory ? I head about keycload that do ldap in one side and oauth2 in other. But i'm a bit scared about this app, the fact it could be a monster to install/configure...

PR #71, opened 2019-01-23
[IMP] auth_session_timeout: Set the date on the session

etobella:
This PR is a refactoring of the module.
The goal of the refactoring is to provide a solution that could work on different SessionStores and shows a way to include more information on the session and manage the session store.


auth_session_timeout/hooks.py
```diff
@@ -0,0 +1,46 @@
...
+                    res = super().authenticate(
+                        db, login=login, password=password, uid=uid)
+                    if not self.update_time:
+                        self.update_time = time()
+                    return res
+
+            store.session_class = OpenERPSessionUpdate
+            return store
+
+    http.root = RootUpdate()
```
pedrobaeza, line 46: Why are you including this monkey-patching?

pedrobaeza (changes requested):
Hook is used unproperly. If you want to monkey-patch anything, you don't have to instantiate a class in the hooks.py file, but declare the hook on manifest and act on that method.

PR #72, opened 2019-01-25
[ADD] Add module for authentication with U2F

fkantelberg:
This module allows odoo to use FIDO U2F hardware tokens as a 2nd factor authentication.

It's tested with the U2F versions of Yubikey and Nitrokey and recent versions of Firefox and Chrome.

### DIFF
--- a/auth_u2f/__manifest__.py
+++ b/auth_u2f/__manifest__.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2017 Joren Van Onder
+# Copyright (C) 2019 initOS GmbH
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+{
+    'name': '2nd factor authentication via U2F',
+    'version': '11.0.1.0.0',
+    'author': "Joren Van Onder, "
+              "initOS GmbH, "
+              "Odoo Community Association (OCA)",
+    "summary": "2nd factor authentication via U2F devices",
+    'category': 'Extra Tools',
+    'license': 'LGPL-3',
+    'depends': ['base', 'web'],
+    'data': [
+        'views/u2f_device.xml',
+        'views/u2f_templates.xml',
+        'views/res_users.xml',
+        'security/ir.model.access.csv',
+        'security/auth_u2f_security.xml',
+    ],
+    'external_dependencies': {
+        'python': ['u2flib_server'],
+    },
+    'installable': True,
+}
--- a/auth_u2f/models/u2f_device.py
+++ b/auth_u2f/models/u2f_device.py
@@ -1,0 +1,73 @@
+# Copyright (C) 2017 Joren Van Onder
+# Copyright (C) 2019 initOS GmbH
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as
+# published by the Free Software Foundation; either version 3 of the
+# License, or (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful, but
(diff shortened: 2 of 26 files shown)


auth_u2f/views/u2f_device.xml
```diff
@@ -0,0 +1,55 @@
...
+                <tree>
+                    <field name="name"/>
+                    <field name="default"/>
+                </tree>
+            </field>
+        </record>
+        <record id="u2f_device_form" model="ir.ui.view">
+            <field name="name">u2f.device.form</field>
+            <field name="model">u2f.device</field>
+            <field name="context">{'default_user_id': uid}</field>
```
jorenvo: Context is not a field on view. This results in:

```
WARNING 11.0-u2f odoo.models: ir.ui.view.create() includes unknown fields: context
```

auth_u2f/models/u2f_device.py
```diff
@@ -0,0 +1,65 @@
...
+        string='Scan',
+        required=True,
+        help='Technical data returned by u2flib or the browser'
+    )
+    user_id = fields.Many2one('res.users', required=True, readonly=True)
+    default = fields.Boolean(help='Device used during login.', readonly=True)
+
+    @api.model
+    def create(self, vals):
+        vals['user_id'] = self.env.user.id
```
jorenvo: I'm not sure why this was added but I wouldn't do this. It prevents an administrator from registering U2F devices for employees which I think could be a common use case. Without this line the admin can go to the user form view and add a device in the U2F tab. With this line the device belongs to the admin instead.

jorenvo:
I tested on GNU/Linux with a Yubikey 4 and a Yubikey Security Key on Firefox 64.0.2 and Chromium 71.0.3578.98 and didn't detect any issues apart from the above.

fkantelberg:
@jorenvo Thanks for the tests and review. I fixed the mentioned things and created the unittests. Which means that the module is ready.

jorenvo:
@fkantelberg no problem, I like the added tests! Let me know if there's anything else I can do to help.

auth_u2f/README.rst
```diff
@@ -0,0 +1,75 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.png
```
mathiasfrancke: Maybe we can use the Readme By Fragments
* https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
* https://github.com/OCA/maintainer-tools/blob/master/README.md

auth_u2f/README.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Known Issues / Roadmap
+======================
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-auth/issues>`_. In case of trouble, please
```
mathiasfrancke: Why is there an underscore?

auth_u2f/__manifest__.py
```diff
@@ -0,0 +1,40 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA
+
+{
+    'name': '2nd factor authentication via U2F',
+    'version': '11.0.1.0.0',
+    'author': "Joren Van Onder, "
+              "initOS GmbH, "
+              "Odoo Community Association (OCA)",
+    "summary": "2nd factor authenticating via U2F devices",
```
mathiasfrancke: `2nd factor authentication via U2F devices`

mathiasfrancke (commented):
I successfully tested it with Yubikey Security Key and two Nitrokeys FIDO U2F

auth_u2f/static/src/js/auth_u2f.js
```diff
@@ -0,0 +1,124 @@
...
+            }
+        },
+    });
+
+    var FieldU2F = AbstractField.extend({
+        className: 'o_field_u2f',
+        supportedFieldTypes: ['char'],
+
+        init: function () {
+            var self = this;
```
mathiasfrancke: This should not be needed.

auth_u2f/README.rst
fkantelberg: Sounds like a good idea.

fkantelberg:
@jorenvo @mathiasfrancke Thanks for the reviews. The ci tests are running again due to a fix in auth_brute_force. Are you approving the PR?

jorenvo (approved):
Yes, I approve.

pedrobaeza:
@fkantelberg there are touched files in this PR that doesn't belong to the module you are mentioning. Why?

fkantelberg:
@pedrobaeza Good question. I think the last rebase broke everything

pedrobaeza:
Please try to rebase again with updated remotes.

fkantelberg:
Thanks. It should be fine now.

PR #75, opened 2019-02-22
auth_keycloak: improve scripts

simahawk:
Store default values on each run so to not type them over and over
for calls to the same service.

### DIFF
--- a/auth_keycloak/examples/get_token.py
+++ b/auth_keycloak/examples/get_token.py
@@ -2,49 +2,80 @@
 # pylint: disable=W7935, W7936, W0403
 
 import click
-from urlparse import urljoin
+import os
+try:
+    from urlparse import urljoin
+except ImportError:
+    from urllib.parse import urljoin
+
 import json
 
 from common import (
     UID, PWD,
     CLIENT_ID, CLIENT_SECRET,
-    DATA_FILE, DOMAIN, REALM,
+    TOKEN_DATA_FILE, DEFAULTS_DATA_FILE,
+    DOMAIN, REALM,
     GET_TOKEN_PATH,
     do_request
 )
 
+default_conf = {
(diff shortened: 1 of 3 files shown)


yvaucher (approved):

auth_keycloak/examples/get_token.py
```diff
@@ -2,49 +2,80 @@
...
+    'username': UID,
+    'password': PWD,
+    'client_id': CLIENT_ID,
+    'client_secret': CLIENT_SECRET,
+}
+
+if os.path.isfile(DEFAULTS_DATA_FILE):
+    click.echo('Loading defaults from %s' % DEFAULTS_DATA_FILE)
+    with open(DEFAULTS_DATA_FILE, 'r') as ff:
+        default_conf = json.loads(ff.read())
```
tarteo, line 34: It would be nice to overwrite only the values that exist in the json file.

tarteo (approved):
LGTM

auth_keycloak/examples/get_token.py
tarteo: Or check if all the keys exist in the json file.

hbrunn: is this comment a show stopper or should i merge?

tarteo: It's not a show stopper, you can merge.

simahawk:
@dreispt I think there's an issue w/ travis conf. To be fixed. I check later.

hbrunn:
I hade the same problem with a v10 PR that pulls server-tools via its requirements. And the only solution I came up with is https://github.com/OCA/project/pull/654/commits/0400673eb7169be8f3a9f3ba85c46c7ef3ed3d8b

simahawk:
fixed upstream, rebased

PR #84, opened 2019-04-10
[FIX] Fixed importing bug in keychain

ntsirintanis:
The `keychain` module defines a `fields.Serialized` [here](https://github.com/ntsirintanis/server-auth/blob/11.0/keychain/models/keychain_backend.py#L17), while the class itself is defined in [base_sparse_field](https://github.com/OCA/OCB/blob/11.0/addons/base_sparse_field/models/fields.py#L92).

[This](https://github.com/ntsirintanis/server-auth/commit/0af70752435fad736e0b136eed089b6470776025) commit introduced a static folder in `keychain` module. Consequently, upon modules loading and more specifically in [http.py](https://github.com/OCA/OCB/blob/11.0/odoo/http.py#L1340), `keychain` is loaded while `base_sparse_field` is not, so `fields.Serialized` in `keychain` remains undefined, and errors such as `AttributeError: module 'odoo.fields' has no attribute 'Serialized'` occur.

This PR corrects the bug that was introduced automatically by OCA-git-bot.

@hbrunn @NL66278 @gfcapalbo @daramousk


NL66278:
For me the fix does not work. I get the error anyway. I think the proper fix is to add base_sparse_field to the requirements in the manifest.

NL66278 (changes requested):
See other comments.

NL66278:
Anyway Serialized is not used independently, only as part of fields, so the import makes no sense anyway. Then I see that base_sparse_field is already a prerequisite, so apparently that does not work either.

hbrunn:
note that there's a similar problem in `mail_environment`/`server_environment`, and you'll only have a properly functioning web client if this is fixed too. (quick fix is to remove the static folders of those modules)

NL66278:
Would it be difficult to solve the root of the problem: the loading of not installed modules, only because there is a static folder?

hbrunn:
yes. Check @ntsirintanis' link to http.py - this doesn't know about databases and installed modules, only about the file system. I thought this was fixed in 13, but master's code looks the same for now.

hbrunn:
@sbidoul @mreficent (pinging you as author of the icon generator) @OCA-git-bot for addons that break just by being loaded python-wise, the automatically added icons are a problem. Can we have a flag in the manifest or file system to inhibit creating an icon?

Modules of course *should* at load in any circumstance via python, but the rewrite for server_environment to do that seems much more effort than what suffices here.

hbrunn:
@ntsirintanis check out the linting error please and mute it

MiquelRForgeFlow:
@hbrunn Interesting. There are always special cases like this one. Yes, no problem to add flags. What kind of flag do you propose in the manifest? Something like `'has_static'` that's True by default would be right? We could do that, easy peasy.

By the way, If I do that, could you help me with fixing https://github.com/OCA/openupgradelib/pull/142#issuecomment-478564332? I fix a thing helping you, and you fix a thing helping me. It's a one by one. What do you say? :)

hbrunn:
thanks for the pointer, I think my proposal fixes the issue. As for the flag: I'd give it the name of the script, as in

```python
if not manifest.get('gen_addon_icon'):
    return
```

MiquelRForgeFlow:
But the problem is unrelated to the `gen_addon_icon`. It's related to the having a static folder. So the manifest flag name should be something about admitting or not admitting a static folder. Imagine in the future adding another thing in @OCA-git-bot that also creates static folders... Or that someone migrates this module and adds a static folder manually. Who knows.

MiquelRForgeFlow:
I called it "preloadable" 😅
See https://github.com/OCA/maintainer-tools/pull/418.

MiquelRForgeFlow (approved):

hbrunn (approved):

MiquelRForgeFlow:
@ntsirintanis please, backport or forward port to all versions affected.

hbrunn:
@ntsirintanis and please do the same for `auth_api_key` in v10

PR #85, opened 2019-04-15
[FIX] 12.0 `password_security` module

shepilov-vladislav:
frontend meter/badgets/popup when using `password_security` module on signup/reset-password fronted pages.

Problems was solved by this PR:

1) Meter on frontend pages did works incorrectly.
2) Original meter from `auth_password_policy_signup` conflicts with meter from `password_security`. I did disable meter from `auth_password_policy_signup`.
3) Recommendations didn't worked correctly.

It's regression after migration to 12.0 and adding new features.

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -4,12 +4,15 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 import re
+import logging
 
 from datetime import datetime, timedelta
 
 from odoo import api, fields, models, _
 
 from ..exceptions import PassError
+
+_logger = logging.getLogger(__name__)
 
 
 def delta_now(**kwargs):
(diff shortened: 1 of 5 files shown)


shepilov-vladislav:
Guys who has experience with testing `qcontext`? Maybe anyone provide me a link with example?

pedrobaeza:
I'm afraid not

shepilov-vladislav:
> I'm afraid not

So, and how I can increase test coverage? Maybe it will be merged?

pedrobaeza:
Yes, codecov is not mandatory, so this has chances to be merged when reviewed.

BT-ojossen:
Hi, what's the state here?

shepilov-vladislav:
> Hi, what's the state here?

I found some more strange moments of behavior of this module.

1) password_minimum options doesn't affect on user' ability of password changing via reset password link

2) password_history value works with errors - old password quantity which are checked is <saved value>-1 (if I set 1 - the system allows to set any password including the current, if I set 2 - the system doesn't allow to set only the current password)

3) password_length and all others are displayed as 1 in hints if 0 value was saved
pass11

4) Reset password token expires at once after attempt to use old password (if History > 1)

5) Password Policy rules are checked firstly at password changing via user preferences. This is a slight vulnerability

Tell me, is it worth fixing them? Or I do not understand how this should work?

shepilov-vladislav:
@pedrobaeza Please suggest.

pedrobaeza:
Well, the problems you have found seems legit, so go ahead if you want to fix them, but if not, simply document them in known issues section

cuongnmtm (approved):

password_security/models/res_users.py
```diff
@@ -164,20 +167,39 @@ def _check_password_history(self, password):
...
+                recent_passes = user.password_history_ids[
+                    0: password_history - 1
                 ]
             if recent_passes.filtered(
                     lambda r: crypt.verify(password, r.password_crypt)):
-                raise PassError(
-                    _('Cannot use the most recent %d passwords') %
-                    rec_id.company_id.password_history
-                )
+                if recent_passes < 0:
```
amkarthik: @shepilov-vladislav 
```suggestion
                if password_history < 0:
```

password_security/models/res_users.py
```diff
@@ -164,20 +167,39 @@ def _check_password_history(self, password):
...
+                    error_msg = _('Cannot use passwords that were previously used.')
+                else:
+                    error_msg = _('Cannot use the most recent %d passwords') % recent_passes
```
amkarthik:
```suggestion
                    error_msg = _('Cannot use the most recent %d passwords') % password_history
```

shepilov-vladislav:
So old

PR #90, opened 2019-04-26
[12.0][MIG] Module auth_u2f

fkantelberg:
Hello,
I would like to share the migration of the module aufh_u2f to version 12.0.


fkantelberg:
@pedrobaeza I have no idea why the runbot is failing. I tested it locally without a problem and I can even access the runbot. Would you mind labelling it with needs_review?

pedrobaeza:
It's a runbot warning:
```
2019-06-22 06:39:04,018 172 WARNING openerp_test odoo.models: The model u2f.device has no _description
```

fkantelberg:
@pedrobaeza Thanks for the hint :+1:

auth_u2f/views/res_users.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data>
+        <record id="view_users_form_inherit_auth_u2f" model="ir.ui.view">
```
tbaden, line 4: optional change:
use original id (view_users_form)

auth_u2f/views/u2f_device.xml
```diff
@@ -0,0 +1,55 @@
...
+        <record model="ir.actions.act_window" id="action_open_u2f_device_tree">
+            <field name="name">U2F devices</field>
+            <field name="res_model">u2f.device</field>
+            <field name="view_mode">tree,form</field>
+            <field name="target">current</field>
+            <field name="context">{'default_user_id': uid}</field>
+            <field name="domain">[('user_id', '=', uid)]</field>
+        </record>
+
+        <record id="view_users_form_simple_modif_inherit_auth_u2f" model="ir.ui.view">
```
tbaden, line 43: optional change:
use original id (view_users_form_simple_modif)

tbaden (approved):
thanks for the work 👍 
I did a code review: LGTM

yucer:
It is great code.

Just one question about this [call](
https://github.com/initOS/server-auth/blob/a73a54c9a898c2bb718f3bd7763fb686571eec40/auth_u2f/models/http.py#L63)

Should one verify the previous result from super() before making the extra verification ?

If super fails, there is no need.

At least that the failed authentication trigger one exception.

fkantelberg:
If super fails it fails with an exception the result isn't really important in that case.

yucer (approved):
Clean code.

fkantelberg:
@pedrobaeza Can we merge this?

pedrobaeza:
Please solve the conflict.

fkantelberg:
@pedrobaeza 
Should work now.

PR #92, opened 2019-05-02
[FIX] change auth.api.key type from TransientModel to Model

mourad-ehm:
Change auth.api.key type from TransientModel to Model to can reference it in other models. 
ex : 
https://github.com/akretion/ak-odoo-incubator/blob/10.0/project_api/models/partner.py#L13

Bug AssertionError: Many2One relationships from non-transient Model to TransientModel are forbidden

### DIFF
--- a/auth_api_key/__manifest__.py
+++ b/auth_api_key/__manifest__.py
@@ -12,6 +12,6 @@
     "website": "https://acsone.eu/",
     "development_status": "Beta",
     "depends": ["server_environment"],
-    "data": [],
+    "data": ['security/ir.model.access.csv',],
     "demo": [],
 }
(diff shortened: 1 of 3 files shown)


mourad-ehm:
ping @sebastienbeau @lmignon

lmignon:
@mourad-ehm  I close this one since `auth.api.key` is no more a persistent model (not stored into database)... Data comes from config parameters.
You can see here https://github.com/akretion/odoo-shopinvader/blob/10.0/shopinvader/models/shopinvader_backend.py#L50 and here https://github.com/akretion/odoo-shopinvader/blob/10.0/shopinvader/models/shopinvader_backend.py#L290 how you can use it.

mourad-ehm:
Hi @lmignon, why don't define auth_api_key_name and user_id fields and related methods (auth_api_key_name, ....) into abstract class into auth_api_key modules ?

lmignon:
> Hi @lmignon, why don't define auth_api_key_name and user_id fields and related methods (auth_api_key_name, ....) into abstract class into auth_api_key modules ?

Indeed, that could be a good idea.

mourad-ehm:
Ok, I'll propose a pr for that,

PR #95, opened 2019-05-08
[ADD] auth_api_key: add an option to allow to use inactive users

Cedric-Pigeon:
In some cases we use inactive user for technical purposes. To avoid any login in the web interface, these users are defined inactive.
If we want to use them for an API, it is impossible.
This PR aims to solve this case

### DIFF
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -25,9 +25,15 @@ class AuthApiKey(models.TransientModel):
             ):
                 if not consteq(api_key, serv_config.get(section, "key")):
                     continue
+                user_model = self.env["res.users"]
+                if serv_config.has_option(section, "allow_inactive_user"):
+                    allow_inactive_user = serv_config.getboolean(
+                        section, "allow_inactive_user")
+                    if allow_inactive_user:
+                        user_model = user_model.with_context(active_test=False)
 
                 login_name = serv_config.get(section, "user")
-                uid = self.env["res.users"].search(
+                uid = user_model.search(
                     [("login", "=", login_name)]).id
 
                 if not uid:
--- a/auth_api_key/readme/USAGE.rst
+++ b/auth_api_key/readme/USAGE.rst
@@ -8,3 +8,9 @@ as value for the 'auth' parameter of your route definition into your controller.
         @route('/my_service', auth='api_key', ...)
         def my_service(self, *args, **kwargs):
             pass
+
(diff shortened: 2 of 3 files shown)


auth_api_key/models/auth_api_key.py
```diff
@@ -25,9 +25,15 @@ def _retrieve_uid_from_api_key(self, api_key):
             ):
                 if not consteq(api_key, serv_config.get(section, "key")):
                     continue
+                user_env = self.env["res.users"]
+                if serv_config.has_option(section, "allow_inactive_user"):
+                    allow_inactive_user = serv_config.get(
+                        section, "allow_inactive_user")
+                    if allow_inactive_user not in ['False', 'false', False]:
+                        user_env = user_env.with_context(active_test=False)
```
sbidoul: ConfigParser has [getboolean](https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.getboolean), which might be useful.

auth_api_key/models/auth_api_key.py
```diff
@@ -25,9 +25,15 @@ def _retrieve_uid_from_api_key(self, api_key):
             ):
                 if not consteq(api_key, serv_config.get(section, "key")):
                     continue
+                user_env = self.env["res.users"]
```
sbidoul: Strictly speaking it's more a user_model than a user_env.

lmignon (approved):
Thank you @Cedric-Pigeon LGTM

hbrunn (approved):

sbidoul (changes requested):
Could you update the README with some hints about the feature and how to use it?
Might be useful to also add [Guewen's comment](https://github.com/OCA/server-auth/pull/56#discussion_r240974549) about the risk of using inactive users in the USAGE section of the README.

lmignon:
@Cedric-Pigeon Can you take a look at Sbidoul's comment.

Cedric-Pigeon:
@lmignon @sbidoul Readme updated

rousseldenis (approved):
Great!

sbidoul:
@Cedric-Pigeon I think we should backport the latest changes from 12 before adding this.
In 12, there is now a table of api keys where the user is defined for each key (while the key itself can still come from server environment). So I the approach will be a bit different (and probably simpler to support inactive users).

rousseldenis:
@Cedric-Pigeon

baimont (approved):

PR #96, opened 2019-05-13
[MIG] auth_dynamic_groups to v10

hbrunn:
This module allows defining groups whose membership is a condition expressed as
python code. For every user, it is evaluated during login if she belongs to
the group or not.

Configuration
=============

Check `Dynamic` on a group you want to be dynamic. Now fill in the condition,
using `user` which is a browse record of the user in question that evaluates
truthy if the user is supposed to be a member of the group and falsy if not.

There is a constraint on the field to check for validity if this expression.
When you're satisfied, click the button `Evaluate` to prefill the group's
members. The condition will be checked now for every user who logs in.

Usage
=====

We have a group called `Amsterdam` and want it to contain all users from
city of Amsterdam. So we use the membership condition

```python
   user.partner_id.city == 'Amsterdam'
```

Now we can be sure every user living in this city is in the right group, and we
can start assigning local menus to it, adjust permissions, etc. Keep in mind
that view overrides can also be restricted by a group id, this gives a lot of
possibilities to dynamically adapt views based on arbitrary properties


hbrunn:
PS: It's enough to review https://github.com/OCA/server-auth/pull/96/commits/cfc3981a293bd47398c7c76ef03823cc4215984e, everything else already is in v8

auth_dynamic_groups/views/res_groups.xml
```diff
@@ -0,0 +1,27 @@
+<odoo>
+    <record model="ir.ui.view" id="view_groups_form">
```
tbaden: optional change:
id before model

auth_dynamic_groups/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2013-2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Dynamic groups",
+    "version": "10.0.1.0.0",
+    "author": "Odoo Community Association (OCA),Therp BV",
+    "license": "AGPL-3",
+    "complexity": "normal",
+    'summary': 'Have membership conditions for certain groups',
+    "category": "Tools",
```
tbaden, line 11: please add the website attribute with the value https://github.com/OCA/server-auth/

auth_dynamic_groups/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "depends": [
+        'base',
+    ],
+    "data": [
+        'views/res_groups.xml',
+    ],
+    "auto_install": False,
+    "installable": True,
+    "external_dependencies": {
+        'python': [],
```
tbaden: if you dont use the attribute, just delete it

NL66278 (approved):
:+1:  Minor comments, not blocking. Just curious for the motivation to move this from the server-tools repo to server-auth.

Also might still be helpfull to squash adjacent Transifex commits.

tbaden (changes requested):

tbaden:
thanks for the work 👍 
just some minor details

hbrunn:
@NL66278 @tbaden thanks, done

tbaden (approved):
thanks 👍

PR #99, opened 2019-05-16
[IMP] Handle xmlrpc Calls

osi-scampbell:
When trying to login via API call, the user variable being passed into the _update_user method returns False. When this happens the method uses the self.env.user variable to update the ldap_entry and the login is successful.


max3903 (approved):

patrickrwilson (approved):

max3903:
@osi-scampbell can you fix travis errors please?

Issue #100, opened 2019-05-20
[11] Password expiration cannot be disabled


javiermatos:
I have noticed password_security/models/res_users.py is different for versions 11 and 12 regarding the _password_has_expired function: in version 12, there is a condition that allows to disable password expiration just by setting password_expiration = 0 (which makes sense),

```
    @api.multi
    def _password_has_expired(self):
        self.ensure_one()
        if not self.password_write_date:
            return True

        if not self.company_id.password_expiration:
            return False

        days = (fields.Datetime.now() - self.password_write_date).days
        return days > self.company_id.password_expiration
```

while in version 11, the function does not accept a value just to disable it.

```
    @api.multi
    def _password_has_expired(self):
        self.ensure_one()
        if not self.password_write_date:
            return True
        write_date = fields.Datetime.from_string(self.password_write_date)
        today = fields.Datetime.from_string(fields.Datetime.now())
        days = (today - write_date).days
        return days > self.company_id.password_expiration
```

Can we "backport" that feature or is there any reason not to implement it that way in version 11? Will you accept a pull request of that feature to be merged in 11.0 branch?

pedrobaeza:
Please do the proper PR with the backport.

javiermatos:
Hola Pedro, I think the PR is done and ready to merge. Please, tell me in case there is something else I have to fix/improve for having this ready.

pedrobaeza:
OK, closing this one as there's a PR. Have you signed CLA?

PR #101, opened 2019-05-20
[11.0][IMP] Disable password expiration as in version 12.0 (backport)

javiermatos:
Implements fix for https://github.com/OCA/server-auth/issues/100

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -101,7 +101,9 @@ class ResUsers(models.Model):
         if not self.company_id.password_expiration:
             return False
 
-        days = (fields.Datetime.now() - self.password_write_date).days
+        write_date = fields.Datetime.from_string(self.password_write_date)
+        today = fields.Datetime.from_string(fields.Datetime.now())
+        days = (today - write_date).days
         return days > self.company_id.password_expiration
 
     @api.multi
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -97,9 +97,11 @@ class ResUsers(models.Model):
         self.ensure_one()
         if not self.password_write_date:
             return True
-        write_date = fields.Datetime.from_string(self.password_write_date)
-        today = fields.Datetime.from_string(fields.Datetime.now())
-        days = (today - write_date).days
+
+        if not self.company_id.password_expiration:
+            return False
+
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why aren't you cherry-picking the commit with the change? And I see that you are also doing some unneeded changes like introducing spaces in some docstrings, which you shouldn't (and more as that's not the way to put them).

javiermatos:
Thank you for the feedback @pedrobaeza. I have never used the cherry-picking feature from git (shame on me...). Anyway, I have located it here https://github.com/OCA/server-auth/commit/d683bf71f72e831a5b3af67fec6ba881291ca532 but I do not think we can just apply it as it introduces many changes for migrating to 12.0 that are not valid in our case. Am I wrong with that?

Regarding the spaces in docstrings, I introduced them as those spaces at the beginning and at the end of the docstrings are present in all the other docstrings for the tests of that module. I though the one that initially wrote the tests just forgot to add them.

I can revert the unneeded changes if you prefer that and just focus on the specific changes that are required.

(I have signed the CLA, so it is ready)

javiermatos:
Could we please integrate this into the 11.0 branch? I think everything is as it should, but if not, just tell me what should be fixed.

oihane (approved):
LGTM

password_security/models/res_users.py
```diff
@@ -97,6 +97,10 @@ def _password_has_expired(self):
         self.ensure_one()
         if not self.password_write_date:
             return True
+
+        if not self.company_id.password_expiration:
```
pedrobaeza, line 101: Where is this field defined?

javiermatos: The field is defined in password_security/models/res_company.py

```
class ResCompany(models.Model):
    _inherit = 'res.company'

    password_expiration = fields.Integer(
        'Days',
        default=60,
        help='How many days until passwords expire',
    )
    ...
```

pedrobaeza: OK, seen. I have mixed modules.

pedrobaeza (approved):

password_security/models/res_users.py
javiermatos: Great! Thank you so much! :)

PR #105, opened 2019-05-31
 [9.0][MIG] auth_saml 

tschanzt:
Migrated it from https://github.com/xcgd/server-auth/tree/11.0 as the original creator made some improvements not fed into the oca yet.


TDu:
@tschanzt Thanks for that, could you attend the couple of linting errors

PR #110, opened 2019-06-07
[11.0] auth_from_http_remote_user_role - Invalid roles if last_http_header_role mismatch with current roles

yvaucher:
There could be various reason last_http_header_role mismatch with role
lines.
Either the role were added or removed manually on the user or the role
codes could have been changed.

In worst cases in can lead to a user not able to recover his login.
Because the presence of last_http_header_role it will bypass the
role assignation and will fail with an error 500 due to access rights
on the first thing it tries to read.

A different approach would be to change the field last_http_header_role itself as the computed field. Not sure why it was done like this.

@guewen @TDu what is your opinion on this?


guewen:
@yvaucher did you have the opportunity to try the alternative we discussed? FTR it was to empty the last_http_header_roles when roles of a user change, which simplifies the logic IMO. It is hard to understand why we need a second field to compare with, as the only thing which matters is to "invalidate" the `last_http_header_role` value.

> A different approach would be to change the field last_http_header_role itself as the computed field. Not sure why it was done like this.

`last_http_header_role` stores the exact header received by the proxy, so it cannot be computed.

Note: if it wasn't clear, the goal of `last_http_header_role` is to store the last header received by the SSO proxy so we know if roles have been changed *outside* of Odoo. Any time the SSO sends different roles, we'll update them in Odoo. In this context, roles **should never be changed manually in Odoo**, but if it happens, then emptying `last_http_header_role` will force a re-evaluation of the SSO roles the next time we receive an HTTP request and reset the roles to what we expect from the SSO.

yvaucher:
@guewen Fixed the last lint issue of the PR.

I rewrote this a while ago without the additional computed field on user following your last comment.

guewen (approved):
Thanks!

PR #117, opened 2019-07-09
[10.0] [WIP] [NEW] Module auth_dynamic_groups_contract.

NL66278:
Add users to groups based on their active contracts.


auth_dynamic_groups_contract/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+We iwant to have a group called `Our Newsletter` that should contain all
```
hbrunn: typo

auth_dynamic_groups_contract/models/res_groups.py
```diff
@@ -0,0 +1,32 @@
...
+        vals = self._check_contract_based(vals)
+        return super(ResGroups, self).create(vals)
+
+    def _check_contract_based(self, vals):
+        """Set dynamic condition for when groups are based on contract."""
+        contract_based = vals.get('is_contract_based', False)
+        # If contract based set fixed condition.
+        if contract_based:
+            vals['dynamic_group_condition'] = \
+                "user.has_group_contract(dynamic_group)"
```
hbrunn: as the UI doesn't force the user to have `is_dynamic` checked, I'd propose to set this flag here unconditionally

auth_dynamic_groups_contract/views/res_groups.xml
```diff
@@ -0,0 +1,18 @@
+<odoo>
+
+    <record id="view_groups_form" model="ir.ui.view">
+        <field name="inherit_id" ref="auth_dynamic_groups.view_groups_form" />
+        <field name="model">res.groups</field>
+        <field name="arch" type="xml">
+            <field name="is_dynamic" position="after">
+                <field name="is_contract_based" />
```
hbrunn: or make this only visible when `is_dynamic` is checked

auth_dynamic_groups_contract/views/res_groups.xml
```diff
@@ -0,0 +1,18 @@
...
+            </field>
+            <field name="is_dynamic" position="attributes">
+                <attribute
+                    name="attrs"
+                    >{'readonly': [('is_contract_based', '=', True)]}</attribute>
+            </field>
```
hbrunn: I'd suggest to make the condition also readonly, we don't want the user to fiddle with the condition then

hbrunn (changes requested):
I think there's arguments to put this into the contract repo, but also for putting it here

NL66278:
@hbrunn I took up your requests, but introduced more changes also in the base module, to make tests pass and improve performance.

Still will do some more changes, so no need to merge now, even if you are satisfied with the changes so far...

NL66278:
This module is replaced by the more extensive (and flexible) code here: https://github.com/OCA/server-auth/pull/118

PR #120, opened 2019-07-18
[12.0][IMP] ADD possibility to save api keys in database by using server_env

mourad-ehm:
ADD possibility to save api keys in database by using server.env.mixin obj

TODO
- [ ] add key generation
- [ ] option for inactive/active user
- [x] fix test

### DIFF
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -1,7 +1,7 @@
 # Copyright 2018 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo import api, models, tools, _
+from odoo import api, fields, models, tools, _
 
 from odoo.tools import consteq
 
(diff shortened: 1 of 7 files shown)


mourad-ehm:
@sebastienbeau

sebastienbeau:
@sbidoul I am going to finish that PR today and try to make it work in multi-db

sebastienbeau:
@sbidoul I am going to fix the test, but this code is working in multidb \o/ !

sbidoul:
@sebastienbeau how does it work in multi-db? You need to add `?db=...` in the URL?

Do we need to add the new api key records in the database when upgrading to this PR? Is there a need for a UI for these records?

sebastienbeau:
@sbidoul natively odoo support multi-db for the controller but we need to filter correctly the database based on the domain name. 
You can test to have 2 domain name and add "dbfilter = ^%d$" (doc is here: https://www.odoo.com/documentation/10.0/setup/deploy.html)
See db_filter implementation here : https://github.com/odoo/odoo/blob/12.0/odoo/http.py#L1507

If we want to support multi-db but with the same domain name, I think we should do it in a separated module as the problematic is generic for every controller.


Then depending on the domain odoo will fetch the right db and the controller will work correctly.

My PR allow to have key per database and do not mix config between database, but it can be done differently.
I choose to add a model in order to follow the new implementation of server-env, that allow to configure even if you do not have server-env file (which is great for demo server)
More over as I depend on the implementation of server env, I am out-of-box compatible with https://github.com/OCA/server-env/pull/23/files that replace keychain and extend server-env

sebastienbeau:
@Cedric-Pigeon @lmignon @sbidoul we should validate this proposal of new architecture before finishing the PR.

auth_api_key/models/auth_api_key.py
```diff
@@ -3,36 +3,64 @@
...
+        base_fields = super()._server_env_fields
+        api_key_fields = {
+            "key": {},
+            "user": {},
+        }
+        api_key_fields.update(base_fields)
+        return api_key_fields
+
     @api.model
     @tools.ormcache("api_key")
```
lmignon: The result of this method can't be put into ormcache since it's not allowed to put a recordset into the cache....

lmignon (changes requested):
I've to look to the new implementation of server_env...before validating this PR
Nevertheless, what's the name of the sections to add into the config file for each auth.api.key configured into my server.
On a security POV, I'm not comfortable to know that it could be possible to access to the api keys from XMLRPC...

sebastienbeau:
The section name is the following : https://github.com/OCA/server-env/blob/12.0/server_environment/models/server_env_mixin.py#L38

Regarding XMLRPC. What we can do is adding a layer of protection that block reading this value and only allow in special case (explicitely in the code)?

lmignon:
@sebastienbeau The code is fixed and tests and doc adapted. I've created a PR on your branch https://github.com/akretion/server-auth/pull/1

lmignon:
ping @sbidoul @Cedric-Pigeon @rousseldenis

Cedric-Pigeon (approved):
Great work! Thanks. Code review OK.

auth_api_key/models/ir_http.py
```diff
@@ -22,16 +22,16 @@ def _auth_method_api_key(cls):
         api_key = headers.get("HTTP_API_KEY")
         if api_key:
             request.uid = 1
-            uid = request.env["auth.api.key"]._retrieve_uid_from_api_key(
-                api_key)
-            if uid:
+            api = request.env["auth.api.key"]._retrieve_api_key(api_key)
```
sbidoul, line 25: s/api/api_key/ ?

sbidoul (changes requested):
As discussed, override `_server_env_section_name` so the section name in config remains the same, and write a migration to initialize the records.

lmignon:
@sbidoul @Cedric-Pigeon superseded by #135

PR #122, opened 2019-07-22
[UPD] oca_dependencies

eilst:


pedrobaeza:
That's a comment for indicating why we put that repo. Why do you want to remove it?

eilst:
Hey, I'm trying to fix the runbot build problem. It fails and I found that after this commit https://github.com/OCA/server-auth/commit/dbca5bfcb7dd252ebbcb239981a31a41e41f7b59 it started. I thought this might have been the problem.

pedrobaeza:
I don't think that's the problem...

Put an issue with the problem you face and then people can give you a tip.

eilst:
OK, I will do that and also try to figure it out. Thanks

eilst:
It seems the problem is in server-backend/requirements.txt'

  subprocess.CalledProcessError: Command '['pip', 'install', '--no-binary', 'pycparser', '-Ur', '/home/odoo/dependencies/server-backend/requirements.txt']' returned non-zero exit status 1

requirements.txt
```diff
@@ -8,4 +8,4 @@ pyotp
 # auth_u2f
 python-u2flib-server
 responses
-
+libmysqlclient-dev
```
eilst:
```suggestion
```

eilst:
Solution: https://github.com/OCA/server-auth/pull/123

PR #123, opened 2019-07-22
[11.0][FIX] missing package in travis.yml

eilst:
Added - libmysqlclient-dev # To install mysqlclient and import MySQLdb and fix missing dependency

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml # because pip installation is slow
+      - libmysqlclient-dev # To install mysqlclient and import MySQLdb
 
 env:
   global:


eilst:
Hey @pedrobaeza, you were right that the problem was not in oca_dependencies.txt. It's here where the package was missing. So the tests on the runbot should pass

rimbendhaou (approved):

wbeverly (approved):

lgharib (approved):

pedrobaeza:
What is the module that requires mysqlclient?

eilst:
It's
base_user_role,  added by this commit: https://github.com/OCA/server-auth/commit/dbca5bfcb7dd252ebbcb239981a31a41e41f7b59
It's an Ubuntu package.

eilst:
By adding that dependency, runbot succeded. Right now is red on 11.0

pedrobaeza:
Uhm, I don't think so. https://github.com/OCA/server-backend/tree/11.0/base_user_role doesn't require MySQL AFAIK

pedrobaeza:
And the test you can do is that it's working in current runbot without this PR. I think this is something on your own runbot, but not here.

eilst:
It fails on branch 11.0 https://github.com/OCA/server-auth/tree/11.0
oca-red-travis

eilst:
There is no problem with 12.0. The module that needs this package is:  auth_from_http_remote_user_role which has not yet been migrated to 12.0

pedrobaeza:
OK, it's a weird side effect, and I suppose it happens with https://github.com/OCA/server-backend/tree/11.0/base_external_dbsource_mysql module, but let's fix it this way.

Issue #119, opened 2019-07-15
[11.0] password_security - redirection fails when password expired and website installed


Martronic-SA:
It seems that sometimes the method in the website module controllers/main.py web_login is overriding the one in password_security and then if the password has expired, user is not redirected on the signup_url but on the standard /web page and therefore cannot login anymore.
How to be sure that password_security is on the top? should it be added to the server_wide_modules in conf? (it seems that it doesn't help..)

amkarthik:
@Martronic-SA 
Please review https://github.com/OCA/server-auth/pull/124.
It will fix this issue.

PR #124, opened 2019-07-23
[FIX] [11.0] password_security - redirection fails when password_secu…

amkarthik:
This PR fixes the redirection to reset password page when the logging in user's password is expired.

'login_success' param is set in request params when the user successfully logged in. This is introduced in the commit https://github.com/odoo/odoo/commit/a1fac87d2509498148d8649f0eb703238e86a4e2 to redirect the user to backend or front end based on their access rights.

As we kick out the user when the logging in user's password is expired, we can set the 'login_success' param to False. This will avoid the redirection in the website module controller and will redirect us to password reset page.

This PR fixes https://github.com/OCA/server-auth/issues/119.

### DIFF
--- a/password_security/controllers/main.py
+++ b/password_security/controllers/main.py
@@ -43,6 +43,8 @@ class PasswordSecurityHome(AuthSignupHome):
         # My password is expired, kick me out
         request.env.user.action_expire_password()
         request.session.logout(keep_db=True)
+        # I was kicked out, so set login_success in request params to False
+        request.params['login_success'] = False
         redirect = request.env.user.partner_id.signup_url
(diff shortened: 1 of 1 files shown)


amkarthik:
PR https://github.com/OCA/server-auth/pull/125 is for v12.0

amkarthik:
@pedrobaeza 
Please review this.

pedrobaeza:
Sorry, but it's out of my scope.

amkarthik:
> Sorry, but it's out of my scope.

Thanks for the reply.

atchuthan (approved):

Martronic-SA (approved):
Perfect!

Issue #21, opened 2018-05-25
[11.0][auth_admin_passkey] #GDPR Is auth_admin_passkey an ilegal module? Should OCA have such a module?


rafaelbn:
Hello,

https://github.com/OCA/server-tools/tree/10.0/auth_admin_passkey

We are not using this module at all, but I think is a good idea yo open this issue and see what Contributors thinks.

Im my opinion this module shouldn't be migrated to v11 inside an OCA repository.

Thanks for other point of view

pedrobaeza:
No feedback, so closing. Don't put that module if you think is not legal. Legal thing depends on your country, not on the code.

legalsylvain:
@rafaelbn ! 
continuing the migration of modules to v12, I took a look on this issue you wrote.

I think it's a good point.
I developped the module ``auth_admin_passkey`` as an "admin" feature, to have the possiblity to log in the erp, to see what my users are seing. In fact, some bug, trouble, ACL limitation etc... depends on the context.
I think that this module is usefull regarding this aspect.

In the other hand, for the time being, it grants this power to the ERP admin user. I think that we could change that, and create a new key in the ``openerp.cfg`` config file, named ``passkey_password = xxxx``.

1) I assume that somobody that has access to that file, has access also to the server, so to the database. So, he can read all the datas, and can change create_uid / write_uid. (In fact admin sys has all the power).

2) If an ERP admin user try to install this module, but if he doesn't have access to that file, it will not be possible for him to use this feature.

I'm not sure if I'm clear.

are you agree with such design ? I see it as more "GDPR spirit".

rafaelbn:
I agree with such design @legalsylvain :+1:

PR #126, opened 2019-08-09
[12.0][MIG+REF] auth_admin_passkey

legalsylvain:
Port ``auth_admin_passkey`` from 10.0 to 12.0

**Main changes**
- repository (from server-tools to server-auth). Ref : https://github.com/OCA/server-tools/pull/1624#issuecomment-513891894
- configuration is now done via ``odoo.cfg`` config file, and not by ``ir.config.parameter`` settings. (more secure, more GDPR, and more simple code). [Ref](https://github.com/OCA/server-auth/issues/21). 
- various OCA convention

CC : @pedrobaeza, @alexis-via, @rvalyi, @eugen-don, @rafaelbn 

**One question to @Fenkiou**
 
You introduced the following code in that [commit](https://github.com/OCA/server-tools/commit/e16b37ae6d9c1cf71915d5a34a9b3651d0a23d12#diff-8ec5d043927ada610c80d69e6804860dR88) (PR [to port the module in V10](https://github.com/OCA/server-tools/pull/776)). I did'nt get this part. (I keep it, but I don't see it necessary).

            # Just be sure that parent methods aren't wrong
            user = self.sudo().search([('id', '=', self._uid)])
            if not user:
                raise


rafaelbn (approved):
nice work @legalsylvain

Olivier-LAURENT:
Tested... 👍 . Thx for the work @legalsylvain

legalsylvain:
Thanks for the reviews, guys.

PR #127, opened 2019-08-10
[12.0][ADD] ldap_sync: sync user directory with LDAP

alexey-pelykh:
Some modules (like `connector_jira`) require users to exist even if they have never logged in or never will, so we need a module to sync users list with LDAP server.


hbrunn:
are you aware of https://github.com/OCA/server-tools/tree/10.0/users_ldap_populate ?

alexey-pelykh:
@hbrunn yes, of course. Though it does not quite do what I have in mind, and unfortunately I see no value in maintaining that one (for me).

ldap_sync/data/cron.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version='1.0' encoding='UTF-8' ?>
+  <!--
+      Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+      License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+  -->
+<odoo>
+
+    <!-- <record id="ldap_sync_cron" model="ir.cron">
```
hbrunn: better uncomment and set as inactive, and make it noupdate

ldap_sync/models/res_company_ldap.py
```diff
@@ -0,0 +1,175 @@
...
+        self.ensure_one()
+        # 1. query all users from LDAP
+        # self._query({ ldap_base: users_dn }, user_object_filter)
+
+        # 2. for all users that are new in LDAP results, look same in existing database
+        #     - link them
+
+        # 3. for each of existing user, by ldap unique ID, select from ldap results and:
+        #     - update if exists: attributes, group membership
+        #     - deactivate if no longer exists and remove link to LDAP
```
hbrunn: is something missing here?

hbrunn (dismissed):

ldap_sync/models/res_company_ldap.py
alexey-pelykh: Yes, still WIP, should be done in an hour or so

ldap_sync/data/cron.xml
alexey-pelykh: 👍

hbrunn:
fair enough. But there is some overlap with `users_ldap_mail` and `users_ldap_groups`, do you think you can use those modules?

alexey-pelykh:
@hbrunn these module can and should coexist, both `users_ldap_mail` and `users_ldap_groups` update user on login sequence, which is super important, while this module does that for users that may never actually log-in.

ldap_sync/models/res_company_ldap.py
alexey-pelykh: Well, I underestimated amount of work, but fine now :)

alexey-pelykh:
@hbrunn ready for review! :)

ldap_sync/data/cron.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version='1.0' encoding='UTF-8' ?>
+  <!--
+      Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+      License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+  -->
+<odoo>
+
+    <record id="ldap_directory_sync_cron" model="ir.cron">
```
hbrunn, line 8: I still think this should be noupdate

ldap_sync/models/res_company_ldap.py
```diff
@@ -0,0 +1,345 @@
...
+    )
+    user_login_attribute = fields.Char(
+        default='uid',
+    )
+    user_email_attribute = fields.Char(
+        default='mail',
+    )
+    user_name_attribute = fields.Char(
+        default='cn',
+    )
```
hbrunn, line 33: how is this supposed to play nice with a migration of users_ldap_mail? The user will have to fill in those fields for both modules

ldap_sync/models/res_company_ldap.py
```diff
@@ -0,0 +1,345 @@
...
+                    (0, False, values)
+                )
+        missing_user_ids = self.ldap_user_ids - existing_user_ids
+        for ldap_user_id in missing_user_ids:
+            odoo_user_id = ResUsers.search(
+                [
+                    ('ldap_user_id', '=', ldap_user_id.id),
+                    '|',
+                    ('active', '=', True),
+                    ('active', '=', False),
```
hbrunn: do this kind of this with `.with_context(active_test=False)`

hbrunn (commented):
A few small things. Overall, I still think you could have done the same thing with a combination of the existing modules mentioned plus some extra code for your use case. That's why I won't approve this PR, but I also don't block it so that you can go on with other reviewers if they disagree with my assessment

ldap_sync/data/cron.xml
alexey-pelykh: Probably valid point for case when this setting is actually modified and we don't want this to be altered on upgrade. Adjusted

ldap_sync/models/res_company_ldap.py
alexey-pelykh: If you're suggesting to make this module as successor of `users_ldap_mail` and thus provide a migration script, I'd appreciate help with that as my experience with writing migration scripts is very limited.

ldap_sync/models/res_company_ldap.py
alexey-pelykh: Thanks! Something new every day :)

ldap_sync/models/res_company_ldap.py
hbrunn: ignore this one, that's more connected to my claim above that I think this module should depend on migrations of `users_ldap_groups`, `users_ldap_mail` and `users_ldap_populate`

alexey-pelykh:
> I still think you could have done the same thing with a combination of the existing modules mentioned plus some extra code for your use case.

That is partly true as I tried that approach initially, but the actual results and configuration was quite ugly, and the modules mentioned tend to be designed to handle on-demand sync instead of periodic one. Initially I'm after Jira/Confluence/Bitbucket-alike approach, there's a User Directory sync process that populates the user database periodically. So, by evaluating mixture of those reasons and fact of having a spare day, here's the result :)

ldap_sync/models/res_company_ldap.py
alexey-pelykh: Ah, got it, sorry for misunderstanding.

Tardo:
This PR deprecates 'users_ldap_populate' no?

alexey-pelykh:
@Tardo well, not quite, `users_ldap_populate` offers one-time import wizard, while this module does not. I think this depends on our cooperation with @hbrunn?

hbrunn:
nothing depends on me fortunately :-)

When `users_ldap_populate` was conceived, it was meant (and is used) for continuous updates, that's why the readme instructs the user how to write a [cronjob](https://github.com/OCA/server-tools/tree/8.0/users_ldap_populate#installation) for that. If I wrote this today, I'd add that as demo data probably.

hbrunn:
PS: As written above, I disagree with all the functionality duplication you introduce here, that's why I abstain from review. But I understand that the unix philosophy of knitting together small stuff vs systemd philosophy of having one behemoth that does it all is not solvable, so I prefer not to press a discussion about it

alexey-pelykh:
@hbrunn you're correct, though one thing may eventually depend on your help, if stars align - helping with migration script if by any chance you decide to eventually support this in the end :)

hbrunn:
I'll be happy to help enthusiastic community members independently of me agreeing with what they do

PR #131, opened 2019-09-02
[ADD] auth_sms

hbrunn:
This module allows you to use SMS as second factor for user authentication.

While SMS is not the most secure way of delivering a secret, it's still safer
than no multi factor authentication at all.


pedrobaeza:
What about https://github.com/OCA/server-tools/tree/10.0/auth_totp ?

hbrunn:
it exists ;-) would be my first choice too indeed, but it's not feasible for the customer. Non-smartphone users and the like

auth_sms/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>
```
NL66278: No more copyright in __init__.py files.

NL66278 (changes requested):
:+1: Looks mostly good to me, but please remove copyright from init files.

PR #132, opened 2019-09-09
[12.0][MIG] users_ldap_populate

Tardo:
Migration of users_ldap_populate from OCA/server-tools 10.0

Cc @tecnativa


users_ldap_populate/models/populate_wizard.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2012 Therp BV (<http://therp.nl>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/gpl.html).
+
+from odoo import models, fields, api
```
Trivedi-Vacha-SerpentCS, line 4: Import statement should follow alphabetic order

users_ldap_populate/models/populate_wizard.py
```diff
@@ -0,0 +1,33 @@
...
+        'res.company.ldap',
+        'LDAP Configuration'
+    )
+    users_created = fields.Integer(
+        'Number of users created',
+        readonly=True
+    )
+    users_deactivated = fields.Integer(
+        'Number of users deactivated',
+        readonly=True
```
Trivedi-Vacha-SerpentCS, line 22: IMHO good to add help on fields.

users_ldap_populate/models/users_ldap.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2012 Therp BV (<http://therp.nl>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/gpl.html).
+
+import re
+
+from odoo import models, fields, api, _, SUPERUSER_ID
```
Trivedi-Vacha-SerpentCS, line 6: Import should follow alphabetic order

rafaelbn:
Please @Yajo revew here thanks

users_ldap_populate/models/users_ldap.py
```diff
@@ -0,0 +1,184 @@
...
+            known_user_ids.extend(item['no_deactivate_user_ids'])
+            deactivate_unknown &= item['deactivate_unknown_users']
+        return deactivate_unknown, known_user_ids
+
+    def _get_ldap_entry_dicts(self, conf, user_name='*', timeout=60):
+        """Execute ldap query as defined in conf.
+
+        Don't call self.query because it supresses possible exceptions
+        """
+        ldap_filter = filter_format(conf['ldap_filter'] % user_name, ())
```
hbrunn, line 122: please cherry-pick https://github.com/OCA/server-tools/pull/1690

users_ldap_populate/tests/test_users_ldap_populate.py
```diff
@@ -63,9 +62,10 @@ def test_users_ldap_populate(self):
             'uid': ['fake'],
             'mail': ['fake@fakery.com'],
         })]):
```
hbrunn, line 64: if you add a second dict here describing to a user that you add before as inactive (demo user will do), you also test the [activating inactive users codepath](https://codecov.io/gh/OCA/server-auth/compare/42b4727d9d0b9fff3dcc6690887168237818b870...e94d80c8cc3ba08efcbc85d8bcae651e1003cb99/diff#D65-71) which probably raises coveragfe enough to let this test pass

hbrunn (changes requested):
please also add

yajo (approved):

users_ldap_populate/tests/test_users_ldap_populate.py
Tardo: I can't use demo user... if the user exists but is inactive the code raises this exception: https://github.com/odoo/odoo/blob/12.0/addons/auth_ldap/models/res_company_ldap.py#L202

Tardo:
@pedrobaeza ready to review

users_ldap_populate/tests/test_users_ldap_populate.py
pedrobaeza: @hbrunn do you agree with the answer?

pedrobaeza:
@Tardo please squash together adjacent translations commits

users_ldap_populate/tests/test_users_ldap_populate.py
hbrunn: no. Turn on creating users in the demo config then

pedrobaeza: @Tardo do you know what to do then? I confess I'm a bit lost with this stuff.

Tardo: @hbrunn Can you verify that '_get_or_create_user' can return a "false" value?
- If the user exists and is actived = The user is used
- If the user exists and is deactivated = Exception
- If the user not exists and creation is not enabled = Exception
- If the user not exists and creation is enabled = Copy/Creates a new user

hbrunn: the function won't return a falsy value, but that's not the point here. Just turn on the `create_user` flag, then you actually test what this addon does.

users_ldap_populate/models/users_ldap.py
hbrunn: this still uses just the string formatting, not `filter_format`

hbrunn (dismissed):

users_ldap_populate/tests/test_users_ldap_populate.py
Tardo: @hbrunn "create_user" is enabled by default: https://github.com/odoo/odoo/blob/12.0/addons/auth_ldap/models/res_company_ldap.py#L34
I don't understand you, sorry.

pedrobaeza (approved):
We can't wait more for this module, and the concerns are only in the tests part, so I'm going to merge it right now as is, and later PRs can improve that if it's a problem.

PR #135, opened 2019-09-27
[12.0][IMP] ADD possibility to save api keys in database by using server_env

lmignon:
supersedes #120

### DIFF
--- a/auth_api_key/migrations/12.0.2.0.0/post-migrate.py
+++ b/auth_api_key/migrations/12.0.2.0.0/post-migrate.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+import odoo
+from odoo.addons.server_environment import serv_config
+
+_logger = logging.Logger(__name__)
+
+
+def migrate(cr, version):
+    _logger.info("Create auth_api.key records from odoo config")
+    with odoo.api.Environment.manage():
+        env = odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
+        for section in serv_config.sections():
+            if section.startswith("api_key_") and serv_config.has_option(
+                section, "key"
+            ):
+                login_name = serv_config.get(section, "user")
+                name = section.replace("api_key_", "")
+                key = "<set from server environment>"
+                user = env["res.users"].search([("login", "=", login_name)])
+                env["auth.api.key"].create(
+                    {"name": name, "key": key, "user_id": user.id}
+                )
+                _logger.info("API Key record created for %s", section)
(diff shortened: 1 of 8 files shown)


auth_api_key/migrations/12.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,27 @@
...
+    _logger.info("Create auth_api.key records from odoo config")
+    with odoo.api.Environment.manage():
+        env = odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
+        for section in serv_config.sections():
+            if section.startswith("api_key_") and serv_config.has_option(
+                section, "key"
+            ):
+                login_name = serv_config.get(section, "user")
+                name = section.replace("api_key_", "")
+                key = "*****"
```
sbidoul: can we leave key empty?

auth_api_key/models/auth_api_key.py
```diff
@@ -1,38 +1,85 @@
...
+        self._retrieve_api_key_id.clear_cache(self.env[self._name])
+        self._retrieve_uid_from_api_key.clear_cache(self.env[self._name])
 
-                if not uid:
-                    raise ValidationError(
-                        _("No user found with login %s") % login_name)
+    @api.model
+    def create(self, vals):
+        record = super(AuthApiKey, self).create(vals)
+        if "key" in vals:
```
sbidoul: `or "user_id" in vals`, because `_retrieve_uid_from_api_key` is cached too?

auth_api_key/migrations/12.0.2.0.0/post-migrate.py
lmignon: The key field is required

auth_api_key/models/auth_api_key.py
lmignon: good catch

auth_api_key/models/auth_api_key.py
```diff
@@ -1,38 +1,85 @@
...
 
-class AuthApiKey(models.TransientModel):
+class AuthApiKey(models.Model):
     _name = "auth.api.key"
-    _description = "API Key Retriever"
+    _inherit = "server.env.mixin"
+    _description = "API Key"
+
+    name = fields.Char(required=True)
+    key = fields.Char(required=True)
```
sbidoul: Right, since this field is required, I suggest some help for users who don't read the documentation. :)

help="The API key. Enter a dummy value in this field if it is obtained from the server environment configuration."

auth_api_key/migrations/12.0.2.0.0/post-migrate.py
sbidoul: nit:

```suggestion
                key = "<set from server environment>"
```

sbidoul (approved):

sbidoul:
@qgroulard can you test the migration? Then we merge.

auth_api_key/migrations/12.0.2.0.0/post-migrate.py
lmignon: done

qgroulard:
@sbidoul @lmignon Tested and approved, great job !

lmignon:
Thank you for the review and tests @qgroulard

PR #143, opened 2019-10-23
[11.0][IMP] Make auth_dynamic_groups extendable.

NL66278:
- Make auth_dynamic_groups extendable bij adding other methods than formula;
- Add option to define group based on partner category;
- Add module to base group on contracts for certain products (membership!);
- Add module to base group op partner relations (board members!).


NL66278:
Close and reopen to try and resolve error.

auth_dynamic_groups_relation/models/res_groups.py
```diff
@@ -0,0 +1,71 @@
...
+        message_parms = {
+            'group': '%d:%s' % (self.id, self.display_name),
+            'user': '%d:%s' % (user.id, user.display_name),
+            'kount': kount,
+        }
+        self._debug_message(
+            _("User %(user)s has %(kount)d relations qualifying"
+              " for dynamic group %(group)s"),
+            message_parms)
+        return kount > 0
```
thomaspaulb, line 71: Since partner relations can also be queried using [regular Odoo domain notation](https://github.com/OCA/partner-contact/blob/11.0/partner_multi_relation/tests/test_partner_relation.py#L248), wouldn't it be easier just to allow to specify a "domain" that defines a dynamic group, in addition to a Python formula. It seems like a simpler and more flexible change.

NL66278: While that would be more flexible, it would also be more complicated for the end-user. This was for a project that has been stopped unfortunately, but still I think just being able to put board members (people that have the relation 'is a board member') into a group giving access to board member type functions - of course this is just an example-, is a rather straight forward functionality.

thomaspaulb: Fair enough, i think it's an and/and thing tho - supporting domain notation would be useful as well for speed of evaluation.

auth_dynamic_groups/README.rst
```diff
@@ -37,12 +58,38 @@ the group or not.
...
-truthy if the user is supposed to be a member of the group and falsy if not.
+Normally group membership is maintained by the administrators. Either manually,
+or by granting authorities on the user record.
+
+You can select other criteria by which the membership of groups is maintained.
+
+You can have membership determined by a formula, and/or by the categories
+of the partner.
+
+Specify that you want to maintain group memebrship automatically by marking
```
thomaspaulb, line 69: Membership

auth_dynamic_groups/models/res_groups.py
```diff
@@ -1,44 +1,181 @@
...
+        relation='dynamic_group_partner_category',
+        column1='gid', column2='category_id',
+        string="Partner categories",
+        help="Used when selection is based on partner category.")
+
+    @api.multi
+    def action_refresh_users(self):
+        """Refresh the users in this group.
+
+        Limit this to 512 users, to prevent huge performance problems.
```
thomaspaulb, line 58: I imagine this could cause regressions on installations with 513 users or more. Maybe this should be a global setting?

auth_dynamic_groups/models/res_groups.py
```diff
@@ -1,44 +1,181 @@
...
+        """
+        res_users = self.env['res.users']
+        for this in self.filtered('is_dynamic'):
+            for user in res_users.search([], limit=512):
+                this.update_group_user(user)
```
thomaspaulb, line 63: To avoid holding long locks, it makes sense to commit after each user.

auth_dynamic_groups/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,24 @@
-This module allows defining groups whose membership is a condition expressed as
-python code. For every user, it is evaluated during login if she belongs to
-the group or not.
+This module allows defining groups whose membership is a set dynamically.
+
+This can be based on a condition expressed as python code and / or on the
+categories attributed to the partner/user.
+
+In addition this module allows for easy extention to add other methods to
```
thomaspaulb, line 6: Extension

auth_dynamic_groups/readme/DESCRIPTION.rst
```diff
@@ -1,3 +1,24 @@
...
+Criteria can be combined. The methods on res.groups define wether a user
+satisfies some test, the value of the corresponding field on res.groups
+determine wether the condition will lead to exclusion of the group, inclusion
+of the group, be ignored, or needs to be combined with other conditions.
+
+If any condition excludes a user from a group, this always 'wins'
+
+Finally a group can be marked as being filled automatically. For instance
+groups that are maintained from an ldap connection can in this way be
+prevented from manual editting.
```
thomaspaulb, line 24: Editing

thomaspaulb (changes requested):
Only the comment about the regression and the commit feels blocking to me, otherwise i think this is a valuable addition and a useful refactor.

PR #144, opened 2019-11-06
add some chinese translations

HengzheLi:


pedrobaeza:
You are changing some files apart from the translations, and translations are handled in https://translation.odoo-community.org/

PR #147, opened 2019-11-25
[11.0][ADD] Auth LDAP User Pattern

bodedra:
This module will support additional features for LDAP authentication.

- It will bypass Standard Odoo user authentication if user pattern match.


bodedra:
Are you available for review?

@b-kannan @smangukiya @gavindav @diverfr @mdodoo @guimarc-br

bodedra:
Ping @dreispt

smangukiya:
@bodedra : It looks good to me. I am commenting here as I am not able to set myself, reviewer.

bodedra:
@smangukiya : Thanks for your quick review.

You can. Please click on "File Changed" > "Review Changes"

smangukiya (approved):

bodedra:
@max3903

auth_ldap_user_pattern/models/res_company_ldap.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def map_ldap_attributes(self, conf, login, ldap_entry):
+        res = super(CompanyLDAP, self).map_ldap_attributes(
+            conf=conf, login=login, ldap_entry=ldap_entry)
+        res.update({'is_ldap_user': True})
+        return res
+
+    @api.multi
+    def get_ldap_dicts(self):
+        res = super(CompanyLDAP, self).get_ldap_dicts()
```
hbrunn, line 24: please explain this override

hbrunn (commented):
can you elaborate what the goal of this module is? So we do avoid Odoo trying to query the user table, please explain the need to do so. Then the module does some unrelated things like warning ldap users that they can't change their password. That shouldn't be in the same module. We have https://github.com/OCA/server-tools/tree/8.0/users_ldap_push in v8 that would allow users to change their passwords for example.

bodedra:
@hbrunn Thanks for your review and comment. Purpose of this module is bypass user authentication if user pattern match. I will back you with more details.

auth_ldap_user_pattern/models/res_company_ldap.py
bodedra: @hbrunn : Purpose to read "user_pattern" field. Now method is optimized in recent commit.

bodedra:
@hbrunn : We have added new security access right to avoid conflict with your module. With this module, User has access right of "Enable LDAP Password Change", will allow to change password otherwise not.

auth_ldap_user_pattern/models/res_company_ldap.py
hbrunn: I see, thanks

auth_ldap_user_pattern/models/res_company_ldap.py
```diff
@@ -0,0 +1,28 @@
...
+        for value in res:
+            user_pattner = self.sudo().browse(value['id']).user_pattern
```
hbrunn: typo

hbrunn:
I still don't get it why we want to avoid the Odoo login for ldap users, can you elaborate on that more?

For implementing this, I've a suggestion that makes this module more general I think. How about saving the user's DN, and also a link to the `res.company.ldap` the user comes from? Then in the password wizard, you very easily can just call the ldap password update.

If you then really want to force the ldap login for ldap users, you can in your login override search for the user, if it has those fields set, you know immediately where to login. This way you don't need the pattern at all, and you can mix usernames whichever way you want.

bodedra:
It is necessary to evaluate and understand different rules based on login patterns and multiple ldap configurations at the client [global implementation with different ldap servers and rules].

Providing more flexibility through configuration avoid local customizations in future as the patterns change / rules change.

Doing it in groups is simpler to maintain for larger customers instead of per user level.

Certain companies would never allow to change password of LDAP users from external application like Odoo. Most enterprise customers, password management is implemented in a centralized manner so as to control password length, content, expiration rules, etc.

auth_ldap_user_pattern/models/res_users.py
```diff
@@ -0,0 +1,55 @@
...
+# Copyright (C) 2019 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import re
+
+from odoo import api, fields, models, registry, SUPERUSER_ID, _
+from odoo.exceptions import UserError
+
+
+class Users(models.Model):
```
nikul-serpentcs, line 11:
```suggestion
class ResUsers(models.Model):
```

auth_ldap_user_pattern/views/res_users_view.xml
```diff
@@ -0,0 +1,35 @@
...
+            </field>
+            <xpath expr="//button[@name='action_reset_password'][1]" position="attributes">
+                <attribute name="attrs">{'invisible': ['|',('state', '!=', 'active'),('is_ldap_user', '=', True)]}</attribute>
+            </xpath>
+            <xpath expr="//button[@name='action_reset_password'][2]" position="attributes">
+                <attribute name="attrs">{'invisible': ['|',('state', '!=', 'new'),('is_ldap_user', '=', True)]}</attribute>
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
nikul-serpentcs, line 35:
```suggestion
</odoo>

```

auth_ldap_user_pattern/views/res_company_ldap_view.xml
```diff
@@ -0,0 +1,14 @@
...
+        <field name="name">res.company.ldap.ext</field>
+        <field name="model">res.company.ldap</field>
+        <field name="inherit_id" ref="auth_ldap.view_ldap_installer_form"/>
+        <field name="arch" type="xml">
+            <field name="user" position="after">
+                <field name="user_pattern"/>
+            </field>
+        </field>
+    </record>
+</odoo>
```
nikul-serpentcs, line 14:
```suggestion
</odoo>

```

auth_ldap_user_pattern/models/res_users.py
```diff
@@ -0,0 +1,55 @@
...
+    def change_password_button(self):
+        for line in self:
+            if not line.new_passwd:
+                raise UserError(_("Before clicking on 'Change Password', "
+                                  "you have to write a new password."))
+            if line.user_id.is_ldap_user and not line.user_id.has_group(
+                    'auth_ldap_user_pattern.group_ldap_password_change'):
+                raise UserError(_("Sorry, You can not change password for "
+                                  "LDAP User!"))
+        res = super(ChangePasswordUser, self).change_password_button()
```
nikul-serpentcs, line 54:
```suggestion
        return  super(ChangePasswordUser, self).change_password_button()
```

nikul-serpentcs:
@bodedra Can you please Added Unit test case, Some file codecov is failed.

dreispt:
@bodedra unfortunately the module description is not clear on what this module tried to achieve.
My guess is that you want to enforce that the user Login must match a certain pattern, to be accepted into Odoo. This is useful for scenarios where some user are valid LDAP users, but we don't want them to be able to be logged in.
If this is the case, why don't you extend `CompanyLDAP.authenticate()` to enforce that validation?
That way you don't need to duplicate the `ResUsers._login()` logic.

A second point is the LDAP change password.
It feels like a useful feature that should require you to install user Patterns check.
I suggest making it a separate module, and turn the option checkbox into a General Settings checkbox.

PR #157, opened 2020-01-16
[13.0] [MIG] auth_ldaps: Migrated the module in v13.

nowheresly:


auth_ldaps/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "website": "https://github.com/OCA/server-auth",
+    "author": "Brainbean Apps (https://brainbeanapps.com), "
+    "Creu Blanca, "
+    "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "installable": True,
+    "application": False,
+    "summary": "Allows to use LDAP over SSL authentication",
+    "depends": ["auth_ldap"],
+    "external_dependencies": {"python": ["ldap"]},
```
etobella: Runbot is failing because of this. I think you should change ldap for pyldap

etobella (approved):
LGTM :+1: 
There is an issue with runbot, I think it is related with my comment

auth_ldaps/__manifest__.py
nowheresly: Indeed! Thanks for your feedback and your time @etobella

nowheresly:
I had no choice but to add `pip install pyldap` in the `.travis.yml` file, otherwise the travis build fail. I'm not sure this is recommended way to do...

PR #161, opened 2020-02-14
[FIX] password_security: translatable error message

yajo:
Without this patch, users must know English to understand why they cannot change their passwords.

@Tecnativa TT21381

### DIFF
--- a/password_security/i18n/es.po
+++ b/password_security/i18n/es.po
@@ -6,16 +6,46 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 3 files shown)


pedrobaeza (commented):
Are you sure in the context where the message is thrown, the language is known? (is it in the context to be evaluated or whatever?)

yajo:
There's no reason why not. As long as this is inside a method, it should have a proper `self.env.lang`.

pedrobaeza:
But as this can be done when setting first time the password not yet logged, I'm not sure about that.

yajo:
Captura de pantalla de 2020-02-17 12-37-42

:point_up: It works, definitely.

pedrobaeza:
OK, great

/ocabot merge minor

PR #162, opened 2020-02-15
[11.0][ADD] group mapping pre validation before creating user

eilst:
Add option to validate group mapping before creating users.
At this time if users exist in the idp (identity provider), they will be created in odoo. Even if there
are no group mappings, i.e. user in odoo whitout any group.
So with this new field, administrators will be able to prevent users with no group mappings to be created in odoo.
Local Authentic server - Odoo

### DIFF
--- a/auth_saml_groups/models/auth_saml.py
+++ b/auth_saml_groups/models/auth_saml.py
@@ -26,6 +26,14 @@ class AuthSamlProvider(models.Model):
             'with SAML groups). If not, manually added groups are preserved.'
         ),
     )
+    create_user_if_mapping = fields.Boolean(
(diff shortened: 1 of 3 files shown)


eilst:
Hey @max3903 @moylop260 , on the last merged PR on branch 11.0 one new module ''auth_saml_groups'', which this PR adds one feature, was added but we cannot see it in the odoo app store. Do you know who could click on the scan button: 
 https://apps.odoo.com/apps/dashboard/repos

moylop260:
Hi @eilst 
Sorry for delay

Is it?
https://apps.odoo.com/apps/modules/11.0/auth_saml_groups/

PR #159, opened 2020-02-03
[IMP] password_security: remove post-install mode in test

yajo:
Usually tests are more resilient like this, and this part of https://github.com/OCA/server-auth/pull/87 was not necessary.

@Tecnativa TT21381

PR #165, opened 2020-02-20
[CI] auth_totp_password_security: test independently

yajo:
Tests in module `auth_totp_password_security` are weird. They seem to be testing implementation rather than behavior. And they are trying to workaround HTTP instead of using an `HttpCase`. As a result, if tests from `password_security` are run along with those of `auth_totp_password_security`, the latter ones fail.

This is not a real fix, but it is the easiest way to respect test resilience in `password_security` and make all go green.

Someone with more interest in `auth_totp_password_security` should fix those nasty tests.

This reverts commit 492201a from https://github.com/OCA/server-auth/pull/159

@Tecnativa TT22338


pedrobaeza:
It seems the exclude doesn't work due to the auto_install. Then you need to exclude both `auth_totp` and `auth_totp_password_security`.

yajo:
Before that, the worse part is that the INCLUDE doesn't work either. It tests password_security and thus breaks tests for auth_totp_password_security.

Is there anything else I can do apart from rewriting the tests? :roll_eyes: I'm out of ideas.

pedrobaeza:
I don't have either...

yajo:
How about removing the bad tests? I know that will decrease coverage, but tests are bad designed anyway...

pedrobaeza:
OK, if it's the only way, and obviously as it's not really testing the feature, only the implementation.

yajo:
OK, wrong tests removed.

pedrobaeza (approved):
Let's fast-track this for solving the CI status

PR #171, opened 2020-03-27
 [12.0][MIG] - keychain migration to 12.0

sbejaoui:


bealdav:
cc @florian-dacosta @hparfr

florian-dacosta:
Hi @sbejaoui 
For which use case do you need this module.
For your information, on our side, and since the server_environment module has elvolved to allow this, we have refactored completly this module in version 12, to depend on server environment and make this module easier to use.

Here are the 2 PRs : 
https://github.com/OCA/server-env/pull/23
https://github.com/OCA/server-env/pull/24

The main advantages is that, with these modules installed, we can encrypt data for each field already managed by sever_environement_* modules.
Example, you have server_environment and data_encryption_server_environment module installed.
If you install mail_environment, it works out of the box.
If you declare the mail field in your config file, it will be taken from there, if not, you can fill it directly in Odoo, and it will be encrypted and environement  dependent.

You can also fill the value for all your environment from the production instance, if all the keychain key (data encryption key in the new module) are configured in the odoo production configuration file.

I am available to discuss it more further if needed, but I think it is important to converge if possible on this one.

@sebastienbeau

sbejaoui:
Hi @florian-dacosta ,

Thank you for the explanation, 

I am porting `base_sms_client` module to V12 (https://github.com/OCA/connector-telephony/pull/209) and find that keychain on its dependencies. 

Now, I double check it and it seems the dependency is useless and the keychain is not really used in `base_sms_client`.

sbejaoui:
I removed the dependency, I don't need this module anymore.

Thank you again @florian-dacosta

PR #152, opened 2019-12-19
12.0 mig auth keycloak

shepilov-vladislav:

PR #174, opened 2020-03-28
12.0 mig auth keycloak

nowheresly:
I just added some _description to avoid warnings.
I had as well to redefine OAuthLogin controller to add a nonce parameter which is now required by keycloak.


rebla:
any ideas when it will be available? Thanks in advance

yvaucher:
have you seen that #152 already exists?

Could you land a review there instead or say why you would superseed it.

PR #175, opened 2020-03-30
[IMP] auth_api_key: decouple server_environment, making it optional

dreispt:
Server Environment is a dependency, but it is not really required.
Server Environment requires specific configuration, and the Odoo service fails to start without it.
IMO there is not good reason to force it to be used.


dreispt:
The build fails while installing, I'm not convinced it is related to my change.

Trying to build locally, I get a different error, coming from `server_environment`:

image

Additional help is appreciated. Thanks!

dreispt:
It was a module bug. Fixed, ready for review!

dreispt:
Needs review please!

lmignon (approved):
Thank you @dreisptè Useful change. LGTM (Code review)

ypapouin (approved):
LGTM - Code Review

PR #186, opened 2020-04-17
[11.0] auth_keycloak: FIX changed param name

acarmisc:
Using the module as is doesn't works against fresh Keycloak setup. 
Validation request response is `400 Bad Request` and further description say the token is missing so I've changed the parameter

### DIFF
--- a/auth_keycloak/models/res_users.py
+++ b/auth_keycloak/models/res_users.py
@@ -21,7 +21,7 @@ class ResUsers(models.Model):
         logger.debug('Calling: %s' % provider.validation_endpoint)
         resp = requests.post(
             provider.validation_endpoint,
-            data={'token': access_token},
+            data={'access_token': access_token},
             auth=(provider.client_id, provider.client_secret)
         )
         if not resp.ok:


gurneyalex:
hello @acarmisc thanks for your contribution.

Do you know which versions of keycloak need this fix?

simahawk:
@acarmisc can you answer @gurneyalex 's question?

acarmisc:
sorry, I've missed the notification 🤷‍♂️ 

I got it with 10.0.0

simahawk (changes requested):
AFAIS this would break it w/ older versions.
I would add a `version` field to the [provider](https://github.com/OCA/server-auth/blob/10.0/auth_keycloak/models/auth_oauth.py) and use the right key based on it.

acarmisc:
@simahawk I haven't check other versions, so it sounds correct... what if we add a field `token_key` instead of a `version` with a default to `token` ?

simahawk:
> @simahawk I haven't check other versions, so it sounds correct... what if we add a field `token_key` instead of a `version` with a default to `token` ?

I thought about it but we might need other hacks based on versions. 
Plus, this should not be something that user can/should change IMO.

Ideally we can use components to lookup for handlers by version in a smart way but at this stage is not really needed.
Let's keep it simple and add a `_auth_token_key()` method and put an IF inside it.

bodedra (commented):
I tried with original following code in Odoo14 EE

    data={'token': access_token},

It is working fine.

gurneyalex:
@acarmisc  any news on this one?

PR #189, opened 2020-04-23
[13.0] making password_security compatible with odoo 13.0

Moro-Code:
Hello, I needed this module for my needs however I found it was not compatible with the latest Odoo version so I made some minor changes to make it compatible.


gurneyalex:
Hello @Moro-Code 

Thanks for this contribution. 

This look a the migration of the module password_security from 12.0 to 13.0. However you did not follow the migration process, and your PR lost the history of commits of that module, therefore we cannot accept it in its current form. 

Can you redo the migration, this time following the process detailed in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 ?

Thanks

gurneyalex (changes requested):
history of commits of the module lost -> needs fixing

gurneyalex:
closed in favor of #239

Issue #33, opened 2018-08-13
[11] auth_session_timeout


Vassy:
Hi all,
I using Odoo 11 CE on Ubuntu 16.04 with Apache 2.4 reverse proxy. 
When i install auth_session_timeout and logout, if load website, odoo redirect to /web/login.
If clear browser cookies, after tray to login, odoo redirect again to /web/login with cycle and login form not load.
If i uninstall auth_session_timeout, no redirect and website load normal.
How to resolve this issue?

pedrobaeza:
Have you tried on runbot? https://runbot.odoo-community.org/runbot/251/11.0 with admin/admin

Vassy:
Same result on runbot:
```
[...]

PR #206, opened 2020-06-17
[12.0][FIX] Remove force expired session

remi-filament:
Forcing the expired to true in case session or modified time could not be found causes that user is always brought to login page in case session could not be find.

This is problematic for a number of cases, including reset_password used for first connection, or for accessing public surveys for instance, and probably also for accessing any public element which should not be redirected to login page...

This PR proposes to keep expired=False in case exception is raised while looking for modified time on session (meaning that either session does not exist or does not have a modified time - although this second case should not be possible afaik)

### DIFF
--- a/auth_session_timeout/models/res_users.py
+++ b/auth_session_timeout/models/res_users.py
@@ -76,8 +76,6 @@ class ResUsers(models.Model):
                 _logger.exception(
                     'Exception reading session file modified time.',
                 )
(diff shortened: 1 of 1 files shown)


remi-filament:
This should fix a number of related issues #33 (although the issue is on version 11.0), once this PR is accepted on 12.0 it could be backported on v11 as well

PR #212, opened 2020-07-29
[FIX] password_security: POT regeneration & Spanish trasnlation (es.po)

gabonog:
- POT file regeneration (last generation made on Odoo 11).
- Spanish translation (es.po only).


pedrobaeza:
POT should be automatically regenerated. If not, check that .travis.yml has latest definition from OCA/maintainer-quality-tools.

PR #214, opened 2020-08-17
[13.0][MIG] auth_saml

theangryangel:
Port of 12.0 auth_saml to 13.0, replacing the use of lasso with pysaml.

~~PR for migrating auth_saml into 13.0.~~

~~There are few changes I've made to this module -~~

1. ~~I've replaced the dependency on lasso with pysaml2. I've found lasso can be a pain to work with and going for something more pure python. I *believe* I've correctly configured it to be secure, but I would appreciate a second pair of eyes.~~
2. ~~Along with the replacement of lasso I've also prevented unsolicited requests from the configured IDP(s)~~
3. ~~I've tried to keep as much the same as possible in the backend, but there are some extra settings, and I've dropped a few others.~~
4. ~~I've tried to add more help information to guide users through the setup on the SAML providers form(s). I don't know if that's a faux-pas?~~

I've tested this internally with SimpleSAMLPHP (https://simplesamlphp.org/) and Duo (https://duo.com/)


theangryangel:
I'll get the failed CI failures fixed up this afternoon/tomorrow morning. In the meantime if there is other feedback I'd appreciate it :)

theangryangel:
~~In terms of adding unit tests I don't think I can stand up a complete IDP in a test (please correct me if I'm wrong and there's some way to orchestrate that within Odoo's testing system)?~~

~~If so that restricts me to testing very simple stuff like ensuring the _check_credentials, and maybe just using pysaml2 to fake some IDP responses and that that various custom methods work correct. I can add them, but I'm unclear if I should/its worth it?~~

Tests have been added using a "fake" IDP.

AntonKhitrovArammeem (approved):

gurneyalex:
@eLBati could you give this PR a look? Do you think the library change deserves backporting to e.g. 12?

eLBati:
Hi @gurneyalex and @theangryangel 

> Do you think the library change deserves backporting to e.g. 12?

It surely does.

`python3-lasso` is a real pain, see https://dev.entrouvert.org/issues/48660

Thanks for pointing this out

theangryangel:
I already have a 12.0 version I can submit if you like.

It has come up this week, internally, that the module probably needs to be company aware. I can submit an update to this PR that adds that and get the 12.0 copy sorted at the same time

eLBati:
> I already have a 12.0 version I can submit if you like

It would be great. Could you publish it somewhere?
Thanks!

theangryangel:
@eLBati sure. I'm booked out solid for client work over the next few days, and have a few days off next week. I'll try and get it public and linked here in the next day or two when I get a sec

eLBati:
I'm sorry for bothering you, but I have the customer waiting.
@theangryangel if you have the v12 version of the module, could you just upload it to a branch? I can take care of testing, reviewing and proposing to OCA.
Thanks!

theangryangel:
@eLBati sorry it's taken me a few days. I'm literally trying to get as much stuff done before my first week off in ages :sweat_smile: 

I've added it to the branch https://github.com/GlodoUK/server-auth/tree/feat/12.0-mig-auth_saml-pysaml

I haven't made a PR yet because it overwrites auth_saml in it's entirety.

We have an internal fork which renamed it to auth_pysaml2 to allow side-by-side installation.

I'm not clear which way, if any, the OCA would want to go with this. I feel doing a bait and switch on the underlying library would be a bad thing for anyone building from source.

Naming aside, both this, and the 13.0 version have an issue on multiple websites which I'm not entirely sure how to address yet. Saml auth requires the IDP to be told where to redirect back to after login.

However on an Odoo installation with multiple websites Odoo (at least in 12.0 and 13.0) redirects the user to that domain when they try and edit pages on it. However, because the domain is completely different there's no cookie set and the user is logged out. Re-doing the saml auth will put you back on the primary domain configured under saml provider (because the saml IDP needs to know where to redirect back to after auth). The only way I can think of tackling is a saml provider all configured per unique domain. Which would suck, but doable if I provide a way to filter the different providers based on current domain on the logon form. For one of our customers using this it would be a huge pain in the ass as it would result in 5+ distinct saml configurations which would be all basically the same.

If anyone has any better ideas on that, I'm all ears.

eLBati:
@theangryangel thanks!

working on https://github.com/OCA/server-auth/pull/234

theangryangel:
Pulled in your changes @eLBati, thanks <3

ygol:
It seem that by switching to pysaml2 it will not be possible to use it on odoo.sh as there is a missing requirement that should be installed with apt-get (apt-get install -y xmlsec1). I tried to use "pip install xmlsec" but it doesn't solve the problem.

theangryangel:
Sorry I don't use Odoo.sh, so forgive a dumb question - but can you use the existing 12.0 version on Odoo.sh, if you can't apt-get install? AFAIK you either have to compile lasso yourself (which I had a painful time integrating into our workflow, which is why I switched libraries) or apt-install python3-lasso? 

The only other maintained option I've found out there in the python ecosystem was onelogin/python3-saml I had problems with incompatible versions of some library (sorry can't remember which).

I don't mind doing the leg work required if I have a place to aim for, but I'll need help finding it.

ygol:
> Sorry I don't use Odoo.sh, so forgive a dumb question - but can you use the existing 12.0 version on Odoo.sh, if you can't apt-get install? AFAIK you either have to compile lasso yourself (which I had a painful time integrating into our workflow, which is why I switched libraries) or apt-install python3-lasso?
> 
> The only other maintained option I've found out there in the python ecosystem was onelogin/python3-saml I had problems with incompatible versions of some library (sorry can't remember which).
> 
> I don't mind doing the leg work required if I have a place to aim for, but I'll need help finding it.

I haven't tried it myself but it seem that lasso can be installed with pip by adding ```python3-lasso``` to requirments.txt in the repo.

cf. [https://medium.com/@divyansht40/saml2-with-odoo13-f36015ac44f4](https://medium.com/@divyansht40/saml2-with-odoo13-f36015ac44f4)

theangryangel:
https://pypi.org/project/python3-lasso/ doesn't get me much, so I assume it's pulling from the Debian or Ubuntu apt repos

To explain how my colleagues and I ended up at the point of replacing lasso with pysaml2

The problem I've got is that we use Doodba, which is based on the Docker Library python images which build from source. I couldn't just apt-install python3-lasso because it wasn't for the appropriate version. I *could* build lasso from source, but it was bloating our build time (think 3-4x). This left me with 3 solutions;
1. Drop Doodba (not politically or practically an option)
2. Create another intermediate layer using Doodba as a base, adding lasso, and then using that image (which would be periodically rebuilt) as our base for projects
3. Replace lasso

When we started this internally the last release of lasso was back in June 2018 and at the time an inaccessible mailing list left us feeling we needed to look elsewhere.

I see since the mailing list is back, and there has been 1 release, which does change things a little. However, I am where I am at this point.

---

I don't like dead PRs just sitting around, because it's another place to get notifications, and it's not fair on anyone.

I guess there are 3 ways to move forward;

1. The OCA is not interested, and I'll close the PR. The company I work for is now invested enough in the pysaml2 version that we need to maintain this forked version, or migrate some installations away from it, so I'd put it up under our company github org at least temporarily.
2. The OCA is interested, but are happy to accept it under a different module name (auth_pysaml2) which I suggested in a [previous comment](https://github.com/OCA/server-auth/pull/214#issuecomment-734470819)
3. The OCA is interested, with some other criteria

ygol:
> I guess there are 3 ways to move forward;
> 
> 1. The OCA is not interested, and I'll close the PR. The company I work for is now invested enough in the pysaml2 version that we need to maintain this forked version, or migrate some installations away from it, so I'd put it up under our company github org at least temporarily.
> 2. The OCA is interested, but are happy to accept it under a different module name (auth_pysaml2) which I suggested in a [previous comment](https://github.com/OCA/server-auth/pull/214#issuecomment-734470819)
> 3. The OCA is interested, with some other criteria

further checking into  using lasso as a requirement in odoo.sh, it also didn't worked. 
So for me there is no preference going with lasso vs pysaml2. 
I would favor your point 2.

eLBati:
I vote for replacing `lasso` with `pysaml2` in `auth_saml`.
`python3-lasso` is a pain and you must pray to not be forced to debug it: https://dev.entrouvert.org/issues/48660

ygol:
Hi.
I am still trying to get it working on odoo.sh. 
The missing dependency (xmlsec1) has been added.

Regarding the Odoo certificate. I tried generating a certificate with:
`openssl genrsa -out for_okta.key 2048
`
and uploaded it as the "Odoo certificate" but this must be wrong as I get the following internal error:
```
2021-02-24 14:21:33,597 110 ERROR ayyeka-test-saml2-2135978 werkzeug: Error on request:
  File "/usr/local/lib/python3.6/dist-packages/werkzeug/serving.py", line 270, in run_wsgi
    execute(self.server.app)
  File "/usr/local/lib/python3.6/dist-packages/werkzeug/serving.py", line 258, in execute
    application_iter = app(environ, start_response)
  File "/home/odoo/src/odoo/odoo/service/server.py", line 439, in app
    return self.app(e, s)
  File "/home/odoo/src/odoo/odoo/service/wsgi_server.py", line 140, in application
    return ProxyFix(application_unproxied)(environ, start_response)
  File "/usr/local/lib/python3.6/dist-packages/werkzeug/contrib/fixers.py", line 152, in __call__
    return self.app(environ, start_response)
  File "/home/odoo/src/odoo/odoo/service/wsgi_server.py", line 117, in application_unproxied
    result = odoo.http.root(environ, start_response)
  File "/home/odoo/src/odoo/odoo/http.py", line 1287, in __call__
    return self.dispatch(environ, start_response)
  File "/home/odoo/src/odoo/odoo/http.py", line 1257, in __call__
    return self.app(environ, start_wrapped)
  File "/usr/local/lib/python3.6/dist-packages/werkzeug/wsgi.py", line 766, in __call__
    return self.app(environ, start_response)
  File "/home/odoo/src/odoo/odoo/http.py", line 1457, in dispatch
    result = ir_http._dispatch()
  File "/home/odoo/src/odoo/addons/web_editor/models/ir_http.py", line 21, in _dispatch
    return super(IrHttp, cls)._dispatch()
  File "/home/odoo/src/odoo/odoo/addons/base/models/ir_http.py", line 238, in _dispatch
    return cls._handle_exception(e)
  File "/home/odoo/src/odoo/odoo/addons/base/models/ir_http.py", line 206, in _handle_exception
    return request._handle_exception(exception)
  File "/home/odoo/src/odoo/odoo/http.py", line 750, in _handle_exception
    return super(HttpRequest, self)._handle_exception(exception)
  File "/home/odoo/src/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/home/odoo/src/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/home/odoo/src/odoo/odoo/addons/base/models/ir_http.py", line 234, in _dispatch
    result = request.dispatch()
  File "/home/odoo/src/odoo/odoo/http.py", line 809, in dispatch
    r = self._call_function(**self.params)
  File "/home/odoo/src/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/odoo/src/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/odoo/src/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/odoo/src/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/home/odoo/src/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/src/user/auth_pysaml/controllers/main.py", line 136, in get_auth_request
    self._get_saml_extra_relaystate(), request.httprequest.url_root.rstrip("/")
  File "/home/odoo/src/user/auth_pysaml/models/auth_saml.py", line 217, in _get_auth_request
    saml_client = self._get_client_for_provider(url_root)
  File "/home/odoo/src/user/auth_pysaml/models/auth_saml.py", line 193, in _get_client_for_provider
    spConfig = self._get_config_for_provider(base_url)
  File "/home/odoo/src/user/auth_pysaml/models/auth_saml.py", line 188, in _get_config_for_provider
    spConfig.load(settings)
  File "/home/odoo/.local/lib/python3.6/site-packages/saml2/config.py", line 337, in load
    self.load_complex(cnf)
  File "/home/odoo/.local/lib/python3.6/site-packages/saml2/config.py", line 271, in load_complex
    self.setattr("", "metadata", self.load_metadata(cnf["metadata"]))
  File "/home/odoo/.local/lib/python3.6/site-packages/saml2/config.py", line 385, in load_metadata
    disable_ssl_certificate_validation=disable_validation)
  File "/home/odoo/.local/lib/python3.6/site-packages/saml2/mdstore.py", line 992, in __init__
    self.security = security_context(config)
  File "/home/odoo/.local/lib/python3.6/site-packages/saml2/sigver.py", line 1027, in security_context
    rsa_key = import_rsa_key_from_file(_file_name)
  File "/home/odoo/.local/lib/python3.6/site-packages/saml2/sigver.py", line 473, in import_rsa_key_from_file
    key = saml2.cryptography.asymmetric.load_pem_private_key(data, None)
  File "/home/odoo/.local/lib/python3.6/site-packages/saml2/cryptography/asymmetric.py", line 12, in load_pem_private_key
    data, password, _backends.default_backend())
  File "/usr/lib/python3/dist-packages/cryptography/hazmat/primitives/serialization.py", line 20, in load_pem_private_key
    return backend.load_pem_private_key(data, password)
  File "/usr/lib/python3/dist-packages/cryptography/hazmat/backends/openssl/backend.py", line 1015, in load_pem_private_key
    password,
  File "/usr/lib/python3/dist-packages/cryptography/hazmat/backends/openssl/backend.py", line 1234, in _load_key
    self._handle_key_loading_error()
  File "/usr/lib/python3/dist-packages/cryptography/hazmat/backends/openssl/backend.py", line 1292, in _handle_key_loading_error
    raise ValueError("Could not deserialize key data.")
ValueError: Could not deserialize key data. - - -
2021-02-24 14:21:33,856 110 INFO ayyeka-test-saml2-2135978 werkzeug: 176.230.58.239 - - [24/Feb/2021 14:21:33] "GET /favicon.ico HTTP/1.0" 404 - 2 0.001 0.006
```
- Do I need to provide a different certificate?
- Is it enough to upload the certificate (private key) or it should also be installed somehow?
thank you.

ygol:
Hi @theangryangel. Odoo has accepted to add the dependency to odoo.sh. so it installs fine now.

I am unable to add the the "SAML User ID" for a user. Getting the following error:
```
Odoo Server Error

  File "/home/odoo/src/odoo/odoo/http.py", line 624, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/odoo/src/odoo/odoo/http.py", line 310, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/home/odoo/src/odoo/odoo/tools/pycompat.py", line 14, in reraise
    raise value
  File "/home/odoo/src/odoo/odoo/http.py", line 669, in dispatch
    result = self._call_function(**self.params)
  File "/home/odoo/src/odoo/odoo/http.py", line 350, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/odoo/src/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/odoo/src/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/odoo/src/odoo/odoo/http.py", line 915, in __call__
    return self.method(*args, **kw)
  File "/home/odoo/src/odoo/odoo/http.py", line 515, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/src/odoo/addons/web/controllers/main.py", line 1327, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/src/odoo/addons/web/controllers/main.py", line 1319, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/home/odoo/src/odoo/odoo/api.py", line 387, in call_kw
    result = _call_kw_multi(method, model, args, kwargs)
  File "/home/odoo/src/odoo/odoo/api.py", line 374, in _call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/home/odoo/src/user/auth_pysaml/models/res_users.py", line 171, in write
    self._autoremove_password_if_saml()
  File "/home/odoo/src/user/auth_pysaml/models/res_users.py", line 160, in _autoremove_password_if_saml
    lambda rec: rec.id != SUPERUSER_ID
  File "/home/odoo/src/odoo/odoo/models.py", line 5344, in filtered
    return self.browse([rec.id for rec in self if func(rec)])
  File "/home/odoo/src/odoo/odoo/models.py", line 5344, in <listcomp>
    return self.browse([rec.id for rec in self if func(rec)])
  File "/home/odoo/src/user/auth_pysaml/models/res_users.py", line 162, in <lambda>
    and not (rec.password or rec.password_crypt)
AttributeError: 'res.users' object has no attribute 'password_crypt'
```
following: https://www.odoo.com/fr_FR/forum/aide-1/retrieve-hashed-password-in-odoo-13-175934
this field has been removed in v12

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+                raise AccessDenied()
+
+    def _autoremove_password_if_saml(self):
+        """Helper to remove password if it is forbidden for SAML users."""
+        if self._allow_saml_and_password():
+            return
+        to_remove_password = self.filtered(
+            lambda rec: rec.id != SUPERUSER_ID
+            and rec.saml_ids
+            and not (rec.password or rec.password_crypt)
```
ygol:
```suggestion
            and not rec.password
```

ygol (approved):

theangryangel:
Thanks, sorry for missing that. I’ll have to investigate why our installations have been working... 😖

ygol:
maybe the PR title should be renamed as there is now https://github.com/OCA/server-auth/pull/251

dreispt:
@theangryangel CI is red, can you have a look please?

auth_pysaml/models/attachment.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Attachment(models.Model):
+    _inherit = "ir.attachment"
+
+    def _save_to_file_system(self):
+        value = self.datas
+        bin_data = base64.b64decode(value) if value else b""
+        checksum = self._compute_checksum(bin_data)
+        self.store_fname = self._file_write(value, checksum)
```
luisg123v: This could cause attachment collisions, because same attachment could be written using the same checksum but different content. I think this file should be removed

auth_pysaml/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import auth_saml
```
luisg123v: Name files according to model names:

```suggestion
from . import auth_saml_provider
```

auth_pysaml/models/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+from . import auth_saml_attribute_mapping
+from . import res_users
+from . import saml_token
+from . import saml_request
```
luisg123v: Same here, model name:

```suggestion
from . import auth_saml_request
```

luisg123v: And here:

```suggestion
from . import auth_saml_request
```

auth_pysaml/models/auth_saml.py
```diff
@@ -0,0 +1,345 @@
...
+        help="Add a CSS class that serves you to style the login button.",
+    )
+
+    body = fields.Char(string="Button Description")
+
+    @api.model
+    def _sig_alg_selection(self):
+        return [(sig[0], sig[0]) for sig in ds.SIG_ALLOWED_ALG]
+
+    sig_alg = fields.Selection(
```
luisg123v: Place all fields before methods

auth_pysaml/models/auth_saml.py
```diff
@@ -0,0 +1,345 @@
...
+        selection=lambda s: s._sig_alg_selection(),
+        required=True,
+        string="Signature Algorithm",
+    )
+
+    @api.onchange("name")
```
luisg123v: Onchange without parameters?

auth_pysaml/models/auth_saml.py
```diff
@@ -0,0 +1,345 @@
...
+    def _onchange(self):
```
luisg123v: `_onchange_fieldname`

auth_pysaml/models/auth_saml.py
```diff
@@ -0,0 +1,345 @@
...
+        for record in self:
```
luisg123v: No need to iterate, onchanges alwais get a single record

auth_pysaml/models/auth_saml.py
```diff
@@ -0,0 +1,345 @@
...
+
+    @api.depends("sp_baseurl")
+    def _compute_sp_metadata_url(self):
+        icp_base_url = (
+            self.env["ir.config_parameter"].sudo().get_param("web.base.url", "")
+        )
+
+        for record in self:
+            if isinstance(record.id, models.NewId):
+                record.sp_metadata_url = None
```
luisg123v:
```suggestion
                record.sp_metadata_url = False
```

auth_pysaml/models/auth_saml.py
```diff
@@ -0,0 +1,345 @@
...
+            record.body = record.name
+
+    @api.depends("sp_baseurl")
+    def _compute_sp_metadata_url(self):
+        icp_base_url = (
+            self.env["ir.config_parameter"].sudo().get_param("web.base.url", "")
+        )
+
+        for record in self:
+            if isinstance(record.id, models.NewId):
```
luisg123v:
```suggestion
            if not record.id:
```

auth_pysaml/models/auth_saml.py
```diff
@@ -0,0 +1,345 @@
...
+
+        vals = {"user_id": matching_value}
+
+        post_vals = self._hook_validate_auth_response(response, matching_value)
+        if post_vals:
+            vals.update(post_vals)
+
+        return vals
+
+    def _get_outstanding_requests_dict(self):
```
luisg123v: I know is a cosmetical change, but there are a lot of blank lines all over the code. For instance, this method has 3 lines and there's a blank line between all of them.

Consider removing all unnecessary blank lines

auth_pysaml/models/auth_saml_attribute_mapping.py
```diff
@@ -0,0 +1,30 @@
...
+
+class AuthSamlAttributeMapping(models.Model):
+    """
+    Attributes to copy from SAML provider on logon, into Odoo
+    """
+
+    _name = "auth.saml.attribute.mapping"
+    _description = "SAML2 attribute mapping"
+
+    provider_id = fields.Many2one("auth.saml.provider", index=True, required=True,)
```
luisg123v: Remove blank lines between fields

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+import passlib
+from passlib import pwd
+
+from odoo import SUPERUSER_ID, _, api, fields, models, tools
+from odoo.exceptions import AccessDenied, ValidationError
+
+_logger = logging.getLogger(__name__)
+
+
+class ResUserSaml(models.Model):
```
luisg123v: - Why this new model?
- If really necessary, move to a different file

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+    _name = "res.users.saml"
```
luisg123v: Missing `_description`

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+                    _(
+                        "This database disallows users to "
+                        "have both passwords and SAML IDs. "
+                        "Errors for login %s"
+                    )
+                    % (self.login)
+                )
+
+    def _auth_saml_validate(self, provider_id, token, base_url=None):
+        p = self.env["auth.saml.provider"].sudo().browse(provider_id)
```
luisg123v: We should avoid one-letter variables, for readability:

```suggestion
                  provider = self.env["auth.saml.provider"].sudo().browse(provider_id)
```

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+        return p._validate_auth_response(token, base_url)
+
+    def _auth_saml_signin(self, provider, validation, saml_response):
+        """ retrieve and sign into openerp the user corresponding to provider
+        and validated access token
+
+            :param provider: saml provider id (int)
```
luisg123v: Indentation:

```suggestion
        :param provider: saml provider id (int)
```

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+    def _check_credentials(self, token):
+        """Override to handle SAML auths.
+
+        The token can be a password if the user has used the normal form...
+        but we are more interested in the case when they are tokens
+        and the interesting code is inside the "except" clause.
+        """
+        try:
+            # Attempt a regular login (via other auth addons) first.
+            super(ResUser, self)._check_credentials(token)
```
luisg123v: - Store result of super
    - Prefer not using old-style super
    
    ```suggestion
                res = super()._check_credentials(token)
    ```

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+        # retrieve and sign in user
+        login = self._auth_saml_signin(provider, validation, saml_response)
+
+        if not login:
+            raise AccessDenied()
+
+        # return user credentials
+        return self.env.cr.dbname, login, saml_response
+
+    def _check_credentials(self, token):
```
luisg123v: Name 2nd parameter as "password", to don't break original signature

```suggestion
    def _check_credentials(self, password):
```

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+        and the interesting code is inside the "except" clause.
+        """
+        try:
+            # Attempt a regular login (via other auth addons) first.
+            super(ResUser, self)._check_credentials(token)
+
+        except (AccessDenied, passlib.exc.PasswordSizeError):
+            # since normal auth did not succeed we now try to find if the user
+            # has an active token attached to his uid
+            res = (
```
luisg123v: Res will be the returned result

```suggestion
            user = (
```

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+                raise AccessDenied()
+
+    def _autoremove_password_if_saml(self):
+        """Helper to remove password if it is forbidden for SAML users."""
+        if self._allow_saml_and_password():
+            return
+        to_remove_password = self.filtered(
+            lambda rec: rec.id != SUPERUSER_ID
+            and rec.saml_ids
+            and not rec.password
```
luisg123v: Password should be removed only if present:

```suggestion
            and rec.password
```

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+        to_remove_password.write(
+            {"password": pwd.genword(length=20, charset="ascii_72")}
```
luisg123v: I think it should be cleared instead, otherwise the condition will be satified on every write

```suggestion
            {"password": False)}
```

auth_pysaml/models/res_users.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+
+    def write(self, vals):
+        result = super().write(vals)
+        if not vals.get("password"):
```
luisg123v: Remove this condition

auth_pysaml/views/base_settings.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" ?>
+<odoo>
+    <!-- Inherit from the configuration form to add a setting. -->
+    <record id="auth_saml_base_settings_form" model="ir.ui.view">
+        <field name="name">auth_saml_base_settings_form</field>
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="base_setup.res_config_settings_view_form" />
```
luisg123v: Inherit the base view instead

```suggestion
              <field name="inherit_id" ref="base.res_config_settings_view_form" />
```

auth_pysaml/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2020 GlodoUK <https://www.glodo.uk/>
+# Copyright (C) 2010-2016 XCG Consulting <http://odoo.consulting>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Saml2 Authentication",
+    "version": "13.0.3.0.0",
```
luisg123v:
```suggestion
    "version": "13.0.1.0.0",
```

auth_pysaml/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "category": "Tools",
+    "author": "XCG Consulting, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/server-auth",
+    "license": "AGPL-3",
+    "depends": ["base_setup", "web"],
```
luisg123v: `base_setup` already depends on `web`:

```suggestion
    "depends": ["base_setup"],
```

auth_pysaml/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "external_dependencies": {"python": ["saml2"]},
+    "demo": [],
+    "data": [
+        "security/ir.model.access.csv",
+        "views/auth_saml.xml",
+        "views/base_settings.xml",
```
luisg123v:
```suggestion
              "views/res_config_settings.xml",
```

auth_pysaml/wizards/__init__.py
```diff
@@ -0,0 +1 @@
+from . import res_config_settings
```
luisg123v: This belongs to models instead of wizards

auth_pysaml/wizards/res_config_settings.py
```diff
@@ -0,0 +1,28 @@
...
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    allow_saml_uid_and_internal_password = fields.Boolean(
+        string="Allow SAML users to posess an Odoo password "
+        "(warning: decreases security)"
+    )
+
+    @api.model
+    def get_values(self):
```
luisg123v: Since this pretty much sets the config parameter, I think it's not necessary, the attribute `config_parameter` could be used in the field, something like:

```
config_parameter="auth_saml.allow_saml.uid_and_internal_password"
default=True
```

luisg123v (changes requested):
- Missing tests, you may check [this](https://github.com/OCA/server-auth/blob/fdf26d55ab40040fd9abdb5186727af635fd23cb/auth_saml/tests/test_saml.py) for a starting point
- Remember to squash migration commits

yvaucher:
@theangryangel to solve the pre-commit issues your can run

`pre-commit run -a` in your local `server-auth` folder.

You can also use `pre-commit install` to make sure each commit is clean on it's own.

For version 14.0 we are interested to migrate it, I think we will name it back to `auth_saml` and keep only one option in Odoo 14.0. We plan to work on this in the upcoming weeks.

theangryangel:
Thanks for the feedback and advice. Myself or one our team will take a look at it one day this week... I’m not sure when yet. We’re in the process of a 14.0 upgrade which requires this module so I can see if we can do both at once?

luisg123v:
I think the name `auth_saml` should be kept

theangryangel:
I've run out of time for unit tests today, I'll have to come back to it next week I'm afraid, sorry guys. It's been an extremely busy week. Sorry.

auth_pysaml/controllers/main.py
```diff
@@ -0,0 +1,234 @@
...
+
+        redirect = request.params.get("redirect") or "web"
+        if not redirect.startswith(("//", "http://", "https://")):
+            redirect = "{}{}".format(
+                request.httprequest.url_root,
+                redirect[1:] if redirect[0] == "/" else redirect,
+            )
+
+        state = {
+            "r": werkzeug.url_quote_plus(redirect),
```
luisg123v: In order to avoid the following warning:

``` 
'werkzeug.url_quote_plus' is deprecated and will be removed in
Werkzeug 1.0. Use 'from werkzeug.urls import url_quote_plus
```

you should import the metod as follows:

`from werkzeug.urls import url_quote_plus`

And then use it:

```suggestion
            "r": url_quote_plus(redirect),
```

auth_pysaml_website/controllers/main.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo.addons.web.controllers.main import Home
+
+
+class SAMLLogin(Home):
+    def _list_saml_providers_domain(self):
+        website_id = getattr(request, "website", None)
+        if website_id:
+            return [
+                "|",
+                ("website_ids.id", "=", website_id.id),
```
luisg123v: The extra `.id` usually is less efficient

```suggestion
                ("website_ids", "=", website_id.id),
```

auth_pysaml_website/controllers/main.py
```diff
@@ -0,0 +1,18 @@
...
+                ("website_ids", "=", False),
+            ]
```
luisg123v: You should also validate that a key is set, otherwise if the attachment is removed or the key is cleared, the provider will be listed but an error will occurr when clicking it:

```suggestion
                            ("sp_pem", "!=", False
                        ]
```

auth_pysaml_website/controllers/main.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo.http import request
+
+from odoo.addons.web.controllers.main import Home
+
+
+class SAMLLogin(Home):
+    def _list_saml_providers_domain(self):
+        website_id = getattr(request, "website", None)
+        if website_id:
```
luisg123v: I think you should define a base domain and, if a website is available, extend it by adding the below terms.

luisg123v (changes requested):
Sharing issues we've found ourselves.

auth_pysaml/models/res_users.py
theangryangel: Adds support for multiple SAML providers per user. This is a real world scenario we encountered.

theangryangel:
I've added some unit tests. They seem to pass locally. I'll take a look at the CI once it completes. I believe I've addressed the outstanding issues. If I've missed any concerns, please let me know, I guess?

Thank you to everyone helping me improve the module, I very much appreciate it :)

> I think the name auth_saml should be kept

There was a previous discussion about this - It was previously auth_saml, but there were concerns about the swapping out of the underlying modules from lasso to pysaml2.

I don't mind renaming it back, but I'd prefer some kind of consensus before renaming it again.

yvaucher:
@theangryangel It seems there is a recursive issue between `auth_user_case_insensitive` and `auth_pysaml`.

The issue is most likely that we have 2 modules in which their `write()` calls an other `write()`.

Suggestion to solve:

Check first if your recordset is not empty before calling `write()`. This should avoid looping on `write()` calls.
An other option would be to inject a context key to avoid going twice in there.

```
if to_remove_password:
    to_remove_password.write({"password": False})
```


```
...
File "/home/travis/build/OCA/server-auth/auth_user_case_insensitive/models/res_users.py", line 31, in write
return super(ResUsers, self).write(vals)
File "/home/travis/build/OCA/server-auth/auth_pysaml/models/res_users.py", line 147, in write
self._autoremove_password_if_saml()
File "/home/travis/build/OCA/server-auth/auth_pysaml/models/res_users.py", line 143, in _autoremove_password_if_saml
to_remove_password.write({"password": False})
File "/home/travis/build/OCA/server-auth/auth_user_case_insensitive/models/res_users.py", line 31, in write
return super(ResUsers, self).write(vals)
File "/home/travis/build/OCA/server-auth/auth_pysaml/models/res_users.py", line 147, in write
self._autoremove_password_if_saml()
File "/home/travis/build/OCA/server-auth/auth_pysaml/models/res_users.py", line 143, in _autoremove_password_if_saml
to_remove_password.write({"password": False})
...
RecursionError: maximum recursion depth exceeded while calling a Python object
```

yvaucher:
@theangryangel for the name of the module, my opinion is that if we give a major version change to `auth_saml` we should be able to merge your work as a replacement of the previous version of `auth_saml`.

IMO we want to limit the number of modules that does the same. Appart from using 2 different librairies, I don't see the point of having 2 different modules.

Creating a new module that duplicates the base of the old module also has some draw backs in term of git history.

And BTW https://github.com/OCA/server-auth/pull/251 was closed saying this PR superseeds it.

I'm not sure what were the reasons to call it `auth_pysaml` IIUC it comes from the assuption that `lasso` can be installed on `odoo.sh`. But it was then disputed here: 
https://github.com/OCA/server-auth/pull/214#issuecomment-752018191

While I think it's usually more an exception to have multiple module with the same scope in the OCA.
Let's check if we can get close to a consensus.

### Naming of the module

(subject to updates)

#### In favor of keeping the name `auth_saml`:

1. @eLBati already integrated pysaml2 in `auth_saml` in version 12.0 and voted to replace lasso by pysaml here https://github.com/OCA/server-auth/pull/214#issuecomment-752118968
2. @luisg123v closed his PR https://github.com/OCA/server-auth/pull/251 in favor of this one to have only one module. And mentionned it here https://github.com/OCA/server-auth/pull/214#issuecomment-792847799
3. @yvaucher I'm in favor of keeping `auth_saml`. (representing Camptocamp opinion)

#### In favor of having a new module named `auth_pysaml`:

1. ?

#### unknown

1. @theangryangel 
2. @ygol raised concerns on compatibility with `odoo.sh` then retracted.
3. @max3903
4. other authors and contributors (contacted by email and did not answer after 7 days)

vincent-hatakeyama:
The name was not auth_lasso so I do not see the reason for a rename. To be able to use different library that provides SAML integration, a generic module would be made, with different implementations (so auth_saml_lasso, auth_saml_pysaml, etc). I am unsure that different implementation are needed ; or maybe when it is needed, the auth_saml module would be split.

We did not have trouble with building lasso ourselves, we also sent a patch to make it compatible with python 3.

Still, we thought of maybe switching library. If we had, we were thinking of using https://github.com/onelogin/python3-saml rather that pysaml2 used here.

theangryangel:
> The name was not auth_lasso so I do not see the reason for a rename. 

At this point I'm yet to see anyone complain about keeping it as auth_saml, so I'll rename it back and will update the PR one day this week.

I'll see if I can sort the code coverage by adding some additional tests, at the same time. It's a short week here in the UK, but I'll do what I can.

> Still, we thought of maybe switching library. If we had, we were thinking of using https://github.com/onelogin/python3-saml rather that pysaml2 used here.

My colleague started with onelogin/python3-saml, but there were issues stemming from that library's use of xmlsec causing some incompatibility issue with Odoo's XML rendering under 12.0 (which is where this code originally started). I don't know the full details. I'll be honest I would have much rather used onelogin/python3-saml rather than pysaml. I feel like I've spent far too much time in the depths of pysaml2 :sweat_smile:

yvaucher:
@theangryangel any news on this?

theangryangel:
I've pushed some updates, hopefully that makes everyone happy? If yes then I can squash the commit if you prefer? I should have time for this over the next few days :)

eLBati:
Hi @theangryangel `DESCRIPTION.rst` is missing

PCatinean:
@theangryangel thanks for the great work of migrating this. I just tested this and seems to work really well! There are however two things I wanted to mention:

1) If I understood correctly when adding the certificate you need to add the private key (without password) and certificate bundled together. The controller that prints the metadata that needs to be added to added to the provider exposes both of these values and it is public https://github.com/OCA/server-auth/pull/214/files#diff-d051ee3f0ce073219b1c7ebc99bfa35206c9ebeefbf62df714fb8c800ad0310cR213. Someone would just need the database name and try a few random db ids for providers and can have access to it. I think it should not be public in this case.

2) The boolean setting to disallow password usage does not work, whenever one sets it false and saves it resets back to True.

theangryangel:
>  I think it should not be public in this case.

You're right, this is a mistake. I don't know why I hadn't noticed :s I'll get that sorted today.

> The boolean setting to disallow password usage does not work, whenever one sets it false and saves it resets back to True.

I'll take a look at this at the same time.

theangryangel:
I've patched both the 13.0 and 14.0 branches to separate out the keys again.

> The boolean setting to disallow password usage does not work, whenever one sets it false and saves it resets back to True.

I'll be coming back to this and the pre-commit issues next week.

PCatinean:
Also a really nice improvement would be to have booleans on the provider for awaiting signature on response and assertion to give a bit more flexibility https://github.com/OCA/server-auth/pull/214/files#diff-19fdbbb274613d822573aba74ad5575708e31bddf1688f6d909e589370f6ec3fR177 but that is a improvement rather than a fix and I can make a PR later for it if needed.

PCatinean (approved):

theangryangel:
> awaiting signature on response and assertion

@PCatinean just to double check I understand, you mean adding options to enable/disable signing and assertions? I did have that in the original version, it won't take much to put it back in :) I can get that sorted at the same time as the pre-commit and disallow password usage

PCatinean:
Yes it was about that, some providers cannot be bothered to change the settings :) If they are hardcoded one needs to do inherit/overwrite on the code

PCatinean:
It seems only coverage is the blocker otherwise it works well and tested and used from my side :+1:

auth_saml/models/res_users.py
```diff
@@ -0,0 +1,203 @@
...
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("auth_saml.allow_saml_uid_and_internal_password", "True")
+        )
+
+    def _ensure_saml_token_exists(self):
+        # workaround for Odoo not seeing the newly created auth_saml.token on
+        # the very first login
+        model_token = self.env["auth_saml.token"].sudo()
+        for record in self.filtered(lambda r: r.saml_ids):
```
PCatinean: This also causes a problem when trying to change the language for example. It triggers the write method which triggers this one and you have an access error from a regular user.

theangryangel: I'm having issues reproducing/understanding this one Paul.. Is it just a case of logging in as a user and changing the language? Or?

PCatinean: Yes, just log in as a regular user with no administrative privileges and try to change the language.

PCatinean: It comes from the fact that the user itself is writing on the res.user object (lang_id change) which triggers the write that triggers this method and tries to read own_user.saml_ids. This winds up in an access error because he does not have permissions to read the saml_ids even though they are attached to his own user

pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0.

PCatinean (approved):
LGTM

dreispt:
There is a conflict preventing a merge, thanks for fixing it.

randall-vx:
@moylop260 @luisg123v

theangryangel:
Rebased and pushed.

randall-vx:
@ovnicraft merged into 13.0, FYI

PR #215, opened 2020-08-27
[FIX] auth_saml: commit transaction after check saml

avkhozov:
`login_and_redirect` uses new transaction, but transaction in origin controller isn't commit yet and tokens in `auth_saml.token` aren't visible in `_check_credentials` method.

Seems that code was lost while migrate from 11.0 branch: https://github.com/OCA/server-auth/blob/11.0/auth_saml/controllers/main.py#L185


gurneyalex:
hello @avkhozov  thanks for your contribution. I merged #228 which I'd reviewed before seeing your identical patch which you had submitted earlier. I'm closing this PR.

PR #229, opened 2020-11-18
[12.0] FIX auth_saml when reading/writing config_parameter: False values are not stored

eLBati:
Also, the default value should be False anyway, for security


gurneyalex (approved):

eLBati:
Included in https://github.com/OCA/server-auth/pull/234

PR #232, opened 2020-11-24
[12.0][MIG] auth_from_http_basic

diggy128:


StefanRijnhart:
Please use the format-patch method described in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

PR #249, opened 2021-02-15
[13.0][MIG] users_ldap_populate: Migration to 13.0

joao-p-marques:
Standard migration + improve tests.

@Tecnativa
TT28154

ping @Tardo @Yajo


users_ldap_populate/models/users_ldap.py
```diff
@@ -0,0 +1,194 @@
...
+                    user_id = self.with_context(
+                        no_reset_password=True
+                    )._get_or_create_user(conf, login, result)
+                except AccessDenied:
+                    # this happens if the user exists but is active = False
+                    # -> fetch the user again and reactivate it
+                    self.env.cr.execute(
+                        "SELECT id FROM res_users WHERE lower(login)=%s", (login,)
+                    )
+                    res = self.env.cr.fetchone()
```
joao-p-marques, line 79: IDK if this could be changed to improve security and prevent a potential SQLi risk. But then, it's also how Odoo does it: https://github.com/odoo/odoo/blob/13.0/addons/auth_ldap/models/res_company_ldap.py#L193
@Tardo @Yajo WDYT?

pedrobaeza: I think it can be OK.

pedrobaeza (approved):

victoralmau (approved):

PR #250, opened 2021-02-17
[13.0][MIG+IMP] users_ldap_groups: Migration to 13.0

joao-p-marques:
Standard migration + Improvement to warn user when configuring Odoo to override the default groups with default ones. One of the groups should be a "User type", otherwise the user won't be correctly created.

Also affected by https://github.com/odoo/odoo/issues/41060, but it is a non-blocking issue. 

@Tecnativa 
TT28153

@Yajo @pedrobaeza can you review please?


pedrobaeza:
@Yajo can you please check?

users_ldap_groups/models/res_company_ldap.py
```diff
@@ -48,5 +55,16 @@ def _get_or_create_user(self, conf, login, ldap_entry):
...
                 )
                 groups.append((4, mapping.group_id.id, False))
+        if (
+            this.only_ldap_groups
+            and len([g[1] for g in groups if g[0] == 4 and g[1] in essential_groups])
+            != 1
+        ):
+            raise UserError(
+                _(
+                    "The created user needs to have one (and only one) of the"
```
pedrobaeza, line 65: Why do you need this check? Odoo already assures this.

pedrobaeza (approved):
Except my comment

users_ldap_groups/models/res_company_ldap.py
joao-p-marques: Not exactly. This issue happens when we select the `only_ldap_groups` option, thus replacing all of the default groups added by Odoo by the ones we define. 
If we don't set one (and only one) of the user types, we will get an inconsistent behavior without a proper warning by Odoo, so IMHO it is better to give this error to the user and stop the creation of the new user.

pedrobaeza (approved):

users_ldap_groups/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2012-2018 Therp BV <https://therp.nl>
+# Copyright 2018 Brainbean Apps <https://brainbeanapps.com>
+# Copyright 2021 Tecnativa - João Marques
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "LDAP groups assignment",
+    "version": "13.0.1.0.0",
+    "depends": ["auth_ldap"],
+    "author": "Therp BV, " "Odoo Community Association (OCA)",
```
primes2h, line 10: Repo website is missing.

users_ldap_groups/models/res_company_ldap_operator.py
```diff
@@ -0,0 +1,41 @@
...
+from odoo import api, models
+
+_logger = getLogger(__name__)
+
+
+class ResCompanyLdapOperator(models.AbstractModel):
+    """Define operators for group mappings"""
+
+    _name = "res.company.ldap.operator"
+    _description = "Definition op LDAP operations"
```
primes2h:
```suggestion
    _description = "Definition of LDAP operations"
```

primes2h (changes requested):

joao-p-marques:
Thanks @primes2h changes made!

primes2h (approved):
:+1:

PR #254, opened 2021-03-12
[14.0][ADD] Module vault and vault_share

fkantelberg:
As suggested on the mailing list I want to push the password manager to this repository. password_security might also be useful in the future.

- Secrets are stored with an e2ee. The encryption and decryption happens
in the browser of the user and not even the server can see the secrets.

- The private keys are stored encrypted in the Odoo database to allow
cross device access.

- Import from various formats. Currently a custom format and mostly
keepass databases are supported. The export to the custom format is also
implemented.

- Internal/external users can share secrets with internal users
(vault.inbox)

- Internal users can share a secret with an external user using an url
and a pin. (vault.share) This is done by the extra module vault_share.
Because the encryption of the secrets is weakened to gain some comfort I
had to exclude it from the main module.


vault/models/abstract_vault.py
```diff
@@ -0,0 +1,105 @@
...
+    @api.model
+    def create(self, values):
+        res = super().create(values)
+        if not self.env.context.get("skip_log", False):
+            _logger.info("Vault log: %s", res.message)
+        return res
+
+
+class AbstractVault(models.AbstractModel):
+    """Models must have the following fields:
```
rousseldenis, line 13: Please put each model in a separate file.

vault/models/abstract_vault.py
```diff
@@ -0,0 +1,105 @@
...
+        readonly=True,
+    )
+    entry_id = fields.Many2one(
+        "vault.entry",
+        "Entry",
+        ondelete="cascade",
+        readonly=True,
+    )
+    user_id = fields.Many2one("res.users", "User", required=True, readonly=True)
+    state = fields.Selection(_get_log_state, readonly=True)
```
rousseldenis: You can use a lambda

vault/models/res_users.py
```diff
@@ -0,0 +1,127 @@
...
+            keys = rec.keys.filtered("current")
+            rec.active_key = keys[0] if keys else None
+
+    @api.depends("inbox_token")
+    def _compute_inbox_link(self):
+        base_url = self.env["ir.config_parameter"].sudo().get_param("web.base.url")
+        for rec in self:
+            rec.inbox_link = f"{base_url}/vault/inbox/{rec.inbox_token}"
+
+    active_key = fields.Many2one(
```
rousseldenis: Please put fields definition on top.

vault/models/res_users.py
```diff
@@ -0,0 +1,127 @@
...
+            "iterations": self.active_key.iterations,
+            "iv": self.active_key.iv,
+            "private": self.active_key.private,
+            "public": self.active_key.public,
+            "salt": self.active_key.salt,
+            "uuid": self.active_key.uuid,
+        }
+
+
+class ResUsersKey(models.Model):
```
rousseldenis: Same as above, one model per file.

vault/models/res_users.py
```diff
@@ -0,0 +1,127 @@
...
+    @api.depends("public")
+    def _compute_fingerprint(self):
+        for rec in self:
+            if rec.public:
+                hashed = sha256(rec.public.encode()).hexdigest()
+                rec.fingerprint = ":".join(re.findall(r".{2}", hashed))
+            else:
+                rec.fingerprint = False
+
+    user_id = fields.Many2one("res.users", required=True)
```
rousseldenis: Same as above, fields definition on top.

vault/models/res_users.py
```diff
@@ -0,0 +1,127 @@
...
+        domain = [
+            ("user_id", "=", self.env.uid),
+            ("private", "=", private),
+        ]
+        if not self.search(domain):
+            # Disable all current keys
+            self.env.user.keys.write({"current": False})
+
+            rec = self.create(
+                {
```
rousseldenis: A useful tip is to have a _prepare_values() function to allow overriding.

Same for domain building above.

vault/models/vault.py
```diff
@@ -0,0 +1,158 @@
...
+        "Owner",
+        readonly=True,
+        default=lambda self: self.env.user,
+        required=True,
+    )
+    right_ids = fields.One2many(
+        "vault.right",
+        "vault_id",
+        "Rights",
+        default=_get_default_rights,
```
rousseldenis: You can use a lambda here.

vault/models/vault.py
```diff
@@ -0,0 +1,158 @@
...
+                {
+                    "user_id": self.env.uid,
+                    "perm_write": True,
+                    "perm_delete": True,
+                    "perm_share": True,
+                },
+            )
+        ]
+
+    user_id = fields.Many2one(
```
rousseldenis: Same as above, put fields definitions on top.

vault/models/vault_entry.py
```diff
@@ -0,0 +1,144 @@
...
+    _order = "name"
+
+    name = fields.Char(required=True)
+
+    _sql_constraints = [
+        ("name_uniq", "unique(name)", "The tag must be unique!"),
+    ]
+
+
+class VaultEntry(models.Model):
```
rousseldenis, line 13: Same as above

vault/models/vault_entry.py
```diff
@@ -0,0 +1,144 @@
...
+            else:
+                rec.complete_name = rec.name
+
+    @api.depends("expire_date")
+    def _compute_expired(self):
+        now = datetime.now()
+        for rec in self:
+            rec.expired = rec.expire_date and now > rec.expire_date
+
+    parent_id = fields.Many2one("vault.entry", "Parent", ondelete="cascade")
```
rousseldenis, line 20: Same as above

vault/models/vault_inbox.py
```diff
@@ -0,0 +1,94 @@
...
+    _sql_constraints = [
+        (
+            "value_check",
+            "CHECK(secret IS NOT NULL OR secret_file IS NOT NULL)",
+            _("No value found"),
+        ),
+    ]
+
+    def read(self, *args, **kwargs):
+        return super(VaultInbox, self.with_context(bin_size=False)).read(
```
rousseldenis, line 58: Maybe a little comment here to say what bin_size does

rousseldenis (changes requested):
LGTM. Thanks for this.

Essentially naming convention issues. I will test it for sure.

rousseldenis:
@fkantelberg Could you check travis ?

vault/views/menuitems.xml
```diff
@@ -0,0 +1,52 @@
...
+
+    <record id="action_vault_right" model="ir.actions.act_window">
+        <field name="name">Rights</field>
+        <field name="res_model">vault.right</field>
+        <field name="view_mode">tree</field>
+        <field name="view_id" ref="view_vault_right_overview_tree" />
+        <field name="search_view_id" ref="view_vault_right_overview_search" />
+    </record>
+
+    <menuitem id="menu_vault" groups="base.group_user" action="action_vault" />
```
rousseldenis: Could you add an icon here. If you use web_enterprise or web_responsive, it is a little bit ugly.

image

vault/models/vault_inbox.py
fkantelberg: I hope this helps. It should just prevent Odoo from passing the size of the binary data to the frontend instead of the binary data breaking the JS widgets.

fkantelberg (commented):
@rousseldenis Thanks for the review. I did the suggested changes.

rousseldenis:
I got this on runbot :

image

I tried it locally on v13 and it's ok

fkantelberg:
@rousseldenis I found 2 things there. First I missed the kdbxweb library because of the gitignore. 2nd you need an https context for the browsers crypto API. But this means it's blocking normal http access. I have to check how I can fully disable the vault for http or what to do in this case. Locally is handled differently in the browser I guess.

vault/wizards/vault_send_wizard.py
```diff
@@ -0,0 +1,48 @@
...
+    ]
+
+    def action_send(self):
+        self.ensure_one()
+        self.env["vault.inbox"].sudo().store_in_inbox(
+            self.name,
+            self.secret,
+            self.secret_file,
+            self.iv,
+            self.key_user,
```
rousseldenis, line 48: I think key_user is not required here, key instead. Old code ?

fkantelberg: `key` here is the aes key wrapped with the own public key. `key_user` is the key wrapped with the other users public key. Because you send it to the other user he should be able to decrypt it. See the js vault_controller.js for this wizard.

rousseldenis: Yes, I did.

image


In the js, key_user is not used I think. Moreover, the create inbox function does not get that argument:

https://github.com/OCA/server-auth/pull/254/files#diff-cb469858e7c43a9acaa02aab5a9ed768ad679a61f1d9a891e3e50acffd5d4699R66

fkantelberg: Does the other user has a public key? Have you logged in once with the other user?

rousseldenis: Ok, so the user should have filled in his key ?

So, the message should be a little bit more explicit no ?

fkantelberg: Yes I try to add a proper message. This works with the rsa key. You can only send to other users if they also have a key pair.

fkantelberg: Ok I added message and fixed the domain there because the `active_key` field wasn't searchable. Only usable users should show up now.

fkantelberg:
@rousseldenis Have you completed your tests and found other issues? I may find time for it in the next 2 weeks and only have the http problem on the list.

rousseldenis (approved):
Code review and functional tests

rousseldenis:
@OCA/core-maintainers Could you review?

PR #260, opened 2021-04-09
[13.0] migrate auth_oidc

sbidoul:
Migration from https://github.com/OCA/server-auth/pull/230

PR #261, opened 2021-04-09
[13.0] auth_oidc improvements

sbidoul:
Based on #260 

I plan the following improvements

- [x] use `id_token token` response_type to avoid confusion between access_token and id_token
- [x] implement the authorization code flow (in addition to the currently implemented implicit flow)
  - [x] with PKCE
  - [x] with client_id/client_secret and basic auth
- [x] cache get_key
- [x] do not reuse the validation_endpoint field to define the jwks keys URL, create a new field instead


sbidoul:
@aschenkels-ictstudio @gurneyalex This may be of interest to you.

sbidoul:
Also added

- a demo keycloak configuration and corresponding demo data in Odoo
- a better secret for nonce
- links to the specs in the code

simahawk:
would this replace https://github.com/OCA/server-auth/tree/10.0/auth_keycloak ?

sbidoul:
@simahawk IIUC the biggest part of `auth_keycloak` is about user management (sync users between Odoo and keycloak).

`auth_oidc` is purely about authentication with the OpenID Connect protocol (and incidentally resolving some weaknesses in the underlying Odoo oauth implementation, thanks to the authorization code flow).

So the [part](https://github.com/OCA/server-auth/blob/a3e4750f3d3a0442568816877d3affaebc08ce0c/auth_keycloak/models/res_users.py#L20-L48) of `auth_keycloak` that is about authentication is best replaced with this one. The rest, which is keycloak-specific, still makes sense in `auth_keycloak`.

sbidoul:
I'd like to add server_environment integration in auth_oidc (and auth_jwt). Is it ok to put in the base module, or should I do a separate server_environment integration module ?

janikvonrotz:
@sbidoul Is it possible to get rid of `python-jose`. A non default dependency makes installation in Docker container much harder.

sbidoul:
@janikvonrotz I don't see how to get rid of it. I'm not going to roll my own jwt implementation.
How is it a problem to `pip install python-jose` ?

janikvonrotz:
@sbidoul You are right! I didn't know this was JWT-related.

janikvonrotz:
@sbidoul I've tested with GitLab:

Provider Name: `Login with GitLab`\
Auth Flow: `OpenID Connect (authorization code flow)`\
Token Map: `sub:user_id`\
Client ID: `1671ed7714ec2c2fb58c43f71f7cced55c5c85355f7d88f85fc6614b9c7c1b1d`\
Client Secret: `**************************************************`\
Allowed: `[x]`\
Body: `Login with GitLab`\
Authorization URL: `https://gitlab.com/oauth/authorize`\
Scope: `openid`\
Token URL: `https://gitlab.com/oauth/token`\
JWKS URL: `https://gitlab.com/-/jwks`

I have a question regarding the token map.

Is it possible that the name, login (email) and user_id are mapped from the token?
I assume I need to update the application scope to get the name and email in the token. Currently I get:

```
{
'iss': 'https://gitlab.com',
'sub': '228843',
'aud': '1671ed7714ec2c2fb58c43f71f7cced55c5c85355f7d88f85fc6614b9c7c1b1d',
'exp': 1621948923,
'iat': 1621948803,
'nonce': '0T0h_fHcLGOjJ2UeqS6Rssw_5XAiJ41M3rTvUh7ulXE',
'auth_time': 1621943878,
'sub_legacy': '807cb7014cda3605203322b03340a2ef454e982a02aff31433633bbf271530b8'
}
```

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -0,0 +1,82 @@
...
+# License: AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import logging
+import secrets
+
+import requests
+
+from odoo import fields, models, tools
+
+try:
```
janikvonrotz, line 12: Lets remove the check in favor of this manifest section:

```
'external_dependencies': {
   'python': ['python-jose'],
}
```
What do you think?

sbidoul:
Yes, to get more information in the token you need to configure the scopes on the identity provider side.

janikvonrotz:
@sbidoul No luck, I tried with profile and email. Always the same properties in the token. Also GitLab does not allow the scope to contain multiple items as showed in the keycloak example.

auth_oidc/models/auth_oauth_provider.py
sbidoul: It's an old OCA rule to protect third party imports. It's probably not necessary anymore in recent Odoo versions.

auth_oidc/controllers/main.py
```diff
@@ -0,0 +1,50 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class OpenIDLogin(OAuthLogin):
+    def list_providers(self):
+        providers = super(OpenIDLogin, self).list_providers()
+        for provider in providers:
+            flow = provider.get("flow")
+            if flow in ("id_token", "id_token_code"):
```
simahawk, line 22: any chance to split this in smaller methods?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -0,0 +1,82 @@
...
+        response = r.json()
+        for key in response["keys"]:
+            if key["kid"] == kid:
+                return key
+        return {}
+
+    def _map_token_values(self, res):
+        if self.token_map:
+            for pair in self.token_map.split(" "):
+                from_key, to_key = pair.split(":")
```
simahawk: I'd call `strip` on them to make sure they are clean

auth_oidc/models/res_users.py
```diff
@@ -0,0 +1,69 @@
...
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    @api.model
+    def auth_oauth(self, provider, params):
+        oauth_provider = self.env["auth.oauth.provider"].browse(provider)
+        if oauth_provider.flow not in ("id_token", "id_token_code"):
+            return super(ResUsers, self).auth_oauth(provider, params)
+
+        if oauth_provider.flow == "id_token":
```
simahawk, line 49: can you split this? Something like:

```py
handler = getattr(self, oauth_provider.flow)
access_token, id_token = handler(params)
```

simahawk (approved):
non blocking remarks

simahawk:
@sbidoul shall we close #260?

janikvonrotz:
@sbidoul I justed tested the current state with Odoo 14.0 and my GitLab configuration. No problems occured.

auth_oidc/controllers/main.py
sbidoul: Roughly, each line sets one param, so... no :)

auth_oidc/models/res_users.py
sbidoul: Good point. Done.

sbidoul:
Thanks for the review @simahawk !

PR #262, opened 2021-04-14
[13.0] auth_oauth_multi_token improvements

sbidoul:
Individual commits should make sense.

The main change is that old tokens are deleted instead of disabled.

@simahawk @fmdl @Yajo as you have contributed to this module, do you remember a reason why tokens are disabled and never deleted (they accumulate forever) ?

### DIFF
--- a/auth_oauth_multi_token/models/auth_oauth_multi_token.py
+++ b/auth_oauth_multi_token/models/auth_oauth_multi_token.py
@@ -1,5 +1,6 @@
 # Copyright 2016 Florent de Labarre
 # Copyright 2017 Camptocamp
+# Copyright 2021 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
 from odoo import api, fields, models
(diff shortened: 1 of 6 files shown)


sbidoul:
I am testing this with `auth_oidc` (#261). It works well, as far as the underlying Odoo oauth implementation is acceptable (Odoo never detects expired tokens for instance).

auth_oauth_multi_token/models/auth_oauth_multi_token.py
```diff
@@ -30,24 +33,17 @@ def create(self, vals):
...
 
     def _oauth_validate_multi_token(self):
         """Check current user's token and clear them if max number reached."""
         user_tokens = self._oauth_user_tokens(self.user_id.id)
         max_token = self.user_id.oauth_access_max_token
         if user_tokens and len(user_tokens) > max_token:
             # clear last token
-            user_tokens[max_token - 1]._oauth_clear_token()
-
-    def _oauth_clear_token(self):
```
simahawk, line 49: I'd keep `_oauth_clear_token` and delete them there. Offers a hook and keeps backward compat.

fmdl:
@sbidoul just for log

yajo (approved):
I don't remember why it was like it was, but this makes sense.

simahawk (approved):

auth_oauth_multi_token/models/auth_oauth_multi_token.py
sbidoul: good point, done

PR #265, opened 2021-04-20
[13.0][ADD] auth_oauth_autologin

sbidoul:
This modules implements an automatic redirection to the configured OAuth
provider login page, if there is one and only one enabled. This effectively
makes the regular Odoo login screen invisible in normal circumstances.

See README for more details.


sbidoul:
@ThomasBinsfeld

ThomasBinsfeld (approved):

mlaitinen (approved):
Excellent module :+1: Tested with Azure AD and works beautifully.

sbidoul:
@mlaitinen thanks for testing. We run it in production and it works very well indeed.

/ocabot merge patch

sbidoul:
Hm, standalone tests of this module  pass on my machine. There might be another module that conflicts with it ?

nilshamerlinck:
Hi @sbidoul it seems due to `auth_saml` that overrides [`_list_providers()`](https://github.com/OCA/server-auth/blob/13.0/auth_saml/controllers/main.py#L59) in an exclusive way. I believe it should be tested apart.

sbidoul:
Good catch @nilshamerlinck, thanks! I have quarantined the rebel module by running `copier update` and answering `["auth_saml"]` to the Rebel modules groups question.

/ocabot merge patch

PR #266, opened 2021-04-20
[12.0][IMP]auth_admin_passkey: Block loggin using admin passkey with some users.

manuelregidor:
New option to block the ability of logging into selected users accounts using the admin passkey.

### DIFF
--- a/auth_admin_passkey/__manifest__.py
+++ b/auth_admin_passkey/__manifest__.py
@@ -11,7 +11,11 @@
     'website': 'https://www.github.com/OCA/server-auth',
     'license': 'AGPL-3',
     'depends': [
+        'base',
         'mail',
+    ],
+    'data': [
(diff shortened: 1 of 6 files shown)


ValentinVinagre (approved):
LGTM 👍

ValentinVinagre:
@HaraldPanten

HaraldPanten:
@gurneyalex @pedrobaeza any clue about why might be runbot failing?

THX!

pedrobaeza:
It's brown, which means a warning, but such warning is not accessible. Maybe Python 3.5 deprecation or a field with same label.

HaraldPanten:
> It's brown, which means a warning, but such warning is not accessible. Maybe Python 3.5 deprecation or a field with same label.

THX Pedro

HaraldPanten (approved):

PR #269, opened 2021-04-28
[14.0][FIX] auth_session_timeout: session timeout applies only to user

dreispt:
Make it so session timeout doe not apply to requests
to a route with auth_method="public".


dreispt:
Forward port of https://github.com/OCA/server-auth/pull/258

sbidoul (approved):

PR #278, opened 2021-07-30
[11.0][MIG] Backport auth_jwt from 14.0 to 11.0

llacroix:
Extracted history from odoo 14 and rebranded for odoo 11.

- Changed the manifest external dependency to point to the jwt module name instead of the pyjwt package name 
- Added pyjwt in requirements.txt
- Remove use of api.create_multi as it's not existant in odoo 11. It shouldn't impact the rest of the code


llacroix:
> Unable to find image 'oca-server-auth:837cbeb_env_2_job_1' locally
> docker: Error response from daemon: pull access denied for oca-server-auth, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
> See 'docker run --help'.

There seems to be an issue with runbot

llacroix:
@sbidoul  any idea what's up with runbot?

llacroix:
Also a bit stuck with pyjwt not supporting python3.5 but it should work just fine with python3.6 is there a way to let travis pass the test even if PyJWKClient can't be imported?

llacroix:
https://app.travis-ci.com/github/OCA/server-auth/jobs/528919126 I also don't quite understand what's the issue with pylint some modules are unrelated but I don't see anything particular about auth_jwt

sbidoul:
@llacroix I fixed the lint issue in #280. You can rebase when it's merge.
Regarding py36 what you have done looks ok. Alternatively you can use python-jose, that's what I've done in the 10.0 backport (#281).

sbidoul (approved):
When green

PR #285, opened 2021-09-08
[12.0][IMP] Add a default Key to the API Key model

danypr92:
This MR generates a random UUID4 value as a key for new API Key instances.
The value is generated with the  uuid4 random generator of UUID package:

https://docs.python.org/3.8/library/uuid.html#uuid.uuid4

Code moved from 'easy_my_coop_api' module of the repository 'coopiteasy/vertical-cooperative':

https://github.com/coopiteasy/vertical-cooperative/blob/12.0/easy_my_coop_api/models/auth_api_key.py#L20

We propose to move this default value to the main `auth_api_key` module instead of having it in`easy_my_coop_api` because we think that this feature can help every user of this module.

### DIFF
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -20,7 +20,7 @@ class AuthApiKey(models.Model):
     name = fields.Char(required=True)
     key = fields.Char(
         required=True,
-        default=_default_key,
+        default=lambda self: self._default_key(),
         help="""The API key. Keep the default value in this field if it is
         obtained from the server environment configuration.""",
     )
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -1,5 +1,6 @@
 # Copyright 2018 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+import uuid
 
 from odoo import api, fields, models, tools, _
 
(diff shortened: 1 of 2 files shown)


danypr92:
@lmignon @rousseldenis Can you review this proposal? 
Thanks!

auth_api_key/models/auth_api_key.py
```diff
@@ -13,10 +14,14 @@ class AuthApiKey(models.Model):
...
     _description = "API Key"
 
+    def _default_key(self):
+        return uuid.uuid4()
+
     name = fields.Char(required=True)
     key = fields.Char(
         required=True,
-        help="""The API key. Enter a dummy value in this field if it is
+        default=_default_key,
```
lmignon: @danypr92 To allow an easy override of the default method, you should avoid to make a direct link to the method into the field definition. `default=lambda self: self._default_key()`

lmignon (changes requested):
@danypr92 Thank you for your contribution. A little comment on my side, otherwise LGTM

auth_api_key/models/auth_api_key.py
danypr92: Thanks @lmignon !!

I push a commit with your suggestion. :rocket:

lmignon (approved):
Thank you for your contribution! LGTM (Code review)

lmignon:
ping @rousseldenis

danypr92:
@rousseldenis, any news?

rousseldenis (approved):

danypr92:
Thanks a lot @rousseldenis and @lmignon!

:heart_eyes: :heart_eyes: :heart_eyes:

lmignon (approved):

PR #286, opened 2021-09-09
[14.0][IMP] auth_api_key: Allow to search on every key even if void

rousseldenis:
If api_key.key is void (due to server_environment config - e.g.: test),
don't raise and break the loop.


rousseldenis:
@lmignon @Cedric-Pigeon

Cedric-Pigeon (approved):
code review

lmignon (approved):

PR #287, opened 2021-09-24
[14.0] Add auth_api_key_group

simahawk:
Allow grouping API keys together.
Grouping per se does nothing. This feature is supposed to be used by other modules
to limit access to services or records based on groups of keys.

CC @gurneyalex

### DIFF
--- a/auth_api_key_group/models/auth_api_key_group.py
+++ b/auth_api_key_group/models/auth_api_key_group.py
@@ -1,0 +1,23 @@
+# Copyright 2021 Camptcamp SA
+# @author: Simone Orsi <simone.orsi@camptocamp.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+
+from odoo import fields, models
+
+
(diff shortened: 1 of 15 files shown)


auth_api_key_group/models/auth_api_key_group.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class AuthApiKeyGroup(models.Model):
+    """Group API keys together."""
+
+    _name = "auth.api.key.group"
+    _description = "API Key auth group"
+
+    name = fields.Char(required=True)
+    code = fields.Char(required=True)
```
sebalix, line 16: I don't know the use case, but you'll probably search on `code`? No need for an index?

sebalix (approved):

auth_api_key_group/models/auth_api_key_group.py
simahawk: not ATM

gurneyalex (approved):

Issue #273, opened 2021-06-16
auth_api_key: License conflict


dreispt:
auth_api_key: The license LGPL-3 conflicts with one of its dependencies. Please make sure that you comply with the licenses compatibility. See our FAQ at https://apps.odoo.com/apps/faq#maintainer_faq_04 .

simahawk:
@lmignon @sebastienbeau could you check this?

lmignon:
Thank you for the issue. We have to check if it's possible to re license server_environment.

simahawk:
https://github.com/OCA/server-env/pull/77 CC @guewen

dreispt:
I work around could be to move the `server_environment` to  a glue module.
And IMO you don't really need `server_environment` to implement API Key support.

simahawk:
OCA/server-env#77 is getting merged. I agree w/ @dreispt that api key could be decoupled from server env. @lmignon what do you think?

lmignon:
@simahawk @dreispt indeed api key could be decoupled from server env. I will be happy to review any PR on this topic

simahawk:
this is solved and server env was decoupled on https://github.com/OCA/server-auth/pull/288

PR #288, opened 2021-09-24
[14.0] Add auth_api_key_server_env

simahawk:
Addresses https://github.com/OCA/server-auth/issues/273

ping @lmignon @dreispt


lmignon (approved):
GREAT! Thank you @simahawk. LGTM (code only)

auth_api_key/migrations/14.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,27 @@
...
+def migrate(cr, version):
+    if not version:
+        return
+
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    module = env["ir.module.module"].search(
+        [("name", "=", "auth_api_key_server_env"), ("state", "=", "uninstalled")]
+    )
+    if module:
+        module.write({"state": "to install"})
```
dreispt, line 22: Nice trick!

auth_api_key_server_env/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+
+This can be very useful to avoid mixing your keys between your various
+environments when restoring databases. All you have to do is to add a new
+section to your configuration file according to the following convention:
+    """,
+    "version": "14.0.1.0.0",
+    "development_status": "Alpha",
+    "license": "LGPL-3",
+    "website": "https://github.com/OCA/server-auth",
+    "author": "Camptcamp,Odoo Community Association (OCA)",
```
dreispt:
```suggestion
    "author": "Camptocamp,Odoo Community Association (OCA)",
```

auth_api_key_server_env/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "depends": ["auth_api_key", "server_environment"],
```
dreispt, line 19: Adding `auto_install : True` would be nice

dreispt (approved):

auth_api_key_server_env/__manifest__.py
lmignon: No auto_install.... IMO it's a choice from the dev/ana to store the api keys into server env even if others informations are managed in this way.

simahawk: yep, that's the reason ;)

PR #294, opened 2021-10-05
[IMP] auth_jwt: add public_or_jwt auth method

sbidoul:
This method is useful for public endpoints that need
to work for anonymous users, but can be enhanced when
an authenticated user is know.

A typical use case is a "add to cart" endpoint that can
work for anonymous users, but can be enhanced by
binding the cart to a known customer when the authenticated
user is known.

TODO:
- [x] test
- [x] update readme
- [x] sanity checks in _authenticate()


lmignon (approved):
Thank you. LGTM (Code review)

PR #305, opened 2021-11-15
[12.0][MIG] auth_oidc

hbrunn:
this backports auth_oidc to v12.

@sbidoul you might be interested in https://github.com/OCA/server-auth/pull/305/commits/982267bcc377583913c75ecb43ea7684d65b4272 to set groups based on claims in the token


sbidoul:
@hbrunn interesting. Would you think it would make sense to do the groups thing in a separate PR (and possibly in a separate module) ?

hbrunn:
if splitting off the new functionality makes it easier to review this, sure I'm happy to do that.

What's your rationale for putting this into a different module? As this code path only happens when people actually configure group lines, I don't see a problem with having it in the base module - it is indeed not directly related to oidc, but then you shouldn't use any other oauth nowadays anyways.

hbrunn:
(as in: Removing this commit and creating a new PR for this I can do tonight, but creating a new module and doing this right will take much longer)

sbidoul:
I'm not sure what's best. At first sight I was a bit surprised with the idea of creating group associations during login. But it might make sens, I need to think about it.

At least extract the non-migration commits to other PR(s) so we can forward port them easily to other versions. I try to do that effort on modules I maintain.

hbrunn:
@sbidoul this took a little longer, but now I also bring more [tests](https://github.com/OCA/server-auth/pull/305/commits/c94d8a954d6b83eaa23e993c7bf6714c1eeaa0b0).

And this seems to expose an [issue](https://runboat.odoo-community.org/api/v1/builds/bd83ada01-7e33-4b2e-85d6-3b9bef657fc9/init-log) with runboat and lower versions, in v12 dependencies aren't the name of the python package but something that can be [imported](https://github.com/OCA/server-auth/blob/c94d8a954d6b83eaa23e993c7bf6714c1eeaa0b0/auth_u2f/__manifest__.py#L37) - should it fall back to requirements.txt in case some dependency can't be pip installed?

I'll rebase my [PR](https://github.com/hbrunn/server-auth/pull/1) for the group stuff when this is merged.

sbidoul:
> in v12 dependencies aren't the name of the python package but something that can be [imported](https://github.com/OCA/server-auth/blob/c94d8a954d6b83eaa23e993c7bf6714c1eeaa0b0/auth_u2f/__manifest__.py#L37)

@hbrunn yes in Odoo < 13 this is a problem. The solution is to declare a mapping of importable names to PyPI package names in setup.py. Look for `external_dependencies_override` in the [setuptools-odoo docs](https://pypi.org/project/setuptools-odoo/).

Doing this is necessary for the install technique that runboat image uses (which comes from the  [oca/oca-ci image](https://github.com/oca/oca-ci)). 
And this will also be necessary for tests to work with GitHub actions when we switch off Travis.
Finally, this is also necessary for the metadata to be correct in the wheels we build.

hbrunn:
thanks, that's now [fixed](https://github.com/OCA/server-auth/pull/340).

And here I'll want to add `responses` to `test_requires` when rebasing after the above is merged (note to self)

requirements.txt
```diff
@@ -8,3 +8,6 @@ python-u2flib-server
 zxcvbn
 # auth_saml
 lasso
+# auth_oidc
+python-jose<3.3.0
```
sbidoul, line 12: @hbrunn Could you add the jose to python-jose mapping in setup.py ?

sbidoul: Hm, python2-secrets is slightly annoying. Let me try something.

hbrunn:
yay, this now should come back green. Thanks

sbidoul (approved):

PR #306, opened 2021-11-16
[13.0][ADD] auth_verification_code_base: Add

kevinkhao:


kevinkhao:
Module is WIP

This module serves as a base for verification codes for login and signup. It allows for configurable confirmation codes to be used on signup and login. 
It is destined for example to be used as:
* User signs up
* Confirmation code is sent by SMS or by email
* User needs to use confirmation code to complete the signup
Same idea for login

CURRENTLY HAS:
* Basic login blocking

NEEDS TO BE IMPLEMENTED:
* Error pages
* Signup blocking
* Clean module split
* Better UX

PR #308, opened 2021-11-24
[MIG][15.0] auth_api_key: Migration to 15.0

yankinmax:


yankinmax:
@lmignon Could you pls do the same https://github.com/OCA/connector/pull/420#issuecomment-978917294 for https://github.com/OCA/server-auth as well ?

lmignon:
@simahawk Why do you've deleted the 15.0 branch? I need it to redo the initialization... Unfortunately, I've no more time today to redo a proper initialization.

simahawk:
@lmignon I take care, don't worry. I checked PRs and I didn't see any TBH :/
The branch is back.

simahawk:
@lmignon the bare init is done already, that works w/ the mt tool. I wanted to generate it again from 14.0 w/ all modules uninstallable but it failed (and to do it you need to delete the branch apparently).

ivantodorovich (approved):
LG

simahawk (approved):
You might want to migrate auth_api_key_sever_env too

simahawk:
@yankinmax seems I screwed history relation when deleting 15.0 (which was the only way to try the MT tool for generating a full branch). Could you rebase your work on current 15.0?

yankinmax:
@dreispt @simahawk can you pls try again to merge this PR?

simahawk:
@sbidoul could we disable caching for pre-commit to avoid such errors? 
```
Error: reserveCache failed: Cache already exists. Scope: refs/heads/15.0-ocabot-merge-pr-308-by-simahawk-bump-nobump, Key: pre-commit-2-6e9292ce7484c9db39ab47966f1138e55cc541e71579416d61487d23def85a68-23b05f1a0616e9488a803945aa248b3968925367ec725f4d0f32e9d3a68935ac, Version: 28cdb9f5496f334116f23e86f0063f5d3a9348c2e22425a33171e071aadada7e
```
https://github.com/OCA/server-auth/runs/4352799172?check_suite_focus=true#step:4:125 

They happen quite often.

sbidoul:
> could we disable caching for pre-commit to avoid such errors?

@simahawk according to https://github.com/pre-commit/action/issues/41 this is a race condition that is not fixable.
So I'm not sure we can fix it ourselves except by disabling the cache (which may or may not create perf issues).
Also the pre-commit GitHub action is unmaintained as the author is developing https://pre-commit.ci/
Last time I checked https://pre-commit.ci/ was not usable for us as it tries to autoupdate the pre-commit configs, which we don't want. Also our big pre-commit runs (with pylint-odoo and such) would probably kill the guy's free-for-opensource budget. But we may want to reach out to see if there are options.

simahawk:
> So I'm not sure we can fix it ourselves except by disabling the cache (which may or may not create perf issues).

Shall we give it a try?

Meanwhile I got this merged here https://github.com/OCA/server-auth/pull/311

PR #316, opened 2021-12-15
[15.0][MIG] auth_jwt: Migration to 15.0

yankinmax:

PR #320, opened 2021-12-29
[15.0][MIG] auth_jwt,auth_jwt_demo: pre-migrate to V15

yankinmax:
This PR is needed to port code from V14 to V15.
Modules will be migrated here:
- [ ] https://github.com/OCA/server-auth/pull/316


yankinmax:
@sbidoul @simahawk 
this PR is related to this suggestion:
https://github.com/OCA/server-auth/pull/316#issuecomment-1002435860

simahawk:
@yankinmax you have to exclude these modules from pre-commit conf

sbidoul:
Ok, so if I understand [this comment](https://github.com/OCA/server-auth/pull/316#issuecomment-1014727922) correctly, this PR should also mark auth_jwt_demo as installable=False ?

yankinmax:
> Ok, so if I understand [this comment](https://github.com/OCA/server-auth/pull/316#issuecomment-1014727922) correctly, this PR should also mark auth_jwt_demo as installable=False ?

I suppose yes. If I'm doing something wrong pls correct me.

simahawk:
@yankinmax fix pre-commit pls

auth_jwt/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Auth JWT",
     "summary": """
         JWT bearer token authentication.""",
-    "version": "14.0.1.2.0",
+    "version": "15.0.1.0.0",
```
sbidoul: Please do not change the version while it is not migrated

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -114,7 +114,7 @@ def _decode(self, token):
                 header = jwt.get_unverified_header(token)
             except Exception as e:
                 _logger.info("Invalid token: %s", e)
-                raise UnauthorizedInvalidToken()
+                raise UnauthorizedInvalidToken() from e
```
sbidoul: This change should not be here yet, as you are just setting the modules as installable=False

auth_jwt_demo/__manifest__.py
```diff
@@ -5,12 +5,13 @@
     "name": "Auth JWT Test",
     "summary": """
         Test/demo module for auth_jwt.""",
-    "version": "14.0.1.2.0",
+    "version": "15.0.1.0.0",
```
sbidoul: Don't change the version yet

sbidoul (changes requested):
Please commit the change to the pre-commit config.
Also, make sure the two commits that set the modules to installable=False are the last in the list and all the other come from git am.

auth_jwt_demo/__manifest__.py
yankinmax: this was done to avoid pre-commit issues

yankinmax:
@sbidoul I'll rebase right now.
1. .pre-commit-config.yaml will leave as after `pre-commit run -a`
2. last two commits with `installable=False`
3. version as it is.

PR #327, opened 2022-01-12
[12.0][FIX] password_security: Allow translations in password constraints

TheRaphael0000:
I've just committed the changed made to password_match_message from OCA/server-auth/14.0, which allows translations

https://github.com/OCA/server-auth/blame/14.0/password_security/models/res_users.py#L84

Please don't take into account the title of the commit.
It's the convention that we use @CompassionCH for our ticketing system.

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -115,7 +115,7 @@ class ResUsers(models.Model):
         if company_id.password_special:
             message.append(
                 _(
-                    "\n* Special character (at least % characters)"
+                    "\n* Special character (at least %s characters)"
                     % str(company_id.password_special)
                 )
             )
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -99,7 +99,8 @@ class ResUsers(models.Model):
             value = int(value)
             if value > 0:
                 at_least_msg = bracket_singular if value <= 1 else bracket_plural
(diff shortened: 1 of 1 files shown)


eicher31 (approved):

password_security/models/res_users.py
```diff
@@ -92,24 +92,41 @@ def password_match_message(self):
         company_id = self.company_id
         message = []
         if company_id.password_lower:
-            message.append('\n* ' + 'Lowercase letter (At least ' + str(
-                company_id.password_lower) + ' character)')
+            message.append(
+                _(
+                    "\n* Lowercase letter (at least %s characters)"
```
valentincastravete, line 97: I think you should use a multiple options word instead of a plural or singular only.
Same with all the other **character** words.
```suggestion
                    "\n* Lowercase letter (at least %s character(s))"
```

primes2h: > @TheRaphael0000 I think you should use a multiple options word instead of a plural or singular only. Same with all the other **"characters"** words.

Even better, this guideline shows how to handle plural forms:
 
https://www.odoo.com/documentation/14.0/developer/misc/i18n/translations.html#plural

ecino (approved):

valentincastravete (approved):

PR #335, opened 2022-01-21
[14.0][FIX] password_security: 2FA workflow not being considered properly

### LINKED ISSUE #333: password_security: Invalid signup token when user has 2FA enabled
## Module
password_security

## Describe the bug
When a user with 2FA enabled tries to login, before he is asked for his confirmation code he will see an error view saying "Invalid signup token".

## To Reproduce
**14.0**:

1. Go to the OdooBot user and change his "password_write_date" to a value that would trigger a neccessary password reset for a normal user, for example 61 days ago with the default configuration of this module (password reset = 60 days), you can do it editing the form view and adding the field or by a server action for example.
2. Enable 2FA in a random user, for example Marc Demo and logout.
3. Try to login with that user and you should see the error message.

**Expected behavior**
The second step of 2FA being shown correctly.

**Additional context**
I tried this with the "try me" button of this module.

IMO the problem might come from this line:
https://github.com/OCA/server-auth/blob/efff3eb63df81657652e33ddfb528ef38500aa28/password_security/controllers/main.py#L43

As I tried it locally and while debugging I saw that request.env.user was different than the actual user being logged in (id=1 -> OdooBot)

Or from the Core Web module:
https://github.com/odoo/odoo/blob/d132c8d4f319af8fe9882d90f3af0d7b4e145b67/addons/web/controllers/main.py#L951

I would say that the uid of the request needs to be changed there before that line In a similar way as in: 
https://github.com/odoo/odoo/blob/d132c8d4f319af8fe9882d90f3af0d7b4e145b67/addons/web/controllers/main.py#L938

This issue might be related with #264

BT-cjimeno:
When a user with 2FA enabled tried to log in, the user whose password expiration was being checked
was not the actual user logging in, instead it was OdooBot. Some changes were introduced to have
the 2fa module (auth_totp) into account.

Fixes #333

### DIFF
--- a/password_security/__manifest__.py
+++ b/password_security/__manifest__.py
@@ -16,6 +16,7 @@
     "depends": [
         "auth_signup",
         "auth_password_policy_signup",
+        "auth_totp",
     ],
     "website": "https://github.com/OCA/server-auth",
     "external_dependencies": {
--- a/password_security/controllers/__init__.py
+++ b/password_security/controllers/__init__.py
@@ -2,3 +2,4 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 from . import main
+from . import home
--- a/password_security/controllers/home.py
+++ b/password_security/controllers/home.py
@@ -1,0 +1,25 @@
+# Copyright 2022 brain-tec AG (https://bt-group.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import http
+from odoo.http import request
+
+from odoo.addons.auth_totp.controllers.home import Home
+
+
+class PasswordSecurity2FAHome(Home):
+    @http.route()
+    def web_totp(self, redirect=None, **kwargs):
+        already_logged_in = request.session.uid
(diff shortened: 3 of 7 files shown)


password_security/__manifest__.py
```diff
@@ -16,6 +16,7 @@
     "depends": [
         "auth_signup",
         "auth_password_policy_signup",
+        "auth_totp"
```
BT-aestebanez: Add "," at the end to ease traceability

BT-aestebanez: Add a comment explaining why you added this dependency

BT-aestebanez (changes requested):

BT-aestebanez (approved):

BT-aestebanez (approved):

BT-cjimeno:
@dreispt @Chandresh-OSI Hi, would you have time for a review? Thanks!

BT-cjimeno:
@dreispt @Chandresh-OSI Hi, would you have time for a review? Thanks!

BT-cjimeno:
Hi @dreispt , could you please add the "no stale" label to this PR? Thanks!

Issue #325, opened 2022-01-12
[14.0] auth_oidc with authorization flow should call validation endpoint 


stellamargonar:
**Module auth_oidc, Odoo 14.0**
I have a OpenID provider configured in Odoo after installing the `auth_oidc` addon from this repository. Currently i'm working on Odoo 14, using https://github.com/OCA/server-auth/pull/261 and the provider is using OpenID Connect (authorisation code flow)

It works, meaning that users can now login in Odoo with the OpenID provider, and the first time the user is created with the default template.
The problem is that the only information that Odoo reads from the provider are the ones in the token, which is only the user id. Even though i configured all the endpoints (authentication, validation, token, jwks and data) the addon is not calling the validation and data (they are the same anyway) endpoint: this api would return all the user claims the client has access to, including email, that would be necessary to properly configure the user in Odoo.

[Accordingly to the OpenID protocol](https://openid.net/specs/openid-connect-basic-1_0.html#rfc.section.1.3) seems like after the authorisation phase, the client should also send a request to the userinfo endpoint, but this is not implemented in `auth_oidc`.

On the other side, i see that this particular call was present in the [`auth_oauth` addon](https://github.com/odoo/odoo/blob/5ab778227dfbc93b2ff415ebb0e40ce35953f13e/addons/auth_oauth/models/res_users.py#L96).

I would suggest to add to the implementation of `addons.auth_oidc.models.res_users.ResUsers.auth_oauth` a call to the validation endpoint (if provided) similarly to what is done in `auth_oauth` addon, and update the `validation` dictionary with the result of the call.

sbidoul:
Sounds reasonable. Would you like to do a PR ?

stellamargonar:
Sure, here it is https://github.com/OCA/server-auth/pull/336

PR #336, opened 2022-01-21
[14.0] auth_oidc add call to validation endpoint

stellamargonar:
add call to oauth provider validation endpoint in the auth_oidc module, as explained in https://github.com/OCA/server-auth/issues/325

### DIFF
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -63,6 +63,10 @@ class ResUsers(models.Model):
             _logger.error("No id_token in response.")
             raise AccessDenied()
         validation = oauth_provider._parse_id_token(id_token, access_token)
+
+        if oauth_provider.validation_endpoint:
+            validation.update(self._auth_oauth_validate(oauth_provider, access_token))
+
         # required check
         if not validation.get("user_id"):
             _logger.error("user_id claim not found in id_token (after mapping).")
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -65,7 +65,7 @@ class ResUsers(models.Model):
         validation = oauth_provider._parse_id_token(id_token, access_token)
 
         if oauth_provider.validation_endpoint:
-            validation.update(self._auth_oauth_validate(oauth_provider, access_token))
+            validation.update(self._auth_oauth_validate(provider, access_token))
 
         # required check
         if not validation.get("user_id"):


ap-wtioit:
needed this to get auth_oidc working with OpenID Connect. [ID Token](https://openid.net/specs/openid-connect-basic-1_0.html#IDToken) does not contain a `user_id` claim. nor is `user_id` listed in the [standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims).

Note: the call to the validation endpoint is not necessary from OpenID Connect Authorization Code Flow (because the ID Token can be and is verified with JWKS according to https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.3.1.3.7)

however auth_oidc/auth_oauth uses the non standard claim `user_id` which makes this necessary. but it only works (with standard claims) because auth_oauth maps `sub` to `user_id` in `_auth_oauth_validate`:
```python
         subject = next(filter(None, [
            validation.pop(key, None)
            for key in [
                'sub', # standard
                'id', # google v1 userinfo, facebook opengraph
                'user_id', # google tokeninfo, odoo (tokeninfo)
            ]
        ]), None)
        ...
        validation['user_id'] = subject
```
https://github.com/odoo/odoo/blob/14.0/addons/auth_oauth/models/res_users.py#L56

@stellamargonar would something like this also work for the errors you encountered in https://github.com/OCA/server-auth/issues/325?
```python
        ....
        # required check
        if "sub" in validation and not "user_id" in validation:
            # set user_id for auth_oauth, user_id is not an OpenID Connect standard claim:
            # https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
            validation["user_id"] = validation["sub"]
        elif not validation.get("user_id"):
            _logger.error("user_id claim not found in id_token (after mapping).")
            raise AccessDenied()
        ...
```

I tested locally together with https://github.com/OCA/server-auth/pull/393 in 15.0 against our oauth/openid provider.

For reference the error i encountered without this patch:
```
odoo_1                                                | 2022-08-09 11:45:00,364 1 ERROR devel odoo.addons.auth_oidc.models.res_users: user_id claim not found in id_token (after mapping). 
odoo_1                                                | 2022-08-09 11:45:00,364 1 INFO devel odoo.addons.auth_oauth.controllers.main: OAuth2: access denied, redirect to main page in case a valid session exists, without setting cookies 
```

PR #338, opened 2022-01-27
[IMP] timeout check triggered on session uid

gfcapalbo:
Triggering timeout check on request session.uid, because env.user may be null.

env,user._auth_timeout_check  will work with an empty recordset ( it is @api.model_cr_context)


BT-ssteiner:
This also applies to higher versions

BT-aestebanez (approved):
LGTM

ghost:
LGTM

BT-ssteiner (approved):
did a functional test. LGTM

ghost (approved):

PR #339, opened 2022-01-29
[FIX] auth_u2f correct dependency

kh-badep:


sbidoul:
Resolved in #340, so closing this one.

sbidoul:
See https://github.com/OCA/server-auth/pull/305#issuecomment-1037119227 for the explanation of the solution.

PR #337, opened 2022-01-27
[15.0][MIG] auth_saml: Migrate to version 15.0

Camille0907:
Standard migration

Issue #321, opened 2021-12-29
[14.0] auth_saml : saml2.sigver.SignatureError: Signature missing for assertion with Google SAML IDP


oussjarrousse:
I am getting the error `Signature missing for assertion` in the odoo logs. 

I am using Google as IdP, and Odoo 14 with the auth_saml as SP. I created a SAML provider and added the required settings and all looks good if I don't check the "Sign" checkbox (although I had to add a dummy cert and a dummy key for the module to work but that is another issue #315, I also had to use another Dockerfile that inherits from the odoo official image to install the requirement as in #295). 

After some investigation it seems to me that the assertion from Google is not signed while the `want_assertions_signed` in the `sp` context for the `Saml2Config` instance is set to `self.sign` which is `True` when the `Sign` checkbox is checked.

I managed to "fix" it by manually altering line 181 in `auth_saml/models/auth_saml_provider.py` from: `"want_assertions_signed": self.sign,` to `"want_assertions_signed": False,`

## Module
auth_saml

## Describe the bug
When using Google as an IdP, and setting cert and key and checking the "sign" check box on Odoo auth_saml saml provider.
Here is the error output:

```
2021-12-29 18:44:22,720 1 INFO xxxxxx saml2.response: status: <ns0:Status xmlns:ns0="urn:oasis:names:tc:SAML:2.0:protocol"><ns0:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success" /></ns0:Status> 
odoo       | 2021-12-29 18:44:22,722 1 ERROR xxxxxx saml2.client_base: XML parse error: Signature missing for assertion 
odoo       | 2021-12-29 18:44:22,722 1 ERROR xxxxxx odoo.addons.auth_saml.controllers.main: SAML2: failure - Signature missing for assertion 
odoo       | Traceback (most recent call last):
odoo       |   File "/mnt/extra-addons/auth_saml/controllers/main.py", line 184, in signin
odoo       |     request.httprequest.url_root.rstrip("/"),
odoo       |   File "/mnt/extra-addons/auth_saml/models/res_users.py", line 94, in auth_saml
odoo       |     validation = self._auth_saml_validate(provider, saml_response, base_url)
odoo       |   File "/mnt/extra-addons/auth_saml/models/res_users.py", line 44, in _auth_saml_validate
odoo       |     return provider._validate_auth_response(token, base_url)
[...]

Issue #315, opened 2021-12-13
[14.0] auth_saml : cert and key mandatory even if no signature needed


remi-filament:
## Module
auth_saml

## Describe the bug
When adding a SAML provider, in case you do not want your requests to IdP signed, you do not need cert and private key.
However, if you do not generate these, you will not be able to generate metadata (you get a 500 Internal Server Error).
Also, Signature Algorithm is mandatory and should not be if Sign checkbox is not ticked.

## To Reproduce
**Affected versions**: 14.0

1. Create new SAML provider
2. Fill in IdP metadata
3. Save
4. Click on the link generated for Odoo metadata (in Metadata URL) --> you get a 500 Internal Server Error)

**Expected behavior**
You should be able to generate metadata without setting certificate / private key in case you do not need to sign requests.
In any case, you may want to add some validation checks (which today raise error on logs when trying to access metadata but are not visible to Odoo user) :
- check that both private key and certificate are provided in case you tick Sign checkbox
- validate XML for IdP metadata

Eventually, xmlsec1 is required by pysaml2 (this information is present in readme/INSTALL.rst but not in generated README.rst, which may need to be re-generated).

**Additional context**
```
odootest | 2021-12-13 10:26:30,792 1 ERROR filament10_test werkzeug: Error on request:
odootest | Traceback (most recent call last):
odootest |   File "/opt/odoo/odoo/odoo/addons/base/models/ir_http.py", line 237, in _dispatch
odootest |     result = request.dispatch()
odootest |   File "/opt/odoo/odoo/odoo/http.py", line 807, in dispatch
odootest |     r = self._call_function(**self.params)
odootest |   File "/opt/odoo/odoo/odoo/http.py", line 360, in _call_function
odootest |     return checked_call(self.db, *args, **kwargs)
odootest |   File "/opt/odoo/odoo/odoo/service/model.py", line 94, in wrapper
odootest |     return f(dbname, *args, **kwargs)
odootest |   File "/opt/odoo/odoo/odoo/http.py", line 348, in checked_call
odootest |     result = self.endpoint(*a, **kw)
odootest |   File "/opt/odoo/odoo/odoo/http.py", line 913, in __call__
odootest |     return self.method(*args, **kw)
odootest |   File "/opt/odoo/odoo/odoo/http.py", line 532, in response_wrap
odootest |     response = f(*args, **kw)
odootest |   File "/opt/odoo/additional_addons/auth_saml/controllers/main.py", line 231, in saml_metadata
[...]

PR #107, opened 2019-06-03
[11.0][IMP] auth_saml

vincent-hatakeyama:

PR #342, opened 2022-02-17
[15.0][FIX][IMP] Fix and improve auth_saml

vincent-hatakeyama:
🚑 Fix the module by adding a transaction to commit the token
🚑 Fix the disallow password for users with SAML ids
Added tests to ensure the feature works correctly.
Admin user is also an exception from not having a password. In Odoo 15.0, this is the standard user to connect for administrative task, not the super user.
✨ Improve provider form and list views
✨⏩ port of 11.0 automatic redirection from 11.0 version. Use disable_autoredirect as a parameter query to disable automatic redirection (for example https://example.com/web/login?disable_autoredirect=)
💄 Add certificate file name fields to improve the UI
📝 Add required on several fields of the SAML provider; without them the server will crash and there is not enough information to make SAML work.
✨ Split signing to have finer control and be compatible with more IDP.
🔨 Integrate token into res.users.saml, removing auth_saml.token. No need for a separate table, and no more need to create lines in the table.
📝 Avoid server errors when user try metadata page without necessary parameters.
🚑 Replace method call from odoo.http.redirect_with_hash to request.redirect as the former does not exists in Odoo 15.0 anymore.
📚 Improved the module documentation
👕 pylint fixes and other fixes or minor changes

I’ve tested the module with a local keycloak. As I wrote in https://github.com/OCA/server-auth/issues/337#issuecomment-1040324260 the module did not work correctly, the base functionality was not working.
I’ve also fixed the allow password option and added tests.
My changes for the sign attribute on the provider comes from testing, where keycloak does not handle all the default signing, and from https://github.com/OCA/server-auth/issues/321 indicating other IdP can have the same kind of needs.
I’ve ported the automatic redirection to a SAML provider from never merged https://github.com/OCA/server-auth/pull/107

I’ve updated the README manually because it does not seem to be generated automatically. It might be missing some files in the readme directory, so I’ve added them. Same for the `static/description/index.html`file.

https://github.com/OCA/server-auth/issues/315 also affects this version, so I’ve made the certificates mandatory for now.


damdam-s (approved):
thanks

simahawk:
Thanks for your contrib!
I'm not the maintainer of this module but 1 commit only for all these changes doesn't look that cool.
Any chance you could split them?
Regarding the readme: if you update it manually it won't be kept.

auth_saml/__manifest__.py
```diff
@@ -24,4 +26,5 @@
     ],
     "installable": True,
     "auto_install": False,
+    "development_status": "Production/Stable",
```
simahawk: is this really stable after all this refactoring?

auth_saml/controllers/main.py
```diff
@@ -211,13 +245,14 @@ def signin(self, req, **kw):
...
     @http.route("/auth_saml/metadata", type="http", auth="none", csrf=False)
+    # pylint: disable=unused-argument
     def saml_metadata(self, req, **kw):
         provider = kw.get("p")
         dbname = kw.get("d")
         valid = kw.get("valid", None)
 
         if not dbname or not provider:
-            raise Exception("Missing parameters")
+            return request.not_found(_("Missing parameters"))
```
simahawk, line 255: you could log an error to help debugging such cases

auth_saml/models/res_users.py
```diff
@@ -21,76 +24,44 @@ class ResUser(models.Model):
...
+    def _auth_saml_validate(self, provider_id: int, token: str, base_url: str = None):
         provider = self.env["auth.saml.provider"].sudo().browse(provider_id)
         return provider._validate_auth_response(token, base_url)
 
-    def _auth_saml_signin(self, provider, validation, saml_response):
-        """retrieve and sign into openerp the user corresponding to provider
-        and validated access token
+    def _auth_saml_signin(self, provider: int, validation: dict, saml_response) -> str:
+        """Retrieve and sign in Odoo the user corresponding to provider and validated
+        access token.
```
simahawk:
```suggestion
        """Sign in Odoo user corresponding to provider and validated access token.
```

auth_saml/models/res_users.py
```diff
@@ -130,74 +101,79 @@ def _check_credentials(self, password, env):
...
-            ).write({"password": self._autoremove_password_if_saml_gen_password()})
+            if token:
+                return
+            raise AccessDenied() from None
 
     @api.model
-    def _autoremove_password_if_saml_gen_password(self):
+    def _allowed_user_ids(self) -> Set[int]:
+        """Users that can have a password even if the option to disallow it is set.
+        It includes superuser and the admin if it exists.
```
simahawk:
```suggestion
    def _saml_allowed_user_ids(self) -> Set[int]:
        """Users that can have a password even if the option to disallow it is set.

        It includes superuser and the admin if it exists.
```

auth_saml/models/res_users.py
```diff
@@ -130,74 +101,79 @@ def _check_credentials(self, password, env):
...
-                )
-                if not token:
-                    model_token.create(
-                        {
-                            "user_id": record.id,
-                            "saml_provider_id": provider.saml_provider_id.id,
-                        }
+    def _set_password(self):
+        """Redefine base method to block setting password on users with SAML ids
+        And also to be able to set password to a blank value"""
```
simahawk: sounds like a comment rather than a docstring

auth_saml/models/res_users.py
```diff
@@ -130,74 +101,79 @@ def _check_credentials(self, password, env):
...
+        if blank_password_users:
+            # similar to what Odoo does in Users._set_encrypted_password
+            self.env.cr.execute(
+                "UPDATE res_users SET password = NULL WHERE id IN %s",
+                (tuple(blank_password_users.ids),),
+            )
+            self.invalidate_cache(["password"], blank_password_users.ids)
+        return
+
+    def allow_saml_and_password_changed(self):
```
simahawk, line 165: this can be called from RPC. Is it safe?

auth_saml/models/res_users_saml.py
```diff
@@ -18,3 +25,11 @@ class ResUserSaml(models.Model):
             "SAML UID must be unique per provider",
         )
     ]
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Redefined to remove password if necessary"""
```
simahawk: that's a comment

auth_saml/readme/CONFIGURE.rst
```diff
@@ -1,5 +1,16 @@
...
    (Settings > Users & Companies > SAML Providers).
 #. Pre-create your users and set the SAML information against the user.
+
+By default, the module let users have both a password and SAML ids.
+To increase security, disable passwords by using the option in Settings.
+Note that the admin account can still have a password, even if the option is activated.
+Setting the option immediately remove all password from users with a configured SAML ids.
+
+If all the users have a SAML id in a single provider, you can set automatic redirection
+in the provider settings. It is still possible to access the login without redirection
```
simahawk: maybe worth to mention that the order (sequence) of the providers matters here.

auth_saml/tests/__init__.py
```diff
@@ -1,2 +1 @@
-from . import fake_idp
```
simahawk, line 1: any reason for such changes?

auth_saml/tests/test_pysaml.py
```diff
@@ -39,6 +47,44 @@ def setUp(self):
 
         self.idp = FakeIDP([self.saml_provider._metadata_string()])
 
+        # Create a user with only password, and another with both password and saml id
+        self.user, self.user2 = (
+            self.env["res.users"]
+            .with_context(no_reset_password=True)
```
simahawk:
```suggestion
            .with_context(no_reset_password=True, tracking_disable=True)
```

auth_saml/tests/test_pysaml.py
```diff
@@ -39,6 +47,44 @@ def setUp(self):
...
+                                },
+                            )
+                        ],
+                    },
+                ]
+            )
+        )
+
+    #
+    # def test_has_password(self):
```
simahawk: still needed?

simahawk (changes requested):

auth_saml/__manifest__.py
vincent-hatakeyama: I could put it back to beta and push a PR after it is in production in a month or two.

vincent-hatakeyama:
> Thanks for your contrib! I'm not the maintainer of this module but 1 commit only for all these changes doesn't look that cool.  Any chance you could split them?

I have been asked to squash history in the past, I believe it is still the process at OCA.

> Regarding the readme: if you update it manually it won't be kept.

All changes have been put in the readme/ directory too.

auth_saml/controllers/main.py
vincent-hatakeyama: This error happens when a user manually enters a URL, or an URL is not valid anymore.

As the end point is public, I don’t want to clutter the log too much. I’ll add something at the debug level.

Note that before my changes here, the page showed a Internal Exception Error.

auth_saml/tests/test_pysaml.py
vincent-hatakeyama: No, forgot to clean up that part.

auth_saml/tests/__init__.py
vincent-hatakeyama: That was changed by isort.

auth_saml/tests/test_pysaml.py
vincent-hatakeyama: Also changed in test_disallow_user_password_no_password_set.

auth_saml/models/res_users.py
vincent-hatakeyama: It only reapplies the policy already present in the database, as it uses `self.allow_saml_and_password()` and not a context value or an argument ; so yes it is safe.

auth_saml/readme/CONFIGURE.rst
vincent-hatakeyama: I added the information.

simahawk:
> I have been asked to squash history in the past, I believe it is still the process at OCA.

Normally we ask to squash fixup commits or little changes on top of another. Here we have a lot of work squashed on a commit that says only " Fix the module by adding a transaction to commit the token " which is just  a tiny piece of your work :wink:

gurneyalex (approved):

auth_saml/controllers/main.py
```diff
@@ -56,19 +56,47 @@ class SAMLLogin(Home):
...
+        return None
+
+    @http.route()
+    def web_client(self, s_action=None, **kw):
+        ensure_db()
+        if not request.session.uid:
+            result = self._saml_autoredirect()
+            if result:
+                return result
+            return super().web_client(s_action, **kw)
```
ivantodorovich:
```suggestion
```
redundant

ivantodorovich:
```suggestion
```
redundant

auth_saml/controllers/main.py
```diff
@@ -56,19 +56,47 @@ class SAMLLogin(Home):
     def _list_saml_providers_domain(self):
         return []
 
-    def list_providers(self):
-        providers = (
-            request.env["auth.saml.provider"]
-            .sudo()
-            .search_read(self._list_saml_providers_domain())
-        )
+    def list_providers(self, with_autoredirect: bool = False) -> models.Model:
```
ivantodorovich: The new argument makes it incompatible with core module `auth_oauth`.
I'm getting the following traceback:

<details>

```python-traceback
  File "/odoo/src/odoo/addons/base/models/ir_http.py", line 237, in _dispatch
    result = request.dispatch()
  File "/odoo/src/odoo/http.py", line 811, in dispatch
    r = self._call_function(**self.params)
  File "/odoo/src/odoo/http.py", line 360, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/odoo/src/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/odoo/src/odoo/http.py", line 349, in checked_call
    result = self.endpoint(*a, **kw)
  File "/odoo/src/odoo/http.py", line 917, in __call__
    return self.method(*args, **kw)
  File "/odoo/src/odoo/http.py", line 536, in response_wrap
    response = f(*args, **kw)
  File "/odoo/src/addons/portal/controllers/web.py", line 26, in web_client
    return super(Home, self).web_client(s_action, **kw)
  File "/odoo/src/odoo/http.py", line 536, in response_wrap
    response = f(*args, **kw)
  File "/odoo/external-src/server-auth/auth_saml/controllers/main.py", line 94, in web_client
    result = self._saml_autoredirect()
  File "/odoo/external-src/server-auth/auth_saml/controllers/main.py", line 78, in _saml_autoredirect
    autoredirect_providers = self.list_providers(True)
Exception

The above exception was the direct cause of the following exception:

  File "/usr/local/lib/python3.9/dist-packages/werkzeug/serving.py", line 306, in run_wsgi
    execute(self.server.app)
  File "/usr/local/lib/python3.9/dist-packages/werkzeug/serving.py", line 294, in execute
    application_iter = app(environ, start_response)
  File "/odoo/src/odoo/service/server.py", line 476, in app
    return self.app(e, s)
  File "/odoo/src/odoo/service/wsgi_server.py", line 112, in application
    return application_unproxied(environ, start_response)
  File "/odoo/src/odoo/service/wsgi_server.py", line 87, in application_unproxied
    result = odoo.http.root(environ, start_response)
  File "/odoo/src/odoo/http.py", line 1314, in __call__
    return self.dispatch(environ, start_response)
  File "/odoo/src/odoo/http.py", line 1280, in __call__
    return self.app(environ, start_wrapped)
  File "/usr/local/lib/python3.9/dist-packages/werkzeug/middleware/shared_data.py", line 220, in __call__
    return self.app(environ, start_response)
  File "/odoo/src/odoo/http.py", line 1508, in dispatch
    result = ir_http._dispatch()
  File "/odoo/src/addons/auth_signup/models/ir_http.py", line 19, in _dispatch
    return super(Http, cls)._dispatch()
  File "/odoo/external-src/odoo-cloud-platform/monitoring_statsd/models/ir_http.py", line 16, in _dispatch
    return super()._dispatch()
  File "/odoo/src/addons/web_editor/models/ir_http.py", line 21, in _dispatch
    return super(IrHttp, cls)._dispatch()
  File "/odoo/src/addons/utm/models/ir_http.py", line 29, in _dispatch
    response = super(IrHttp, cls)._dispatch()
  File "/odoo/src/addons/http_routing/models/ir_http.py", line 515, in _dispatch
    result = super(IrHttp, cls)._dispatch()
  File "/odoo/src/odoo/addons/base/models/ir_http.py", line 241, in _dispatch
    return cls._handle_exception(e)
  File "/odoo/src/addons/utm/models/ir_http.py", line 34, in _handle_exception
    response = super(IrHttp, cls)._handle_exception(exc)
  File "/odoo/src/addons/http_routing/models/ir_http.py", line 611, in _handle_exception
    return super(IrHttp, cls)._handle_exception(exception)
  File "/odoo/src/odoo/addons/base/models/ir_http.py", line 209, in _handle_exception
    return request._handle_exception(exception)
  File "/odoo/src/odoo/http.py", line 749, in _handle_exception
    return super(HttpRequest, self)._handle_exception(exception)
  File "/odoo/src/odoo/http.py", line 302, in _handle_exception
    raise exception.with_traceback(None) from new_cause
TypeError: list_providers() takes 1 positional argument but 2 were given - - -

```

</details>


https://github.com/odoo/odoo/blob/8b4d7fa3ae9d470bd12d5c09d6270a637311c0c4/addons/auth_oauth/controllers/main.py#L54

Maybe it has always been incompatible though? :thinking:
Could we rename this method to `_list_saml_providers` maybe?

vincent-hatakeyama: It has been named like that since version 11.0. It should indeed be renamed to avoid conflicts.

vincent-hatakeyama: I’ve renamed it `list_saml_providers` as the module was designed to be similar to auth_oauth.

ivantodorovich (approved):

auth_saml/controllers/main.py
vincent-hatakeyama: I’ve removed the redundant code. (see https://github.com/OCA/server-auth/pull/342/files#diff-d051ee3f0ce073219b1c7ebc99bfa35206c9ebeefbf62df714fb8c800ad0310cR97)

gurneyalex (approved):
Tested with Azure Active Directory. Works well.

gurneyalex:
oca bot issue should be solved by https://github.com/OCA/server-auth/pull/365

yvaucher:
Feedback on

> 📝 Add required on several fields of the SAML provider; without them the server will crash and there is not enough information to make SAML work.

This breaks auth_saml_environment module where the 2 key fields becomes non mandatory as we provide an alternative field with a path instead.

Also seems https://github.com/OCA/server-auth/issues/315 is not addressed correctly as the required key fields seem to be conditional depending on signature.

vincent-hatakeyama:
The fields were changed to mandatory to avoid the issue, not to solve it. (I indicated *for now* but nobody proposed a fix since then).

PR #343, opened 2022-02-17
[14.0] [IMP] auth_jwt: Add validator.next_validator_id to allow validator chaining

paradoxxxzero:
This PR adds `next_validator_id` field to `auth.jwt.validator` to allow multiple jwt validators to be checked in chain.

This PR also removes the cursor creation for the validator execution since it prevents to implement for instance an auto create partner strategy being on a different transaction. (It also remove the now useless validator patch in tests).

### DIFF
--- a/auth_jwt/exceptions.py
+++ b/auth_jwt/exceptions.py
@@ -30,3 +30,17 @@ class UnauthorizedInvalidToken(Unauthorized):
 
 class UnauthorizedPartnerNotFound(Unauthorized):
     pass
+
+
+class CompositeJwtError(Unauthorized):
+    """Indicate that multiple errors occurred during JWT chain validation."""
+
+    def __init__(self, errors):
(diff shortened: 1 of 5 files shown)


sbidoul (commented):
Hmmm why did I use a different cursor? I must have had a good reason for that.

paradoxxxzero:
> Hmmm why did I use a different cursor? I must have had a good reason for that.

I wondered the same thing but it creates problem for tests and automatic partner creation. (#346)

hparfr (approved):
LGTM

sbidoul (approved):

sbidoul:
> Hmmm why did I use a different cursor? I must have had a good reason for that.

Alright, I can't remember so maybe there was no good reason. We'll see.

Thanks for the improvement!

/ocabot merge major

PR #346, opened 2022-03-01
[14.0] [IMP] auth_jwt: Add email_create partner strategy

paradoxxxzero:
This new strategy will automatically create unknown partners.
The automatically created partners will have a `created_by_jwt` flag set to True.
All partners will now have a `auth_jwt_email` set at first auth link which can be different from partner email, similar to how core `auth_oauth` addon handles oauth uid.

### DIFF
--- a/auth_jwt/models/auth_jwt_validator.py
+++ b/auth_jwt/models/auth_jwt_validator.py
@@ -5,7 +5,7 @@ import logging
 from functools import partial
 
 import jwt  # pylint: disable=missing-manifest-dependency
-from jwt import PyJWKClient
+from jwt import PyJWKClient  # pylint: disable=missing-manifest-dependency
 from werkzeug.exceptions import InternalServerError
 
 from odoo import _, api, fields, models, tools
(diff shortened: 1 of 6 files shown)


sbidoul:
Can you rebase ?

paradoxxxzero:
All green!

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -170,17 +175,48 @@ def _get_and_check_uid(self, payload):
...
+            partner = self.env["res.partner"].search(
+                [("email", "=", email), ("auth_jwt_email", "=", False)]
+            )
+
+        return partner
+
     def _get_partner_id(self, payload):
         # override for additional strategies
-        if self.partner_id_strategy == "email":
+        if self.partner_id_strategy in ["email", "email_create"]:
```
sbidoul, line 202: Could you implement the new strategy in a separate branch? I'm not too comfortable with the mixing of the two strategies, as it makes the code harder to read. And since this is security sensitive, it's even more important to keep the code as simple as possible.

auth_jwt/models/res_partner.py
```diff
@@ -0,0 +1,8 @@
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    created_by_jwt = fields.Boolean("Created by JWT", default=False)
+    auth_jwt_email = fields.Char("Auth JWT Email")
```
sbidoul, line 8: Why do we need these two fields? Can't we do the same with only one of the two?

sbidoul (changes requested):
I'm wondering if it would not be better to implement this strategy in a separate module, for the sake of keeping the core module as simple as possible to audit and maintain.

PR #350, opened 2022-03-06
[IMP][12.0] auth_oidc: allow assign groups from token claims

hbrunn:
this is the oauth equivalent of [users_ldap_groups](https://github.com/OCA/server-auth/tree/13.0/users_ldap_groups), allowing to also configure all authorizations in your identity provider.

### DIFF
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -105,8 +105,12 @@ class AuthOauthProviderGroupLine(models.Model):
         self.ensure_one()
 
         class Defaultdict2(collections.defaultdict):
+            """Class keeping yielding defaultdicts"""
             def __init__(self, *args, **kwargs):
                 super().__init__(Defaultdict2, *args, **kwargs)
+
+            def __eq__(self, other):
+                return False
 
         return tools.safe_eval(
             self.expression, {
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -2,12 +2,13 @@
 # Copyright 2021 ACSONE SA/NV <https://acsone.eu>
 # License: AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
+import collections
 import logging
 import secrets
 
 import requests
 
-from odoo import fields, models, tools
+from odoo import api, exceptions, fields, models, tools
 
 try:
     from jose import jwt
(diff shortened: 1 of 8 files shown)


auth_oidc/views/auth_oauth_provider.xml
```diff
@@ -19,6 +19,16 @@
                 <field name="token_endpoint" />
                 <field name="jwks_uri" />
             </field>
+            <xpath expr="//sheet/group[last()]" position="after">
+                <group name="mappings">
+                    <field name="group_line_ids">
+                        <tree editable="bottom">
+                            <field name="group_id" />
+                            <field name="expression" />
```
sbidoul: Some help / examples about what to fill-in in `expression` would be welcome.

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -80,3 +84,33 @@ def _parse_id_token(self, id_token, access_token):
...
+                this._eval_expression(self.env.user, {})
+            except (AttributeError, KeyError, NameError) as e:
+                raise exceptions.ValidationError('\n'.join(e.args))
+
+    def _eval_expression(self, user, token):
+        self.ensure_one()
+
+        class Defaultdict2(collections.defaultdict):
+            def __init__(self, *args, **kwargs):
+                super().__init__(Defaultdict2, *args, **kwargs)
```
sbidoul, line 110: Why this Defaultdict2 ?

sbidoul (commented):
LGTM. Code review only.

auth_oidc/models/auth_oauth_provider.py
hbrunn: it yields defaultdicts again that in turn yield defaultdicts. This way the user can write expressions in a quite carefree manner like `token['key1']['key2']['key3'] == 'something'` and it doesn't fail when `key1` doesn't exist

hbrunn: did I do something stupid here?

auth_oidc/views/auth_oauth_provider.xml
hbrunn: would

```suggestion
                            <field name="expression" placeholder="token['scope']['value'] == 'hello world'" />
```

suffice?

auth_oidc/models/auth_oauth_provider.py
sbidoul: Nope, just too clever for me :)

auth_oidc/views/auth_oauth_provider.xml
sbidoul: Good enough for me :)

auth_oidc/models/auth_oauth_provider.py
hbrunn: just added a comment there

auth_oidc/views/auth_oauth_provider.xml
hbrunn: and took this one

ap-wtioit (approved):
LGTM

sbidoul (approved):
I had lost track of this. SGTM.
It'd be nice to have a little word about the feature in the README, but not blocking as the view is explicit enough.

PR #364, opened 2022-03-30
[13.0][password_security] : Missing translations in password constraints

MarineGCH:
Impacted versions: 13.0
The two problems identified concern the front and the back-office

### **First problem:**

When we are in a language other than English and the password is wrong, the warning message displayed is not translated and it always remains in English regardless of the language of the user.

_Exemple in french :_
image

Steps to reproduce: 

1. Keep module default settings
2. Change the admin user’s language for the back-office test and change the portal user’s language for the front-end test (any language other than English)
3. Try to change a user’s password by 1234 in front and back-office

### **Second problem :**

When the password is right but not complex enough, a suggestion should be displayed but this message category is not returned. So we get an empty message.
image

Steps to reproduce:

1. Go to settings > Scroll until Users part
2. Change the module settings values to those on the capture below
image

3. Try to change a user’s password, in front and back-office,  by one of those written below :
o5G4d1m6
K528f846

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -80,47 +80,180 @@ class ResUsers(models.Model):
 
     @api.model
     def get_estimation(self, password):
-        return zxcvbn.zxcvbn(password)
+        msg_translate = {
+            "Use a few words, avoid common phrases.": _(
+                "Use a few words, avoid common phrases."
+            ),
+            "No need for symbols, digits, or uppercase letters.": _(
+                "No need for symbols, digits, or uppercase letters."
+            ),
+            "Add another word or two. Uncommon words are better.": _(
+                "Add another word or two. Uncommon words are better."
+            ),
+            "Straight rows of keys are easy to guess.": _(
+                "Straight rows of keys are easy to guess."
+            ),
+            "Short keyboard patterns are easy to guess.": _(
+                "Short keyboard patterns are easy to guess."
+            ),
+            "Use a longer keyboard pattern with more turns.": _(
+                "Use a longer keyboard pattern with more turns."
+            ),
+            'Repeats like "aaa" are easy to guess.': _(
+                'Repeats like "aaa" are easy to guess.'
+            ),
+            'Repeats like "abcabcabc" are only slightly harder to guess than "abc".': _(
+                'Repeats like "abcabcabc" are only slightly harder  to guess than '
+                '"abc".'
+            ),
+            "Avoid repeated words and characters.": _(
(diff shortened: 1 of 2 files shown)


MarineGCH:
Hello @pedrobaeza,

the commit 5704073ee4b633ebdc5ab428c8e8b742c63d8797 corrects the same problem as the one found in your pull request https://github.com/OCA/server-auth/pull/388,  however I proposed a different correction.

Can you check my pull request please ?

pedrobaeza:
If I don't remember bad, the CI is now correct in 13.0. In any case, your solution is changing the behavior, so it's not correct. Just rebase on top of the current one and remove your commit to see if it's green.

MarineGCH:
@pedrobaeza  indeed this commit is no longer necessary, the CI has been fixed in version 13.0 by this commit d7fe32d25146769ba97aaf857b3f17e41e8fb908

Warlocklogan (approved):

MarineGCH:
> @pedrobaeza indeed this commit is no longer necessary, the CI has been fixed in version 13.0 by this commit [d7fe32d](https://github.com/OCA/server-auth/commit/d7fe32d25146769ba97aaf857b3f17e41e8fb908)

@pedrobaeza It's good for you now ?

pedrobaeza:
Why do you need such weird dictionary translation mapping?

MarineGCH:
> Why do you need such weird dictionary translation mapping?

Because in the front and back-office when a password error is raised by the library, the error or suggestion message is not translatable ( see the "first problem" part of this message https://github.com/OCA/server-auth/pull/364#issue-1186718417)

MarineGCH:
> Why do you need such weird dictionary translation mapping?

For example in this line and in this all file https://github.com/dwolfhub/zxcvbn-python/blob/f596bef499c02bb6f0c8aef33b260364add2dc2b/zxcvbn/feedback.py#L48 these warnings are not translatable.

In my dictionnary, I replace english message of the zxcvbn library by english message translatable with the `_()` of Odoo then be able to translate them into any language in the odoo back-office.
This message does not have a key in the library so I rewrite the string by adding this `_()`.

In this library you can use gettext for the translation but is not implemented.
So if I use gettext I would have to create the translation file on the server and I wouldn't have access to the translation in the Odoo back office.

pedrobaeza:
I consider this very convoluted and fragile depending on the `zxcvbn` version to be worth. Anyways, it seems that `zxcvbn` includes translation mechanisms, so you should focus on making use of such translation methods.

Issue #80, opened 2019-03-26
[11] auth_session_timeout Have to clear all cookies after logout


wilcobergacker:
Im using auth_session_timeout in combination with Odoo 11 EE and we regularly have an 50_x error "Service unavailable" when we're using Chrome.

The user is logged out properly, but after this  error we have to clear all cookies before we can login  in Odoo again. 

Any idea how to fix this?

storagerepo:
Hi wilcobergacker,

I am also facing this issue,did you fix that issue?

PR #366, opened 2022-04-04
[11.0][FIX] auth_session_timeout: prevent infinite redirections

tack-yabeQuartile:
Infinite redirections loop has been reported in the auth_session_timeout module.
ex.) [issue-33](https://github.com/OCA/server-auth/issues/33), [issue-80](https://github.com/OCA/server-auth/issues/80) for V11.
and 208, 216 for V12.

Actions to reproduce the issue:
1. Set "inactive_session_time_out_delay" to any short time you like. (We used 15 and 1800)
2. Log in with your own account and leave the session to be logged out by the module.
3. Repeat logging in and being logged out. (three times we did.)
4. Change your device or browser (secret mode browser is also OK.) and log in with the same account.
5. Then you would face to the redirection loop when access to the {domain}/ or {domain}/web/login and so on.
Once redirection starts, everyone seems to face it. 
If you delete the cookie, loop would not happen temporally.

Solution:
as mentioned in issue-33, /web/login and /web/reset_password should be always ignored by checking sessions.

We checked this change is effective in V11.

### DIFF
--- a/auth_session_timeout/models/res_users.py
+++ b/auth_session_timeout/models/res_users.py
@@ -60,6 +60,9 @@ class ResUsers(models.Model):
         if not http.request:
             return
 
+        if http.request.httprequest.path in ['/web/login','/web/reset_password']:
+            return
+
         session = http.request.session
 
         # Calculate deadline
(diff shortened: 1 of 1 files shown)


tack-yabeQuartile:
adjust lint error

tack-yabeQuartile:
adjust lint error

PR #368, opened 2022-04-06
auth_oidc: allow assign groups from token claims

websavetnik:
this is Odoo v14.00 of the oauth equivalent of [users_ldap_groups](https://github.com/OCA/server-auth/tree/14.0/users_ldap_groups), allowing to also configure all authorizations in your identity provider.


sbidoul:
Hi, this looks similar to https://github.com/OCA/server-auth/pull/350. Can you please cherry-pick the commits to preserve attribution ?

PR #373, opened 2022-04-13
[FIX] auth_session_timeout: `getmtime` always return current time

### LINKED ISSUE #259: [14.0] auth_session_timeout: expiration checking
https://github.com/OCA/server-auth/blob/ceed639b21f2ce07685accb1f06de2b91d29b9be/auth_session_timeout/models/res_users.py#L73

Hello,
it is working the expiration checking? I set the system parameter to 120 seconds, and after that time the user session is not terminated. Am I missing a detail?

DonatasV:
Fixed: #259
We can't assume that this module is the only one modifying the time of session files. For some reason, on Docker instances of Odoo (that includes Odoo.SH) these session files get modified on each page reload even before `getmtime` in this module is called, which means it never expires.

This solution seems more elegant and common among the whole WWW when it comes to expiring sessions.
It also removes the need for try/exept blocks and the tests that came with them; Removes the need of os imports.

### DIFF
--- a/auth_session_timeout/models/res_users.py
+++ b/auth_session_timeout/models/res_users.py
@@ -3,8 +3,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import logging
-from os import utime
-from os.path import getmtime
 from time import time
 
 from odoo import api, http, models
@@ -67,16 +65,7 @@ class ResUsers(models.Model):
         # Check if past deadline
         expired = False
         if deadline is not False:
-            path = http.root.session_store.get_session_filename(session.sid)
-            try:
-
-                expired = getmtime(path) < deadline
-            except OSError:
-                _logger.exception(
-                    "Exception reading session file modified time.",
-                )
-                # Force expire the session. Will be resolved with new session.
-                expired = True
+            expired = session.get("last_active_time", time()) < deadline
 
         # Try to terminate the session
         terminated = False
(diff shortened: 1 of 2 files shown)


gaikaz (approved):
💯

gaikaz:
@alexeirivera87 @lonelyleaves @prakashjain 
Could you guys please try this out, maybe give your review?

prakashjain:
@gaikaz Thanks for the update. I tried updated fix but still same problem. session.get("last_active_time", time()) always showing current time. Inactive user timeout functionality not working configured 2 or 5 minutes, but timeout not working

gaikaz:
@prakashjain
That's strange. Maybe you have a different issue after all... You see, even before this PR the module works, but inconsistently and only on Docker or Odoo.SH instances. 
As far as we saw, it works ok on on-premise Ubuntu 18.04 servers. 

Does your Odoo run on Docker or SH?

prakashjain:
@gaikaz  Odoo 14 community edition in my local system Ubuntu OS checked. Docker not installed.

gaikaz:
@prakashjain It must be a different issue for you then as it should work in your case IMO.
Maybe you have some other modules (multiple-databases?) that "touch" the session files more often.

prakashjain:
@gaikaz After adding  inactive_session_time_out_ignored_url   Value /longpolling/im_status working.
        
Updated value:-

        <field name="key">inactive_session_time_out_ignored_url</field>
        <field name="value">/calendar/notify,/longpolling/poll,/longpolling/im_status</field>

I have one query in my custom module  inactive user session log out some users restrict this functionality. For Example, some user login and access data through API postman. How to restrict this functionality while user login or access screen through API user?. 
Thanks

gaikaz:
@prakashjain 
Sorry for not mentioning this sooner, but yeah, we found the same thing last week, that adding `/longpolling/im_status` is also needed for this. I'm glad you figured it out and it works 👍🏻 

About the API stuff, I suppose you could also add the `/xmlrpc/2/common` and `/xmlrpc/2/object` endpoints to ignored url list, so it doesn't extend the user's session time. I'm not sure what your use case is, tho, I don't think it's a good practice in general to use the same user for API and normal GUI use.

chrisb-c01:
Would it be possible to get this PR merged as the current solution in 14.0 is not working?

PR #374, opened 2022-04-14
[FIX][14.0][vault] Lock out of users when login changed

fkantelberg:
During the work on disaster recovery I discovered a bug regarding the decryption of the private key. Previously the password for encrypting/decrypting the private key was formed from the login of the user and the given password and/or passfile. If the user logjn is changed the first part and therefore the password will always be wrong locking out the user from his private key and all vaults effectivly.

The fix is taking care of it by introducing a versioning to the res.users.key to indicate if a migration of the key is required. The javascript should migrate the private keys with the next usage automatically. The private key will be re-encrypting without the login and stored in the database as a new res.users.key with version 1.

It will only cause a problem if the login already changed. In that case it's recommended to change the login back shortly. Alternativly the res.users.key could be flagged as version 1 and the user should input `<username>|<password>`.

### DIFF
--- a/vault/__manifest__.py
+++ b/vault/__manifest__.py
@@ -10,7 +10,7 @@
     "application": True,
     "author": "initOS GmbH, Odoo Community Association (OCA)",
     "category": "Vault",
-    "depends": ["web"],
+    "depends": ["base_setup", "web"],
     "data": [
         "security/ir.model.access.csv",
         "security/ir_rule.xml",
--- a/vault/models/res_users.py
+++ b/vault/models/res_users.py
@@ -61,4 +61,5 @@ class ResUsers(models.Model):
             "public": self.active_key.public,
             "salt": self.active_key.salt,
             "uuid": self.active_key.uuid,
+            "version": self.active_key.version,
         }
--- a/vault/models/res_users_key.py
+++ b/vault/models/res_users_key.py
@@ -26,6 +26,7 @@ class ResUsersKey(models.Model):
     salt = fields.Char(required=True, readonly=True)
     iv = fields.Char(required=True, readonly=True)
     iterations = fields.Integer(required=True, readonly=True)
+    version = fields.Integer(readonly=True)
     # Encrypted with master password of user
     private = fields.Char(required=True, readonly=True)
 
@@ -38,7 +39,7 @@ class ResUsersKey(models.Model):
             else:
                 rec.fingerprint = False
 
-    def _prepare_values(self, iterations, iv, private, public, salt):
+    def _prepare_values(self, iterations, iv, private, public, salt, version):
         return {
             "iterations": iterations,
             "iv": iv,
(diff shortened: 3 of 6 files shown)


fkantelberg:
@AntoniRomera Could you also please review here?

AntoniRomera (approved):

fkantelberg:
@rousseldenis @CarlosRoca13 Could you also please review this PR?

CarlosRoca13:
When I try to create a new private key is doing nothing:

fkantelberg:
@CarlosRoca13 Thank you for testing.

I suspect the missing secure context behind the doing nothing. This modules requires a HTTPS connector or to run on localhost. Both are required for the secure context of the browsers which functions are used.

CarlosRoca13:
When I'm trying to install it locally I'm getting an error:

`ValueError: External ID not found in the system: base_setup.res_config_settings_view_form`

fkantelberg:
@CarlosRoca13 I totally forgot the dependency. I pushed the fix to the branch here. Thank you

vault/__manifest__.py
```diff
@@ -5,12 +5,12 @@
     "name": "Vault",
     "summary": "Password vault integration in Odoo",
     "license": "AGPL-3",
-    "version": "14.0.1.6.0",
+    "version": "14.0.1.6.1",
```
CarlosRoca13: This change will be done by the bot, please restore to the previous version

CarlosRoca13 (commented):
Just a comment

It is working well

vault/__manifest__.py
fkantelberg: Done

pedrobaeza:
Please squash commit history according the business logic separation you want to do

CarlosRoca13 (approved):

fkantelberg:
@pedrobaeza I squashed everything

pedrobaeza:
What is going to happen with existing module users? Is this compatible?

fkantelberg:
@pedrobaeza 
Most of the stuff I added was to ensure it's backward compatible. I added a version field to the model storing the keys in the database and a on-the-fly migration in the javascript. Because it requires a decrypted private key it's only possible when a user decrypts it. In case the private key is loaded with the old version it's re-encrypted with the new material and will create a new entry in the database. This is done via the javascript `_check_key_migration`.

New keys will always be created with the current version 1. If we need to change something later again we can do the migration the same way. This makes it basically backwards compatible without breaking.

You are right asking because it's crucial. This was also the reason I created and added the link to vault-recovery to support a disaster recovery from backups more easily.

pedrobaeza:
Great, thanks for the detailed explanation.

There's an error in the CI that I think is not related to this PR. I'll switch as part of my community involvement to GH actions and see if the CI becomes green.

fkantelberg:
@pedrobaeza The error is in the module `auth_signup_verify_email` in one of the test cases. Quite strange but I see no relation here. Do you need anything else from me here?

PR #382, opened 2022-05-20
[14.0][FIX] oauth_oidc token mapping after upstream change

theangryangel:
Upstream now starts injecting values that don't exist in the OAuth response (see: https://github.com/odoo/odoo/commit/fb3c4845b1549bc2e1378620a5f01e52aa4dbbdb#diff-9928f68e69da2abd2aa17e05de18875e0fea83edc2d55e942c9bbb65776f6937R53-R66), based on guess work.

This interferes with existing installations where a token mapping was being used to provide a user_id mapping.

Unfortunately in the case of Azure Active Directory the user_id upstream now copies across is difficult to find to provision users with since it has no bearing on things like the UPN, etc. that would be considered "more normal" in the AD world (I'm sure it can be found, but it seems difficult).

More over it breaks existing installations where a token mapping was providing a different value.

--

I'm not clear if this is the best way to handle this situation.

If anyone has any better ideas, I'll happily take them for feedback, I just needed a quick fix this morning to allow us into our own Odoo installation.

### DIFF
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -60,8 +60,7 @@ class AuthOauthProvider(models.Model):
         if self.token_map:
             for pair in self.token_map.split(" "):
                 from_key, to_key = [k.strip() for k in pair.split(":", 1)]
-                if to_key not in res:
-                    res[to_key] = res.get(from_key, "")
+                res[to_key] = res.get(from_key, "")
         return res
 
     def _parse_id_token(self, id_token, access_token):


sbidoul:
Yeah, getting the latest Odoo commits is always a risky proposition.

At first glance I'd say your change is fine. It might have some backward compatibility implications, I'm not sure.

Is it correct that that upstream commit forces you to update the token mapping to produce `sub` instead of `user_id` ?

theangryangel:
For now user_id does still work, but the comment does indicate that’ll be removed.

The only breaking change I can think of it is if someone’s setup a mapping which sometimes needs to remap a field and sometimes not. I don’t think I’ve ever seen a scenario that would necessitate that…

I could add some syntax to indicate allowing overrides. That doesnt seem very discoverable though.. Or turning it into a fully fledged model (and adding a migration) so that additional options can be specified, but that seemed a bit like overkill :/

I’m happy to do the leg work, if you’ve got any guidance

theangryangel:
I’ve missed something with my
patch it seems as we’ve had a reoccurrence this morning. For now I’ve rolled back and I’ll take another stab at figuring out what I’ve missed. Apologies for the noise 😕

PR #384, opened 2022-05-23
[13.0][MIG] vault, vault_share

rousseldenis:


vault/views/assets.xml
```diff
@@ -0,0 +1,38 @@
...
+                href="/vault/static/src/scss/vault.scss"
+            />
+        </xpath>
+    </template>
+
+    <template id="tests_assets" inherit_id="web.tests_assets">
+        <xpath expr="." position="inside">
+            <script type="text/javascript" src="/vault/static/tests/vault_tests.js" />
+        </xpath>
+    </template>
```
fkantelberg, line 31: Isn't supported in 13.0 and was commented out before.

vault/views/vault_entry_views.xml
```diff
@@ -0,0 +1,172 @@
...
+                />
+                <filter
+                    string="Vault"
+                    name="vault"
+                    domain="[]"
+                    context="{'group_by': 'vault_id'}"
+                />
+                <searchpanel>
+                    <field name="parent_id" string="Entries" enable_counters="1" />
+                </searchpanel>
```
fkantelberg, line 168: searchpanel is also only 14.0+

fkantelberg (changes requested):
I tested the main functions and all seems to work good when you apply the 2 requested changes. Otherwise there is only the possibility that a user can get locked out of the vault when the login is changed.

vault/views/vault_entry_views.xml
rousseldenis: Nope, I found them in 13, but attribute enable_counters not.

CarlosRoca13 (commented):
I have tested the migration locally and everything seems to be working perfectly except that users without sharing permissions can perform this action. Can you check it?

fkantelberg:
@CarlosRoca13 
What do you mean with action `share`? The share permission is allowing users to add additional users to the vault (to share the vault with a different user). It isn't blocking a user from sharing secrets using the inbox or the external shares.

CarlosRoca13 (approved):
Ohh!! Ok, so it's working as expected.

Thanks for the clarification

fkantelberg (approved):

CarlosRoca13:
So please rebase and run pre-commit :sweat_smile:

pedrobaeza:
@rousseldenis can you fix pre-commit? Or do we supersede it respecting your authorship?

rousseldenis:
> @rousseldenis can you fix pre-commit? Or do we supersede it respecting your authorship?

Yes I will do it.

Was on dad-holidays :sweat_smile:

pedrobaeza:
Hey, congratulations! We have superseded meanwhile the PR in #386. Do you mind? Or do you prefer to continue in this one?

rousseldenis:
Nope, that's great! Thanks

PR #388, opened 2022-06-12
[14.0][FIX] auth_signup_verify_email: Use real email address for tests 

pedrobaeza:
In one part of the CI pipeline, the IANA example domains are not accepted, getting this message from `email_validator` library:

The domain name example.com does not accept email

So we switch to a real domain for avoiding the problem, using our famous main OCA list address, which at the end, it's not secret and won't be disabled in a short future.

@Tecnativa

### DIFF
--- a/auth_signup_verify_email/tests/test_verify_email.py
+++ b/auth_signup_verify_email/tests/test_verify_email.py
@@ -44,6 +44,6 @@ class UICase(HttpCase):
     @mute_logger("odoo.addons.auth_signup_verify_email.controllers.main")
     def test_good_email(self):
         """Test acceptance of good emails."""
-        self.data["login"] = "good@example.com"
+        self.data["login"] = "contributors@odoo-community.org"
         doc = self.html_doc(data=self.data)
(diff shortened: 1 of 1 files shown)


rvalyi:
@pedrobaeza the requirements.txt is now generated based on the modules `__manifest__.py` and I have seen no option to specify a package version in external_dependencies so I wonder what would be the best way to work around such issues.  cc @sbidoul

pedrobaeza:
I'm just trying some things. Pinning doesn't serve for this case, but if it would serve, the option to "consolidate" it is to put that version in the manifest in the `external_dependencies` part.

sbidoul:
> I have seen no option to specify a package version in external_dependencies so I wonder what would be the best way to work around such issues

I'm not sure what the problem is here, but to answer this specific question, I'm pretty sure you can put version specifiers in external_dependencies in Odoo >= 13.

pedrobaeza:
At the end, I have needed to switch to a real email address for passing the tests, as example domain is no longer valid in the CI environment (it was passing in local environment though), but it's not a big deal.

/ocabot merge patch

PR #396, opened 2022-07-22
[14.0][FIX] auth_api_key: fix archived user allowed key

TDu:
If the user is archived his api key should not be valid anymore.

Ports:
- https://github.com/OCA/server-auth/pull/398
- https://github.com/OCA/server-auth/pull/397

### DIFF
--- a/auth_api_key/tests/test_auth_api_key.py
+++ b/auth_api_key/tests/test_auth_api_key.py
@@ -43,3 +43,9 @@ class TestAuthApiKey(SavepointCase):
         )
         with self.assertRaises(ValidationError):
             self.env["auth.api.key"]._retrieve_uid_from_api_key("api_key")
+
+    def test_user_archived(self):
+        demo_user = self.env.ref("base.user_demo")
+        demo_user.active = False
+        with self.assertRaises(ValidationError), self.env.cr.savepoint():
+            self.env["auth.api.key"]._retrieve_uid_from_api_key("api_key"), demo_user.id
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -23,6 +23,7 @@ class AuthApiKey(models.Model):
         help="""The user used to process the requests authenticated by
         the api key""",
     )
+    active = fields.Boolean(related="user_id.active")
 
     _sql_constraints = [("name_uniq", "unique(name)", "Api Key name must be unique.")]
 


simahawk (approved):

simahawk:
can you please fwd to 15.0 and backport to 13.0?

TDu:
> can you please fwd to 15.0 and backport to 13.0?

Done

sebalix (approved):

auth_api_key/tests/test_auth_api_key.py
```diff
@@ -43,3 +43,9 @@ def test_cache_invalidation(self):
         )
         with self.assertRaises(ValidationError):
             self.env["auth.api.key"]._retrieve_uid_from_api_key("api_key")
+
+    def test_user_archived(self):
+        demo_user = self.env.ref("base.user_demo")
+        demo_user.active = False
+        with self.assertRaises(ValidationError), self.env.cr.savepoint():
```
simahawk, line 50: @TDu question: why do you need a savepoint here?

TDu: I just used the same code that is used in the previous check above.

simahawk: But the test before works w/out :thinking: 
I might be wrong but I'm pretty sure is useless in both cases in this ctx. 
Anyway, not a blocker.

sebastienbeau:
@lmignon @simahawk @hparfr 
In case that you have a Shopinvader Website with no specific user this will break your installation (as system user is inactive)

lmignon:
@TDu This change will break our installations... Can you revert-it? It's a common use case (and the initial one) to use a system user to run services protected by an api key.

lmignon:
ping @simahawk @sebalix (see my previous comment)

simahawk:
I agree on reverting it to not break existing installation but I don't like the assumption that you can have an inactive user which is able to use an api key.

I guess that this is because you want to prevent logins -> this should be done explicitly IMO.

lmignon:
We want to prevent login but also the use of a pay user for running services.

lmignon:
@simahawk We could maybe declare the active field as computed with readonly=False. To no break the current implementation, you could use a pre-migrate script to add the column 'active' with the value set to True? Not tested, therefore I don't know if the update of the dependent field will update the computed field in the case where the field has not been manually set.

TDu:
This is the revert PR https://github.com/OCA/server-auth/pull/400 for this version

TDu:
I also created a revert pr for 13.0 #402 and 15.0 #401

PR #400, opened 2022-07-26
[14.0] Revert auth_api_key: fix api key archived user

TDu:


lmignon (changes requested):
@TDu You can't revert the bump version... The wheel is already published on [pypi](https://pypi.org/project/odoo14-addon-auth-api-key/)

TDu:
> @TDu You can't revert the bump version... The wheel is already published on [pypi](https://pypi.org/project/odoo14-addon-auth-api-key/)

Make sense, I kept the revert on the translation file, tho.

sebalix (approved):

sebalix:
@lmignon is it OK for you?

lmignon (approved):

simahawk:
/ocabot merge patch

@TDu could you propose another solution?
What about having a config param to enable this policy and explicitly check for user.active on lookup?

PR #403, opened 2022-07-28
[15.0][MIG] auth_brute_force.

nikunjantala:


Rad0van:
You did not follow the migration steps: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0 History is completely missing.

Rad0van:
Also the tests in fact do not run and all sort of fail (without failing) on `Skip AttributeError` You need to adjust them to v15.0

Rad0van:
Better, but still not there yet. The correct migration should ending up with 2 commits (result of the mentioned procedure) like this:
```
[IMP] auth_brute_force: black, isort, prettier
[MIG] auth_brute_force: Migration to 15.0
```
Please follow the procedure.

CRogos:
To fix the build/tests you need to rebase.

Rad0van:
@nikunjantala watch this in 7 minutes: https://odoo-community.org/event/oca-days-2022-liege-2022-10-10-2022-10-11-131/track/how-to-contribute-to-oca-doing-reviews-and-submitting-your-first-module-144 I guess it will be helpful. OCA rulez!

nileshsheliya:
Yes, thank you.

dsolanki-initos:
Hello @yajo  having an issue in running tests, So can you please help me to figure out why I'm facing this issue?
here is the PR ref: https://github.com/OCA/server-auth/actions/runs/3580460398/jobs/6022607328

Also, I have attached the image for the same.
auth_brute_force

Thank you

auth_brute_force/models/res_users.py
```diff
@@ -0,0 +1,152 @@
...
+    #         return original_fn(environ, start_response)
+    #
+    #     wsgi_server.application_unproxied = _patch
+
+    # Helpers to track authentication attempts
+    @classmethod
+    @contextmanager
+    def _auth_attempt(cls, login):
+        """Start an authentication attempt and track its state."""
+        cls.environ = request.httprequest.environ
```
yajo: You're using the `request` object from within a model.

When you're writing a controller, you can expect that there'll always be a bound request. But models can be called without a request. For example, during normal unit tests written within a `TransactionCase` subclass. Since there's no request, the test fails.

Since it makes no sense to prevent brute-force attacks when there's no request, you should capture that unbound request error and just pass in case it's produced.

alihashaam: @yajo 

Thank you for your pointer. We have fixed this in our recent push(es)

PR #405, opened 2022-08-03
[MIG] [12.0] oauth_provider

hbrunn:
this is the v12 version of https://github.com/OCA/server-tools/tree/9.0/oauth_provider

Note I added a few changes on top of the migration:

- previously, the permissions on clients and some other models were too lax. in v12, only managers have access to clients and scopes, everything else goes via sudo(). That's why a bunch of methods are private now
- implemented client type `backend application`
- implemented http auth method `oauth_provider`

Those involved some necessary refactoring, that's why I rolled that together.


hbrunn:
tests are fixed in https://github.com/OCA/server-auth/pull/406

antonioburic (approved):

astirpe (approved):

victor-champonnois:
@hbrunn Can this be merged ? I'm going to port the code to 14.0 based on this PR

hbrunn:
@victor-champonnois there's no need to wait until this is merged, you can just take the commits from here. Actually, on a higher version the code has a better chance to get merged as probably more people are interested

hbrunn:
CI is fixed in https://github.com/OCA/server-auth/pull/465

oauth_provider/models/oauth_provider_token.py
```diff
@@ -0,0 +1,98 @@
...
+        ('token_unique', 'UNIQUE (token, client_id)',
+         'The token must be unique per client !'),
+        ('refresh_token_unique', 'UNIQUE (refresh_token, client_id)',
+         'The refresh token must be unique per client !'),
+    ]
+
+    @api.multi
+    def _compute_active(self):
+        for token in self:
+            token.active = fields.Datetime.now() < token.expires_at
```
remytms, line 44: Shouldn't this method be prefixed by `@api.depends("expires_at")` as the method use this field ?

hbrunn: we need the dependency when we store the field, but we don't

remytms:
Module in 16.0 is [here](https://github.com/OCA/server-auth/pull/795).

PR #406, opened 2022-08-03
[FIX][12.0] auth_signup_verify_email: Turn off deliverability checks when testing

hbrunn:
since version 1.2.0, the underlying library checks for mx records, which we don't want during tests


pedrobaeza:
This solution is simpler: https://github.com/OCA/server-auth/commit/d7fe32d25146769ba97aaf857b3f17e41e8fb908

antonioburic (approved):

hbrunn:
thanks. I personally think unittests should never have dependencies to anything external like in this case some dns record existing, but as that's the accepted solution and it's an old version, I'm happy to just cherry-pick this

pedrobaeza:
I don't agree with that, as tests should mimic reality, not to mock/fake it, so thanks for the cherry-picking.

/ocabot merge patch

PR #410, opened 2022-08-31
[13.0][FIX] vault: complete_name not computed when created from other entry

CarlosRoca13:
Before the path if we create a new child entry from other entry, the
complte_name of the entry is not changing when we save the new entry.
vault_01

After doing this changes, we are getting the complete_name correctly.
vault_02

cc @Tecnativa TT38729

ping @victoralmau  @pedrobaeza

### DIFF
--- a/vault/views/vault_entry_views.xml
+++ b/vault/views/vault_entry_views.xml
@@ -53,6 +53,7 @@
                                 name="parent_id"
                                 options="{'no_open': true, 'no_create_edit': true, 'no_quick_create': true}"
                             />
+                            <field name="complete_name" invisible="1" />
                             <field name="name" />
                             <field name="url" widget="url" />
                             <field name="tags" widget="many2many_tags" />


CarlosRoca13:
Set as draft to correct another issue

victoralmau:
A change has been added for correct display (not showing unnecessary records).

**Before**
vault-antes

After
vault-despues

victoralmau (approved):

vault/models/vault_entry.py
```diff
@@ -1,17 +1,21 @@
...
+from odoo import _, api, fields
 from odoo.exceptions import ValidationError
 
+from odoo.addons.web.models import models
+
 _logger = logging.getLogger(__name__)
 
 
-class VaultEntry(models.Model):
+class VaultEntry(odoo.models.Model):
```
pedrobaeza: Why changing this?

pedrobaeza:
I think this PR should be limited to the first patch and let the other in another PR.

victoralmau:
> I think this PR should be limited to the first patch and let the other in another PR.

https://github.com/OCA/server-auth/pull/411

CarlosRoca13:
So ready 4 review

pedrobaeza (approved):

pedrobaeza:
Please fw-port it to v14

PR #411, opened 2022-08-31
[13.0][FIX] vault: Prevent 200 records limit in search panels.

victoralmau:
Changes done:
- [x] Prevent 200 records limit in search panels.
- [x] Show in the left sidebar only the records that have children.

This is quite similar to https://github.com/OCA/dms/pull/167

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT38731

### DIFF
--- a/vault/models/vault_entry.py
+++ b/vault/models/vault_entry.py
@@ -1,4 +1,5 @@
 # © 2021 Florian Kantelberg - initOS GmbH
+# Copyright 2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import logging
@@ -71,10 +72,28 @@ class VaultEntry(models.Model):
 
     @api.model
     def search_panel_select_range(self, field_name):
-        """Add context to show just the name in searchpanel instead of full path"""
+        """We add the following contexts related to searchpanel:
+        - entry_short_name: Show just the name instead of full path.
+        - search_panel_limit: It will be a small hack to NOT limit records.
+        """
         return super(
-            VaultEntry, self.with_context(entry_short_name=True)
+            VaultEntry,
+            self.with_context(search_panel_limit=False, entry_short_name=True),
         ).search_panel_select_range(field_name)
+
+    def search_read(self, domain=None, fields=None, offset=0, limit=None, order=None):
+        """Changes related to searchpanel:
+        - Remove the limit of records (default is 200). In v14 it is not necessary.
+        - Add a domain to only show records with children.
+        """
+        limit = self.env.context.get("search_panel_limit", limit)
+        domain = domain if domain else []
+        if "search_panel_limit" in self.env.context:
+            domain += [("child_ids", "!=", False)]
+        res = super().search_read(
+            domain=domain, fields=fields, offset=offset, limit=limit, order=order
+        )
+        return res
 
     @api.depends("name", "complete_name")
     def _compute_display_name(self):
--- a/vault/tests/test_vault.py
+++ b/vault/tests/test_vault.py
@@ -1,4 +1,5 @@
 # © 2021 Florian Kantelberg - initOS GmbH
+# Copyright 2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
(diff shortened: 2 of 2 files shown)


vault/models/vault_entry.py
```diff
@@ -1,17 +1,21 @@
...
+from odoo import _, api, fields
 from odoo.exceptions import ValidationError
 
+from odoo.addons.web.models import models
+
 _logger = logging.getLogger(__name__)
 
 
-class VaultEntry(models.Model):
+class VaultEntry(odoo.models.Model):
```
pedrobaeza: I insist: why this change?

pedrobaeza:
Not sure we should do this feature change. @fkantelberg what do yo think?

vault/models/vault_entry.py
CarlosRoca13: It's because he is adding this models: https://github.com/OCA/server-auth/pull/411/files#diff-40d91c6c9961012399719d24970edab4a33704ffb9d72e74ba0528c9b966f378R13

pedrobaeza: Then better to do `from odoo.addons.web.models import models as web_models`

vault/tests/test_vault.py
```diff
@@ -155,3 +159,13 @@ def test_search_expired(self):
         self.assertTrue(domain[0] == "|")
         self.assertIn(("expire_date", "=", False), domain)
         self.assertTrue(any(("expire_date", ">=") == d[:2] for d in domain))
+
+    def test_vaul_entry_search_panel_limit(self):
```
fkantelberg: Typo in `vault`

vault/models/vault_entry.py
```diff
@@ -71,10 +75,25 @@ def _compute_complete_name(self):
...
+        """Add context to show just the name in searchpanel instead of full path.
+        We need to override function to prevent 200 records limit in search panel."""
+        old_value = models.SEARCH_PANEL_LIMIT
+        models.SEARCH_PANEL_LIMIT = False
+        _self = self.with_context(
+            search_panel_domain_override=True, entry_short_name=True
+        )
+        res = super(VaultEntry, _self).search_panel_select_range(field_name)
+        models.SEARCH_PANEL_LIMIT = old_value
+        return res
```
fkantelberg: I get the meaning behind it but the fact that you already did that here in multiple locations sounds like a required core patch instead of a hack like this in every location.

I would rather suggest to patch odoo.addons.web.models.models like this:
```
search_panel_limit = self.env.context.get("search_panel_limit", SEARCH_PANEL_LIMIT)
return {
    'parent_field': parent_name,
    'values': Comodel.with_context(hierarchical_naming=False).search_read([], fields, limit=search_panel_limit),
}
```
and set the context here.

vault/models/vault_entry.py
```diff
@@ -71,10 +75,25 @@ def _compute_complete_name(self):
...
+
+    def search_read(self, domain=None, fields=None, offset=0, limit=None, order=None):
+        """We apply a domain to not list records that do not have children."""
+        if self.env.context.get("search_panel_domain_override"):
+            domain = [("child_ids", "!=", False)]
```
fkantelberg: From a navigational perspective this makes sense to not show the leaves of the tree. I'm not quite sure if you want to fully overwrite it or extend the domain here. It could make a difference if your view/filter is set.

fkantelberg (changes requested):
Thank you for your suggestions. See my comments. For large databases a change like this makes absolutely sense.

vault/models/vault_entry.py
victoralmau: Finally, this is not necessary.

vault/models/vault_entry.py
victoralmau: Finally, this is not necessary.

I have changed the approach in another way to simplify code.

victoralmau:
Changes done.

fkantelberg (approved):
Looks less hacky now :+1:

pedrobaeza:
Please explain a bit in the commit message and code comment what leads to that implementation.

victoralmau:
Changes done.

pedrobaeza (approved):

PR #386, opened 2022-06-06
[13.0][MIG] vault and vault_share: Migration to 13.0

CarlosRoca13:
cc @Tecnativa TT37022

superseded from https://github.com/OCA/server-auth/pull/384

ping @fkantelberg @pedrobaeza

PR #413, opened 2022-09-01
[14.0][IMP+FIX] vault: Show in the left sidebar only the records that have children.

victoralmau:
Changes done:
- [x] [IMP] Show just name instad of full path in searchpanel (Forgotten in https://github.com/OCA/server-auth/pull/386)
- [x] [IMP] Show in the left sidebar only the records that have children. (FWP from 13.0: https://github.com/OCA/server-auth/pull/411)
- [x] [IMP] Set limit to `False` in searchpanel (FWP from 13.0: https://github.com/OCA/server-auth/pull/411)
- [x] [IMP] Add Vaults to searchpanel view. 

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT38731

### DIFF
--- a/vault/models/vault_entry.py
+++ b/vault/models/vault_entry.py
@@ -73,6 +73,20 @@ class VaultEntry(models.Model):
             else:
                 rec.complete_name = rec.name
 
+    @api.model
+    def search_panel_select_range(self, field_name):
+        """Add context to show just the name in searchpanel instead of full path"""
+        return super(
+            VaultEntry, self.with_context(entry_short_name=True)
+        ).search_panel_select_range(field_name)
+
+    @api.depends("name", "complete_name")
+    def _compute_display_name(self):
+        if not self.env.context.get("entry_short_name", False):
+            return super()._compute_display_name()
+        for record in self:
+            record.display_name = record.name
+
     @api.depends("expire_date")
     def _compute_expired(self):
         now = datetime.now()
--- a/vault/models/vault_entry.py
+++ b/vault/models/vault_entry.py
@@ -1,4 +1,5 @@
 # © 2021 Florian Kantelberg - initOS GmbH
+# Copyright 2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import logging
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

CarlosRoca13 (approved):

vault/models/vault_entry.py
```diff
@@ -73,6 +74,36 @@ def _compute_complete_name(self):
...
+    @api.model
+    def search_panel_select_range(self, field_name, **kwargs):
+        """We add the following contexts related to searchpanel:
+        - entry_short_name: Show just the name instead of full path.
+        - from_search_panel: It will be used to overwrite domain.
+        """
+        return super(
+            VaultEntry,
+            self.with_context(from_search_panel=True, entry_short_name=True),
+        ).search_panel_select_range(field_name, **kwargs)
```
fkantelberg, line 88: In the kwargs is a limit parameter with v14+. You might want to set it to None here otherwise the default 200 comes from the JS.

fkantelberg (changes requested):

vault/models/vault_entry.py
victoralmau: Since v14 limit is None by default: https://github.com/odoo/odoo/blob/14.0/addons/web/models/models.py#L490

fkantelberg: too-many

Doesn't fit my tests. You are right limit is None in python but it's getting called from JS and the DEFAULT_LIMIT is set there https://github.com/odoo/odoo/blob/14.0/addons/web/static/src/js/views/search_panel_model_extension.js#L13

victoralmau: Oops, I didn't see that odoo had moved the limit to js. 
Thanks, it's changed.

fkantelberg:
@victoralmau Thank you.

I also just noticed it now that the default happens before the eval in [1]. Which means setting `limit="0"` as attribute in the searchpanel fields is enough. Should be good enough here. I pull it into the 15.0 branch and try to finish my testing there asap.

[1] https://github.com/odoo/odoo/blob/14.0/addons/web/static/src/js/views/search_panel_model_extension.js#L432

PR #416, opened 2022-09-04
[14.0][FIX] vault: reencryption

fkantelberg:
If an user gets deleted from a vault the entire data has to be re-encrypted to provide the best security. The current implementations was creating a new master key and setting it to the first 40 people of the vault while not changing the fields/files because these weren't loaded because of invisible.

The changes will load all the rights, fields, and files of a load separatly ignoring the view definition to ensure that everything will be done correctly.

### DIFF
--- a/vault/models/vault_file.py
+++ b/vault/models/vault_file.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Same as v13

pedrobaeza:
And fast-tracking as well:

/ocabot merge patch

PR #417, opened 2022-09-04
[13.0][FIX] reencryption

fkantelberg:
If an user gets deleted from a vault the entire data has to be re-encrypted to provide the best security. The current implementations was creating a new master key and setting it to the first 40 people of the vault while not changing the fields/files because these weren't loaded because of invisible.

The changes will load all the rights, fields, and files of a load separatly ignoring the view definition to ensure that everything will be done correctly.

### DIFF
--- a/vault/models/vault_file.py
+++ b/vault/models/vault_file.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Hello, @CarlosRoca13 is on holidays 2 weeks, but we have deployed the patch for testing on our customer to see if it's OK (on runbot can't be tested due to the HTTP thing).

Please rebase the branch over latest 13.0 for removing the README update commit.

fkantelberg:
@pedrobaeza Thanks for the notification. I rebased it. Last time I checked you can just change http->https in the url of the runboat and accept the certificate and it's working.

pedrobaeza (approved):
Indeed the trick of changing `http` by `https` (although it says the certificate is not valid, and we have to accept to continue), serves for trying the module in runboat. Maybe we should put such text in the README.

Tested and working correctly.

pedrobaeza:
Fast-tracking for solving the problem:

/ocabot merge patch

PR #420, opened 2022-09-08
[13.0][IMP][vault] Invalidate keys

fkantelberg:
This allows an user to invalidate the current key and removing all access. This should only be done if the password is lost because all access to vaults are lost.

After hitting the recreation of the inbox link stay in the view.

### DIFF
--- a/vault/controllers/main.py
+++ b/vault/controllers/main.py
@@ -125,4 +125,4 @@ class Controller(http.Controller):
             master_key = keys.get(right.vault_id.uuid)
 
             if isinstance(master_key, str):
-                right.key = master_key
+                right.sudo().key = master_key
--- a/vault/controllers/main.py
+++ b/vault/controllers/main.py
@@ -83,6 +83,22 @@ class Controller(http.Controller):
 
         return {"public_key": user.active_key.public}
 
+    @http.route("/vault/inbox/get", auth="user", type="json")
+    def vault_get_inbox(self):
+        inboxes = request.env.user.inbox_ids
+        return {inbox.token: inbox.key for inbox in inboxes}
+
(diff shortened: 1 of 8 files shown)


CarlosRoca13:
Doing some tests, I tried to invalidate the current private key on demo user and I get next error: 
imagen

The problem comes when the user has no more permissions than read.

fkantelberg:
@CarlosRoca13 I added a small exception here and in 14.0 which allows an owner of the vault.right to always write the key field. Tested it locally and works there. Can you test it again please?

CarlosRoca13:
I'm having same exception...

fkantelberg:
I ran it in the runboat instance (admin has admin and demo has demo as password for the key pair) and it was working. I also added unittests which seem to run flawless too. Can you see anything more than the error message?

CarlosRoca13:
Are you trying to do the operation without checking any kind of permission?

fkantelberg:
@CarlosRoca13 What are you referring to? Do you mean because of the extra routes?

CarlosRoca13:
See next GIF:
Peek 24-10-2022 09-30

fkantelberg:
Ah I see the problem. I forgot to push the sudo in the 13.0 branch in the controller.

CarlosRoca13:
I am getting same result :cry:

fkantelberg:
Ok I will check on a local v13. Maybe something with the backport went wrong. Takes a bit of time

fkantelberg:
@CarlosRoca13 Found the issue. On invalidation the vault.rights of the users are deleted and this caused the exception. I put a sudo in between for current branches. Thank you

CarlosRoca13 (approved):
Code review and tested :+1: 

Thanks!

PR #421, opened 2022-09-08
[14.0][IMP][vault] Invalidate keys

fkantelberg:
This allows an user to invalidate the current key and removing all access. This should only be done if the password is lost because all access to vaults are lost.

After hitting the recreation of the inbox link stay in the view.


CarlosRoca13 (approved):
Tested, Thanks :smile:

pedrobaeza:
Can you please squash?

fkantelberg:
Should be one commit now.

PR #425, opened 2022-09-28
[FIX] password_security: Non-admin users can keep the same password

Urvisha-OSI:


max3903:
CLA is fixed.

max3903:
@luchorck92 @jantoniorivera97

max3903 (approved):

Issue #424, opened 2022-09-26
[15.0] external dependency "cryptography" causes error 'X509_V_FLAG_CB_ISSUER_CHECK'


CRogos:
## Module auth_jwt
The name of the module that has a bug.
auth_jwt has an external dependency "cryptography" which seems not be compatible with Odoo anymore:
https://github.com/odoo/odoo/commit/55d2da4dd28975e78ace2a6beeae4168d5692f09

## Describe the bug
The Odoo server is not starting because of dependency incompatibilities. 
similar topics:
https://github.com/odoo/odoo/pull/89798
https://www.odoo.com/de_DE/forum/hilfe-1/attributeerror-module-lib-has-no-attribute-x509-v-flag-cb-issuer-check-when-creating-new-staging-branch-202955

## To Reproduce
**Affected versions**: 15.0
Odoo does not start when this repository is added to the addon folder.
Error: 
  File "/usr/lib/python3/dist-packages/OpenSSL/__init__.py", line 8, in <module>
    from OpenSSL import crypto, SSL
  File "/usr/lib/python3/dist-packages/OpenSSL/crypto.py", line 1553, in <module>
    class X509StoreFlags(object):
  File "/usr/lib/python3/dist-packages/OpenSSL/crypto.py", line 1573, in X509StoreFlags
    CB_ISSUER_CHECK = _lib.X509_V_FLAG_CB_ISSUER_CHECK
AttributeError: module 'lib' has no attribute 'X509_V_FLAG_CB_ISSUER_CHECK'

1. I am using Odoo.sh with an empty database referencing this repository. The server with an empty db does not start.

**Expected behavior**

**Additional context**
I've added the following line to the requirements.txt and the error is gone:
cryptography==36.0.2 # last version working with pyopenssl==19.0.0

But I am not using the auth_jwt module, so I am not sure if the module is still working with the changed requirements.txt?
Is this something this repository should react on, or will this probably be fixed in odoo?

sbidoul:
This has been fixed in Odoo: https://github.com/odoo/odoo/pull/99829

CRogos:
> This has been fixed in Odoo: [odoo/odoo#99829](https://github.com/odoo/odoo/pull/99829)

@sbidoul are you sure? I still get the same error when not setting the cryptography version in the requirements.txt.

sbidoul:
I don't know which version your odoo sh instance has, but I'm 100% sure we should not pin the version in this repo.

[...]

PR #427, opened 2022-10-05
[15.0][FIX] cryptography dependencies

CRogos:
Addresses issue: 
https://github.com/OCA/server-auth/issues/424
https://github.com/odoo/odoo/pull/99829


CRogos:
@sbidoul could you review my fix?

sbidoul:
We should avoid placing hard pins whenever possible. And in this case the problem does not actually come from auth_jwt.
See https://github.com/OCA/server-auth/issues/424#issuecomment-1268492655 and #427.

sbidoul:
#429 is green so I'm going ahead and closing this.
See https://github.com/OCA/server-auth/pull/316 for the auth_jwt migration.

PR #429, opened 2022-10-05
Place upper bound in cryptography indirect dependency

### LINKED ISSUE #424: [15.0] external dependency "cryptography" causes error 'X509_V_FLAG_CB_ISSUER_CHECK'

sbidoul:
Fixes #424

### DIFF
--- a/auth_saml/__manifest__.py
+++ b/auth_saml/__manifest__.py
@@ -11,7 +11,14 @@
     "license": "AGPL-3",
     "depends": ["base_setup"],
     "external_dependencies": {
-        "python": ["pysaml2"],
+        # Place an upper bound on cryptography version to be compatible with
+        # pyopenssl 19 mentioned in Odoo 15's requirements.txt. If we don't do
+        # this, installing this module will try to upgrade cryptography to the latest
+        # version because the minimum required version in pysaml2 (>=3.1) is greater than
+        # version 2.6 (from Odoo's requirement.txt). Since cryptography/pyopenssl don't
+        # declare minimum supported versions, this lead to inconsistencies.
+        # https://github.com/OCA/server-auth/issues/424
+        "python": ["pysaml2", "cryptography<37"],
         "bin": ["xmlsec1"],
         # special definition used by OCA to install packages
         "deb": ["xmlsec1"],
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 # generated from manifests external_dependencies
+cryptography<37
 pysaml2


sbidoul:
runboat is green @CRogos

CRogos:
Have you tested that pysaml2 is working with cryptography < 3.1?
Otherwise it would be best to limit pysaml2 to version < 7.2 and not limit cryptography.

https://github.com/IdentityPython/pysaml2/compare/v7.1.2...v7.2.0

sbidoul:
> Otherwise it would be best to limit pysaml2 to version < 7.2 and not limit cryptography.

Why? In this case installing `pysaml2 cryptography<37` yields cryptography 36, so this should be fine?
I'm not using auth_saml though so I can't tell for sure.

CRogos:
Because with the update of pysaml2 7.1.2 -> 7.2.0 the dependency was changed:
image

There might be a reason?

CRogos (approved):
I confirm that the change fixes the build issue.

sbidoul:
> Have you tested that pysaml2 is working with cryptography < 3.1?
> Otherwise it would be best to limit pysaml2 to version < 7.2 and not limit cryptography.

In any case this should be a question for auth_saml users and maintainers.

/ocabot merge patch

CRogos:
@sbidoul I agree. Thank you, for your support. I've learned a lot.

PR #436, opened 2022-11-03
[12.0] auth_oidc: fixes

ap-wtioit:
* enable handling Open ID Connect without a custom mapping in Odoo
* enable handling Open ID Connect ID Tokens without key id

We are running this code for a few months on our servers now, so it was time to try to contribute back. The plan is to forward port it to 13.0, 14.0, 15.0, 16.0 (as we are running it for those versions too)

Info @wt-io-it

### DIFF
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -12,6 +12,7 @@ from odoo import api, exceptions, fields, models, tools
 
 try:
     from jose import jwt
+    from jose.exceptions import JWTError, JWSError
 except ImportError:
     logging.getLogger(__name__).debug("jose library not installed")
 
@@ -51,14 +52,15 @@ class AuthOauthProvider(models.Model):
     )
 
     @tools.ormcache("self.jwks_uri", "kid")
-    def _get_key(self, kid):
+    def _get_keys(self, kid):
         r = requests.get(self.jwks_uri)
         r.raise_for_status()
         response = r.json()
-        for key in response["keys"]:
-            if key["kid"] == kid:
-                return key
-        return {}
+        # the keys returned here should follow
+        # JWS Notes on Key Selection
+        # https://datatracker.ietf.org/doc/html/draft-ietf-jose-json-web-signature#appendix-D
+        return [key for key in response["keys"]
+                if kid is None or key.get("kid", None) == kid]
 
     def _map_token_values(self, res):
         if self.token_map:
@@ -72,18 +74,30 @@ class AuthOauthProvider(models.Model):
         self.ensure_one()
         res = {}
         header = jwt.get_unverified_header(id_token)
-        res.update(
-            jwt.decode(
-                id_token,
-                self._get_key(header.get("kid")),
-                algorithms=["RS256"],
-                audience=self.client_id,
-                access_token=access_token,
-            )
-        )
-
+        res.update(self._decode_id_token(access_token, id_token, header.get("kid")))
         res.update(self._map_token_values(res))
         return res
+
+    def _decode_id_token(self, access_token, id_token, kid):
+        keys = self._get_keys(kid)
+        if len(keys) > 1 and kid is None:
+            # https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.10.1
+            # If there are multiple keys in the referenced JWK Set document, a kid
+            # value MUST be provided in the JOSE Header.
+            raise JWTError("OpenID Connect requires kid to be set if there is more"
+                           " than one key in the JWKS")
+        error = None
+        # we accept multiple keys with the same kid in case a key gets rotated.
+        for key in keys:
+            try:
+                values = jwt.decode(id_token, key, algorithms=["RS256"],
+                                    audience=self.client_id, access_token=access_token)
+                return values
+            except (JWTError, JWSError) as e:
+                error = e
+        if error:
+            raise error
+        return {}
 
 
 class AuthOauthProviderGroupLine(models.Model):
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -70,7 +70,12 @@ class ResUsers(models.Model):
             ).json()
             validation.update(data)
         # required check
-        if not validation.get("user_id"):
+        if "sub" in validation and "user_id" not in validation:
+            # set user_id for auth_oauth, user_id is not an OpenID Connect standard
+            # claim:
+            # https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+            validation["user_id"] = validation["sub"]
+        elif not validation.get("user_id"):
             _logger.error("user_id claim not found in id_token (after mapping).")
             raise AccessDenied()
         # retrieve and sign in user
--- a/auth_oidc/readme/HISTORY.rst
+++ b/auth_oidc/readme/HISTORY.rst
@@ -1,3 +1,24 @@
+12.0.1.2.0 2022-11-22
+~~~~~~~~~~~~~~~~~~~~~
+
+* Fix handling OpenID Connect responses without custom mapping  by using the ``sub``  claim as user id
+* Fix handling OpenID Connect ID Tokens without Key ID (``kid``)
+
+12.0.1.1.0 2022-11-22
+~~~~~~~~~~~~~~~~~~~~~
+
+* Enable allowing assigning groups from token claims
+
+12.0.1.0.1 2022-02-28
+~~~~~~~~~~~~~~~~~~~~~
+
+* Updated readme and pot
+
+12.0.1.0.0 2022-02-12
+~~~~~~~~~~~~~~~~~~~~~
+
+* Backport to Odoo 12
+
 13.0.1.0.0 2020-04-10
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/auth_oidc/__manifest__.py
+++ b/auth_oidc/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Authentication OpenID Connect",
-    "version": "12.0.1.1.0",
+    "version": "12.0.1.2.0",
     "license": "AGPL-3",
     "author": (
         "ICTSTUDIO, André Schenkels, "
--- a/auth_oidc/readme/CONTRIBUTORS.rst
+++ b/auth_oidc/readme/CONTRIBUTORS.rst
@@ -1,2 +1,3 @@
 * Alexandre Fayolle <alexandre.fayolle@camptocamp.com>
 * Stéphane Bidoul <stephane.bidoul@acsone.eu>
+* Andreas Perhab <andreas.perhab@wt-io-it.at>
--- a/auth_oidc/tests/test_auth_oidc_auth_code.py
+++ b/auth_oidc/tests/test_auth_oidc_auth_code.py
@@ -8,9 +8,12 @@ from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 from urllib.parse import parse_qs, urlparse
 from jose import jwt
+from jose.exceptions import JWTError
+from jose.utils import long_to_base64
 
 import odoo
 from odoo.tests import common
+from odoo.exceptions import AccessDenied
 
 from odoo.addons.auth_oidc.controllers.main import OpenIDLogin
 from odoo.addons.website.tools import MockRequest as _MockRequest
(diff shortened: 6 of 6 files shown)


ap-wtioit:
Tests failing for `travis_install_nightly` because somehow they are still running with python3.5 and [pylint-plugin-utils](https://pypi.org/project/pylint-plugin-utils/) requires: `Python >=3.6.2`:
```
DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.

DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.

ERROR: Could not find a version that satisfies the requirement pylint-plugin-utils==0.7 (from -r /home/travis/maintainer-quality-tools/travis/pylint_odoo_requirements.txt (line 1)) (from versions: 0.1, 0.1.1, 0.2, 0.2.1, 0.2.2, 0.2.3, 0.2.4, 0.2.6, 0.3, 0.4, 0.5, 0.6)

ERROR: No matching distribution found for pylint-plugin-utils==0.7 (from -r /home/travis/maintainer-quality-tools/travis/pylint_odoo_requirements.txt (line 1))

The command "travis_install_nightly" failed and exited with 1 during .
```

sbidoul:
Trying to fix ci in https://github.com/OCA/server-auth/pull/438

auth_oidc/readme/HISTORY.rst
```diff
@@ -1,3 +1,19 @@
+12.0.2.0.0 2022-11-03
+~~~~~~~~~~~~~~~~~~~~~
+
+* Fix handling OpenID Connect responses without custom mapping
```
sbidoul:
```suggestion
* Fix handling OpenID Connect responses without custom mapping by using the ``sub``  claim as user id
```

auth_oidc/readme/HISTORY.rst
```diff
@@ -1,3 +1,19 @@
...
+* Fix handling OpenID Connect ID Tokens without key ids
```
sbidoul: I'm not sure what that means. Do you have a link to the standard that talks about this?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -68,15 +66,21 @@ def _parse_id_token(self, id_token, access_token):
...
-                id_token,
-                self._get_key(header.get("kid")),
-                algorithms=["RS256"],
-                audience=self.client_id,
-                access_token=access_token,
-            )
-        )
-
+        error = None
+        for key in self._get_keys(header.get("kid")):
```
sbidoul: Do you have link to the spec that shows that kid can contain a list?
I don't find that in https://openid.net/specs/openid-connect-core-1_0.html for instance.

auth_oidc/models/res_users.py
```diff
@@ -64,7 +64,11 @@ def auth_oauth(self, provider, params):
             raise AccessDenied()
         validation = oauth_provider._parse_id_token(id_token, access_token)
         # required check
-        if not validation.get("user_id"):
+        if "sub" in validation and not "user_id" in validation:
+            # set user_id for auth_oauth, user_id is not an OpenID Connect standard claim:
+            # https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+            validation["user_id"] = validation["sub"]
```
sbidoul, line 77: This looks like a useful improvement, thanks.

sbidoul (commented):
Thanks for this contrib. I've a couple of questions.

auth_oidc/models/auth_oauth_provider.py
ap-wtioit: i interpreted the following:
>  If there are multiple keys in the referenced JWK Set document, a kid value MUST be provided in the JOSE Header

from [OpenID Connect Core Section 10.1](https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.10.1)

As the `kid` is optional. And missed the as long as there is only one key in the JWK Set (jwks_uri) part. So with your question in mind, it should be unique and not a list. (An our current provider implementation is wrong). I will update the merge request.

auth_oidc/readme/HISTORY.rst
ap-wtioit: thanks, added

auth_oidc/readme/HISTORY.rst
ap-wtioit: When a JWKS has only one key it is not required for the signed ID Token to contain a `kid` parameter. Before the fix one would get an error like this in this case:
```
odoo_1                               | 2022-11-04 13:16:12,248 1 ERROR devel odoo.addons.auth_oauth.controllers.main: OAuth2: 'kid' 
odoo_1                               | Traceback (most recent call last):
odoo_1                               |   File "/opt/odoo/custom/src/odoo/odoo/tools/cache.py", line 88, in lookup
odoo_1                               |     r = d[key]
odoo_1                               |   File "/opt/odoo/custom/src/odoo/odoo/tools/func.py", line 69, in wrapper
odoo_1                               |     return func(self, *args, **kwargs)
odoo_1                               |   File "/opt/odoo/custom/src/odoo/odoo/tools/lru.py", line 44, in __getitem__
odoo_1                               |     a = self.d[obj].me
odoo_1                               | KeyError: ('auth.oauth.provider', <function AuthOauthProvider._get_key at 0x7f4d8d0d5950>, 'http://redacted.doodba/redacted/jwks', None)
odoo_1                               | 
odoo_1                               | During handling of the above exception, another exception occurred:
odoo_1                               | 
odoo_1                               | Traceback (most recent call last):
odoo_1                               |   File "/opt/odoo/auto/addons/auth_oauth/controllers/main.py", line 170, in signin
odoo_1                               |     credentials = env['res.users'].sudo().auth_oauth(provider, kw)
odoo_1                               |   File "/opt/odoo/auto/addons/auth_oidc/models/res_users.py", line 65, in auth_oauth
odoo_1                               |     validation = oauth_provider._parse_id_token(id_token, access_token)
odoo_1                               |   File "/opt/odoo/auto/addons/auth_oidc/models/auth_oauth_provider.py", line 75, in _parse_id_token
odoo_1                               |     self._get_key(header.get("kid", None)),
odoo_1                               |   File "<decorator-gen-116>", line 2, in _get_key
odoo_1                               |   File "/opt/odoo/custom/src/odoo/odoo/tools/cache.py", line 93, in lookup
odoo_1                               |     value = d[key] = self.method(*args, **kwargs)
odoo_1                               |   File "/opt/odoo/auto/addons/auth_oidc/models/auth_oauth_provider.py", line 55, in _get_key
odoo_1                               |     if key["kid"] == kid:
odoo_1                               | KeyError: 'kid'
odoo_1                               | 2022-11-04 13:16:12,250 1 INFO devel werkzeug: 172.16.3.129 - - [04/Nov/2022 13:16:12] "GET /auth_oauth/signin?code=redacted&state=redacted HTTP/1.0" 303 - 2 0.001 0.362
```

I updated the fix to handle only the case where a ID Token without `kid` in header is used with a JWKS that contains a single key.

from [OpenID Connect Core Section 10.1](https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.10.1)
>  If there are multiple keys in the referenced JWK Set document, a kid value MUST be provided in the JOSE Header

auth_oidc/models/auth_oauth_provider.py
ap-wtioit: i updated the commit to only handle cases for JWKS with a single key

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -52,8 +52,9 @@ def _get_key(self, kid):
         r.raise_for_status()
         response = r.json()
         for key in response["keys"]:
-            if key["kid"] == kid:
+            if key.get("kid", None) == kid:
                 return key
+
```
sbidoul: I think it would be easier to read (by me at least :) if it was `if kid is None: return somekey else: return key["kid"]`.

And, do you have a standard reference that says which key should be used in that case? Here you take the "first" one. But why the first?

ap-wtioit: Well, no i do not take the first one. I only take the key that has no `kid` specified. As this is only allowed when only one key is specified in the JWKS it should be also the first (and last) one.

Your proposed code would get rid of the check that if ID Token has no kid specified in header the JWKS also needs to have no kid specified for the key.

auth_oidc/models/auth_oauth_provider.py
ap-wtioit: Well, as i started checking our implementation of JWK, i checked the docs again

[OpenID ID Token Validation](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation)
Would be our starting point.
Step 6:
> If the ID Token is received via direct communication between the Client and the Token Endpoint (which it is in this flow), the TLS server validation MAY be used to validate the issuer in place of checking the token signature. The Client MUST validate the signature of all other ID Tokens according to [JWS](https://openid.net/specs/openid-connect-core-1_0.html#JWS) [JWS] using the algorithm specified in the JWT alg Header Parameter. The Client MUST use the keys provided by the Issuer.

Would be the one that defines we MUST validate the signature according to JWS

[JWS Note on Key Seelection](https://datatracker.ietf.org/doc/html/draft-ietf-jose-json-web-signature#appendix-D)
Defines which keys should be considered.
Step 2:
>Filter the set of collected keys.  For instance, some applications will use only keys referenced by "kid" (key ID) or "x5t" (X.509 certificate SHA-1 thumbprint) parameters. If the application uses the "alg" (algorithm), "use" (public key use), or "key_ops" (key operations) parameters, keys with keys with inappropriate values of those parameters would be excluded.       Additionally, keys might be filtered to include or exclude keys with certain other member values in an application specific manner.  For some applications, no filtering will be applied

And the following sections all sound like we should allow multiple keys (so the initial for loop was more correct than the current implementation)

So it seems for OpenID Connect:
* when no kid is specified in ID Token, the JWKS can only contain one key (According to the OpenID Connect Spec)
* when kid is specified in ID Token we need to find the matching Key from JWKS because there can be multiple keys for the same kid. The note at the end of 
[JWS Note on Key Seelection](https://datatracker.ietf.org/doc/html/draft-ietf-jose-json-web-signature#appendix-D) explicitly allow the behaviour of checking the signature against all keys with the same kid:
> Note that it is reasonable for some applications to perform signature
   or MAC validation prior to making a trust decision about a key, since
   keys for which the validation fails need no trust decision.

As JWS only specifies key selection i would conclude that if ID Token gives no kid it is just a missing hint for the correct key so it should still be allowed to have an ID Token without a kid and a JWKS with a single key that has a kid and validate against it.

auth_oidc/models/auth_oauth_provider.py
ap-wtioit: As per my findings in https://github.com/OCA/server-auth/pull/436#discussion_r1015223912 i would roll back the roll back of the for loop (So we have a _get_keys() again).

And i would also change the code so an ID Token without a kid can match a single key in JWKS.

@sbidoul Let me know if this approach looks good to you.

sbidoul: Yes, sounds good, thanks!

hbrunn (approved):

hbrunn:
@ap-wtioit please rebase

ap-wtioit:
@hbrunn i rebased the pull request

hbrunn:
@sbidoul could you re-review?

hbrunn:
@sbidoul can you have another look?

ap-wtioit:
@sbidoul do you have some time to review this again?

wtaferner (approved):

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -72,19 +74,30 @@ def _parse_id_token(self, id_token, access_token):
...
+    def _decode_id_token(self, access_token, id_token, kid):
+        keys = self._get_keys(kid)
+        if len(keys) > 1 and kid is None:
+            # https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.10.1
+            # If there are multiple keys in the referenced JWK Set document, a kid
+            # value MUST be provided in the JOSE Header.
+            raise JWTError("OpenID Connect requires kid to be set if there is more"
+                           " than one key in the JWKS")
+        error = None
+        for key in keys:
```
ap-wtioit, line 91: Review at oca days: add a comment explaining multiple keys with same kid when key get rotated (for still allowing tokens within their lifetime)

ap-wtioit: done with 0d15649acb159fa1ba834837f9597922fd91daa9

sbidoul:
Tests are red due to auth_saml which has a dependency on lasso which is not available on PyPI.
So merging manually, as I don't see an easy way out for lasso.

PR #445, opened 2022-11-22
[12.0][IMP] auth_oidc: temporarily enable signup for user creation

hbrunn:
This allows to give the template user high privileges (ie internal user). Without this, you'd have to allow signup and anyone would end up as highly privileged user

### DIFF
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -30,7 +30,6 @@ class ResUsers(models.Model):
         response = requests.post(
             oauth_provider.token_endpoint,
             data=dict(
-                client_id=oauth_provider.client_id,
                 grant_type="authorization_code",
                 code=code,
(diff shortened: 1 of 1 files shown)


auth_oidc/models/res_users.py
```diff
@@ -30,7 +30,6 @@ def _auth_oauth_get_tokens_auth_code_flow(self, oauth_provider, params):
         response = requests.post(
             oauth_provider.token_endpoint,
             data=dict(
-                client_id=oauth_provider.client_id,
```
hbrunn, line 33: this is necessary for [okta](https://okta.com), as it validates more strictly - also according to the [specs](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) it's wrong to pass this when the authentication method is `client_secret_basic` (the default, we don't do anything else)

antonioburic (approved):

PR #448, opened 2022-11-28
[14.0][FIX] password_security: Correctly look at previous password history

adekock11:
If a user has previous passwords stored and the `password_history` is set to 0 to disable the check, the line of code would evaluate to:

```
rec_id.password_history_ids[0:-1]
```

The above returns all items in the list except for the first last entry. One expects the above to evaluate to:

```
rec_id.password_history_ids[0:0]
```

The above returns an empty list which would then correctly not raise a warning.

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -198,7 +198,7 @@ class ResUsers(models.Model):
             if recent_passes < 0:
                 recent_passes = rec_id.password_history_ids
             else:
-                recent_passes = rec_id.password_history_ids[0 : recent_passes - 1]
+                recent_passes = rec_id.password_history_ids[0 : recent_passes]
             if recent_passes.filtered(
                 lambda r: crypt.verify(password, r.password_crypt)
             ):


ivantodorovich (changes requested):
I haven't tested, but I think you're correct, without this fix the number of history passwords to check is not computed properly, but I would expand it a little bit more to completely skip this check in case `recent_passes == 0`, and that would save us the read on `password_history_ids`.

```py
if not recent_passes:    # disabled
    continue
elif recent_passes < 0:    # unlimited
    recent_passes = rec_id.password_history_ids
else:
    recent_passes = rec_id.password_history_ids[:recent_passes]
```

Unit tests to reproduce the bug you're fixing would be appreciated 🙏🏻 it's the best way to fight regressions

astirpe:
Hi @adekock11, I think the proposal of @ivantodorovich is a nice one. Are you going to implement it and complete the tests?

amh-mw (changes requested):
@adekock11 For password_security/tests/test_res_users.py

```python
    def test_check_password_history(self):
        # Disable all password checks except for history
        set_param = self.env['ir.config_parameter'].sudo().set_param
        set_param('auth_password_policy.minlength', 0)
        user = self.env.ref('base.user_admin')
        user.company_id.update({
            'password_estimate': 0,
            'password_length': 0,
            'password_lower': 0,
            'password_history': 1,
            'password_numeric': 0,
            'password_special': 0,
            'password_upper': 0,
        })

        self.assertEqual(len(user.password_history_ids), 0)

        user.write({'password': 'admin'})
        self.assertEqual(len(user.password_history_ids), 1)

        with self.assertRaises(UserError):
            user.write({'password': 'admin'})
        user.write({'password': 'admit'})
        self.assertEqual(len(user.password_history_ids), 2)

        user.company_id.password_history = 2
        with self.assertRaises(UserError):
            user.write({'password': 'admin'})
        with self.assertRaises(UserError):
            user.write({'password': 'admit'})
        user.write({'password': 'badminton'})
        self.assertEqual(len(user.password_history_ids), 3)

        user.company_id.password_history = 0
        user.write({'password': 'badminton'})
        self.assertEqual(len(user.password_history_ids), 4)
```

Edit: Added test for password_history = 0

PR #449, opened 2022-11-29
[FIX] Update copier template

bruno-zanotti:
To fix the Flake8 Gitlab problem.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,5 +1,5 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.11.0
+_commit: v1.11.1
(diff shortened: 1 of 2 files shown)


bruno-zanotti:
Hi @pedrobaeza !

This is a FW port of your [PR](https://github.com/OCA/server-auth/pull/447) in 14 to be able to migrate modules to V16. Could you merge it?

pedrobaeza (approved):
Thanks for doing it. You can obtain it as well running `pipx run copier -f update`, as we have a template with the fix that can be propagated to OCA repos since 13.0.

Anyway, let's merge it, as the result is the same!

bruno-zanotti:
@pedrobaeza 
Oh nice, I didn't know that, thanks!

PR #456, opened 2023-01-02
[15.0][mig] password_security

dsolanki-initos:


password_security/tests/test_password_security_home.py
```diff
@@ -0,0 +1,259 @@
...
+            main.AuthSignupHome, **{m: mock.DEFAULT for m in methods}
+        ) as _super:
+            mocks = {}
+            for method in methods:
+                mocks[method] = _super[method]
+                mocks[method].return_value = MockResponse()
+            with mock.patch("%s.request" % IMPORT) as request:
+                with mock.patch("%s.ensure_db" % IMPORT) as ensure:
+                    with mock.patch("%s.http" % IMPORT) as http:
+                        http.redirect_with_hash.return_value = MockResponse()
```
amh-mw: Mock request.redirect here instead?

password_security/models/res_users.py
```diff
@@ -0,0 +1,208 @@
...
+        password_regex = [
+            "^",
+            "(?=.*?[a-z]){" + str(company_id.password_lower) + ",}",
+            "(?=.*?[A-Z]){" + str(company_id.password_upper) + ",}",
+            "(?=.*?\\d){" + str(company_id.password_numeric) + ",}",
+            r"(?=.*?[\W_]){" + str(company_id.password_special) + ",}",
+            ".{%d,}$" % int(company_id.password_length),
+        ]
+        if not re.search("".join(password_regex), password):
+            raise UserError(self.password_match_message())
```
amh-mw: Throw ValidationError here instead to match auth_saml test expectations?

amh-mw (changes requested):
I was just starting to migrate this module myself, but noticed that you have already made some good progress, so let me see if I can help you out instead. I am still investigating the "object unbound" test failures.

password_security/models/res_users.py
amh-mw: Alternately, #457 should work around this without having to change any exception classes.

amh-mw (changes requested):
It looks like `odoo.http.request` is a proxy for `werkzeug.local.LocalStack` and the "object unbound" errors are due to the stack being accessed when it does not have a current request. There are some examples of how to work around this, i.e. `odoo.addons.website.tools.MockRequest` and the setUp function of `odoo/addons/base/tests/test_xmlrpc.py`, but nothing readily usable. I've worked up a small patch [^1] that pushes mock requests onto the stack for the duration of the tests.

[^1]: https://gist.github.com/amh-mw/a7cee5ede165fc4b54a3ca504cbbebe9

password_security/models/res_users.py
dsolanki-initos: @amh-mw, I've done changes as per your feedback.

dsolanki-initos:
> It looks like `odoo.http.request` is a proxy for `werkzeug.local.LocalStack` and the "object unbound" errors are due to the stack being accessed when it does not have a current request. There are some examples of how to work around this, i.e. `odoo.addons.website.tools.MockRequest` and the setUp function of `odoo/addons/base/tests/test_xmlrpc.py`, but nothing readily usable. I've worked up a small patch [1](#user-content-fn-1-de421ae020878bd41fe58498609f3625) that pushes mock requests onto the stack for the duration of the tests.
> 
> ## Footnotes
> 1. https://gist.github.com/amh-mw/a7cee5ede165fc4b54a3ca504cbbebe9 [leftwards_arrow_with_hook](#user-content-fnref-1-de421ae020878bd41fe58498609f3625)

@amh-mw, Thank you for the suggestion regarding the test case failure, I'll test your patch locally and give you update regarding it.

dsolanki-initos:
@amh-mw, The test case failure due to "object unbound" is fixed with your suggested patch. Also, fixed other errors Please review it.

Thank you for the support

amh-mw (changes requested):
Rather than review the full hundred files, I have individually reviewed the last four commits. I think you need to `git fetch upstream`, then `git rebase -i upstream/15.0`, deleting merge commit 285b284f9d2c5e6ac6172b5792fa89188167a582 and squashing 46c46663ac77ef383ddb8b77032581079b356c97 into the migration commit, then force push back up to get a clean commit history. Otherwise LGTM.

dsolanki-initos:
I'll fix the commits as per your feedback.

amh-mw (approved):
@OCA/core-maintainers LGTM. Merge?

amh-mw:
@OCA/tools-maintainers Ready for review and merge?

password_security/models/res_users.py
```diff
@@ -102,7 +102,7 @@ def password_match_message(self):
             )
         if company_id.password_special:
             message.append(
-                _("\n* Special character (at least % characters)")
+                _("\n* Special character (at least %s characters)")
```
thomaspaulb, line 105: If you change it here, you should also update it in the `.pot` file.

password_security/models/res_users.py
```diff
@@ -198,6 +198,7 @@ def _check_password_history(self, password):
                     _("Cannot use the most recent %d passwords")
                     % rec_id.company_id.password_history
                 )
+        return
```
thomaspaulb: Does this really do anything?

thomaspaulb (approved):
Only minor things LGTM

password_security/models/res_users.py
dsolanki-initos: It does nothing, I'll remove it.

ivantodorovich (approved):
LG

PR #459, opened 2023-01-11
[13.0][FIX] vault_share: File is not downloaded when clicking sharing link

CarlosRoca13:
I add a commit to fix precommit too

cc @Tecnativa TT41171

Before this patch, when we share a file and click on generated link, the file was not downloaded:
Error_Vault

After this patch, the file is normally downloaded as it does on backend:
Error_vault_resuelto

ping @pedrobaeza @chienandalu @fkantelberg


pedrobaeza (commented):
Can we set a buffered stream or similar to avoid to download the whole file at the beginning? There should be another technique...

chienandalu (approved):
The implementation of the module is kinda ugly. No easily fixable without heavy refactoring (I'd convert all this stuff in the browser to calls to python controllers doing the thing in the server).

The patch does the job anyway :+1:

fkantelberg:
The problem seems to come from the encoding in the share frontend. Which is especially problematic for binary files. The fix does the job.

You can only delay the file download until the pin is entered the first time. Storing giant encrypted files isn't really healthy for the module and should be done on different ways.

Moving the ugly stuff from the browser is against the design of the module where the encrypting/decrypting should happen on the client. (end-to-end-encryption)

chienandalu:
> Moving the ugly stuff from the browser is against the design of the module where the encrypting/decrypting should happen on the client. (end-to-end-encryption)

Ok, I understand :)

pedrobaeza (approved):
OK, let's accept this with its limitations.

/ocabot merge patch

Please fw-port it to upper versions

pedrobaeza:
Hey, there's a new linter in town. You have to remove `users_ldap_mail/i18n/en.po`, @CarlosRoca13

CarlosRoca13:
@pedrobaeza I add it the change to fix pre-commit commit

PR #463, opened 2023-01-17
[FIX][14.0]users_ldap_mail force the cast in string in case of the Active directory is prov…

flotho:
…inding byte-code strings

when working with AD, email are sometimes returned as byte-code issuing this message :
```
  File "/home/odoo_14_upa/additionnal_repositories/OCA-partner-contact/partner_email_check/models/res_partner.py", line 37, in email_check
    for email in emails.split(",")
Exception

The above exception was the direct cause of the following exception:

  File "/home/odoo_14_upa/server/odoo/http.py", line 641, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/odoo_14_upa/server/odoo/http.py", line 317, in _handle_exception
    raise exception.with_traceback(None) from new_cause
TypeError: a bytes-like object is required, not 'str'
```

This fix handle this case already used in other part of this module

### DIFF
--- a/users_ldap_mail/models/users_ldap_model.py
+++ b/users_ldap_mail/models/users_ldap_model.py
@@ -3,7 +3,7 @@
 
 import logging
 
-from odoo import fields, models
+from odoo import fields, models, tools
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


flotho:
ping @primes2h @etobella any chance to have a review on this one ?

etobella (approved):

flotho:
ping @joao-p-marques any chance to have a reviw on this one ?

pedrobaeza (approved):
Joao is no longer in the Odoo would.

/ocabot merge patch

flotho:
> Joao is no longer in the Odoo would.

thanks for the information @pedrobaeza and alos thanks for the merge on a WE ;-)

PR #464, opened 2023-01-18
[15.0][IMP] password_security: Match defaults to NIST 800-63B.

amh-mw:
See https://pages.nist.gov/800-63-3/sp800-63b.html#memsecret

Edit: See also https://xkcd.com/936/
Edit: See also https://www.youtube.com/watch?v=vf37jh3dV2I (research talk by zxcvbn author)


amh-mw:
Depends on #456.

amh-mw:
@astirpe Opinions on this one?

password_security/tests/test_res_users.py
```diff
@@ -67,9 +72,34 @@ def test_check_password_returns_true_for_valid_password(self):
...
     def test_check_password_raises_error_for_invalid_password(self):
+        self.main_comp.password_length = 12
         with self.assertRaises(UserError):
             self.rec_id._check_password("password")
 
+    def test_check_password_lower(self):
+        self.rec_id._check_password("AB3!")
+        self.main_comp.password_lower = 1
+        with self.assertRaises(UserError):
+            self.rec_id._check_password("AB3!")
```
astirpe, line 84: I see all the tests are calling `with self.assertRaises(UserError)` without any distinction. So actually there's no check that the correct error message is shown.

Making use of `self.assertRaisesRegex()` would improve the tests, ensuring that the expected error is raised.

For example here it would be something like:

```python
        with self.assertRaisesRegex(UserError, ".*Lowercase letter (at least %s characters).*" % str(company_id.password_lower)):
            self.rec_id._check_password("AB3!")
```

What do you think?

astirpe (commented):
@amh-mw I don't have any opinion regarding the functionality you are implementing, since it seems to me more an US specific requirement rather that a generic requirement.
Maybe a functional expert guy could answer your question, much better then me for sure :)

What I liked in the old implementation is that, when the user installs the module, all the restrictions are applied by default. This way the user has a more "secure" password policy already applied, as it would be expected by just installing the module.

amh-mw:
> What I liked in the old implementation is that, when the user installs the module, all the restrictions are applied by default. This way the user has a more "secure" password policy already applied, as it would be expected by just installing the module.

I think this is the trap. I just added a YouTube link up top to a research talk the author of zxcvbn gave. These sorts of rules are challenging for end users and cause them to create passwords that are hard to remember, but still easy for password crackers to brute force.

password_security/tests/test_res_users.py
amh-mw: Absolutely, and the worst part is that I was watching Simone Orsi's Odoo Days 2020 talk [^1] a couple weeks ago and he made the same suggestion about testing error messages.

[^1] https://www.youtube.com/watch?v=pQ7TZELSpKY

amh-mw:
Rebased and repaired tests.

amh-mw:
@OCA/tools-maintainers May I please have `nostale` applied or `stale` removed?

pedrobaeza:
Label removed.

PR #470, opened 2023-02-02
[14.0][FIX] Update copier template

sergio-teruel:
cc @Tecnativa


pedrobaeza:
14.0 shouldn't needed for now to update the copier template, but it's not bad also to have the latest definition.

Issue #141, opened 2019-10-17
`password_security` module throws warning during tests


jmcvetta:
The `password_security` module throws a warning during tests.  Although it is just a warning, it makes odoo.sh flag the tests as yellow.  So an Odoo instance using `password_security` cannot pass its tests on odoo.sh.

Sample warning from the log:
```
2019-10-10 11:18:13,150 7 WARNING jmcvetta-dev-632767 odoo.http: <function odoo.addons.password_security.controllers.main.web_auth_signup> returns an invalid response type for an http request 
```

Warning is thrown from within `odoo.http.route` decorator, which is wrapping `password_security.controllers.main.web_auth_signup()`.  

Within that method we have the [line](https://github.com/OCA/server-auth/blob/191da97f415f93aed3d2bf4892dc2c137f63c428/password_security/controllers/main.py#L62):

```python3
            return request.render('auth_signup.signup', qcontext)
```

`@route` is unhappy because the response from `request.render()` is an instance of `MagicMock`.  Whereas `@route` is expecting an instance of `odoo.http.Response`, or one of a few other types.  

Need to change the [mocking setup](https://github.com/OCA/server-auth/blob/191da97f415f93aed3d2bf4892dc2c137f63c428/password_security/tests/test_password_security_home.py#L44), so the `MagicMock` object will report itself as an instance of `odoo.http.Response`.

pedrobaeza:
Odoo version? Can you propose the pull request?

Issue #89, opened 2019-04-22
[11] Password Security - Disable history requires 1 not 0


samdohrey:
Hello,

We are hoping to install the Password security module but noticed a strange issue. We wish to disable the history check, as per the pop-up instructions "0 to disable". However this does not seem to work, we have to enter the value 1. This seems minor but could cause frustration to new users of the module

**Steps:**
In the password policy set (for ease of use):
Days = 0
Minimum Hours = 0
Characters = 0
**History = 0**
Lower = 1
Upper = 1
Numeric = 1
Special = 0

Ensure users can reset their own passwords
Create a new user, login and logout as said user
Using the Reset password mechanism on the login page
Set new password as Abcd1       Login
Set new password as Abcd1       Again. Login. User was able to use same password
Set new password as Abcd1       User sees error "Cannot use the most recent 0 passwords"

Login as a user with settings permissions and in password policy set History to 1.
Repeat the above test, the user can reuse the same password unlimited times

Expected:
On the third attempt the user can use the repeat password when history = 0

Kind regards,

Sam

LuceEasyPme:
Thank you! It helped a lot!
Luce

PR #482, opened 2023-02-20
[16.0][MIG/IMP] password_security

astirpe:
Together with the module migration, I'm proposing a revised version of the module:

- [x] Removed old widget, replaced by "Password Meter" widget by standard Odoo
- [x] Unify field `password_length` with standard field `minlength`, to avoid incompatibility with standard module `auth_password_policy_portal`
- [x] Rewrite tests for controller, needed because of new Odoo 16 changes
- [x] Includes https://github.com/OCA/server-auth/pull/492
- [x] Includes https://github.com/OCA/server-auth/pull/494
- [x] Remove password estimate functionality (see https://github.com/OCA/server-auth/pull/482#issuecomment-1653293088)
- [x] Fixes warning in tests https://github.com/OCA/server-auth/issues/141


amh-mw:
@astirpe Nice job; tremendous amount of work. It feels like it should be six pull requests. If you would be willing to break it up, I'd be willing to review all the parts. I think it's a bit dangerous to do it all in one go, especially with the unit tests being rewritten. How would you feel about fixing each of the items you marked for back-porting in 15.0 as separate pull requests and getting those merged first, so that your migration commit isn't quite so expansive?

astirpe:
@amh-mw, yes that's a good idea!

I backported two of the fixes: https://github.com/OCA/server-auth/pull/491 (constraint of password_estimate range) and https://github.com/OCA/server-auth/pull/492 (Reset Password issue)

The fix for https://github.com/OCA/server-auth/issues/89 is the one already discussed in https://github.com/OCA/server-auth/pull/448 and that's still an open PR.

The fix for https://github.com/OCA/server-auth/issues/141 is just a consequence of the complete rewrite of the mocked tests.

For the rest, I don't think the other points should be backported to V15.
Once the above PRs for V15 will be merged, I will include the related commits in this PR. Is that ok for you?

amh-mw:
> Once the above PRs for V15 will be merged, I will include the related commits in this PR. Is that ok for you?

I think it would make more sense to rebase against 15.0 afterwards, so that you just have the one migration commit.

password_security/models/res_company.py
```diff
@@ -14,11 +14,6 @@ class ResCompany(models.Model):
         default=60,
         help="How many days until passwords expire",
     )
-    password_length = fields.Integer(
-        "Characters",
-        default=12,
-        help="Minimum number of characters",
-    )
```
amh-mw: Do you need to provide a migration for this field? Maybe something like:

```
ICP = self.env['ir.config_parameter']
ICP.set_param('auth_password_policy.minlength', max(
    ICP.get_param('auth_password_policy.minlength'),
    *self.env['res.company'].search([]).mapped('password_length'))
))
```

password_security/controllers/main.py
```diff
@@ -1,42 +1,30 @@
...
-        return request.env["res.users"].get_estimation(password)
+        user = request.env.user
+        user._check_password(password)
+        return super().do_signup(qcontext)
 
     @http.route()
     def web_login(self, *args, **kw):
         ensure_db()
-        response = super(PasswordSecurityHome, self).web_login(*args, **kw)
+        response = super().web_login(*args, **kw)
```
amh-mw: On a big change like this, I would leave style items like this alone to make the diff smaller. Odoo codebase seems to prefer the explicit form anyway.

password_security/views/res_config_settings_views.xml
```diff
@@ -14,10 +14,10 @@
         <field name="arch" type="xml">
             <!-- We hide Odoo's minlength sections so it does not create confusion -->
             <xpath expr="//label[@for='minlength']/../.." position="attributes">
-                <attribute name="style">display: none</attribute>
+                <attribute name="invisible">1</attribute>
```
amh-mw, line 17: Seems weird to continue to hide `minlength` now that it has been adopted over `password_length`.

amh-mw (changes requested):
Just poking around a bit while waiting for the maintainers to merge the 15.0 pull requests. I'll give it a more thorough review after.

password_security/models/res_company.py
astirpe: Yes, seems a good idea :+1:  Thanks!

password_security/views/res_config_settings_views.xml
astirpe: It's just because I wanted to embed the `minlength` inside the group of fields "Password Policy"

https://github.com/OCA/server-auth/blob/6eab9a5d01c5fa49263ea9aba33bade290bfc9cf/password_security/views/res_config_settings_views.xml#L53-L57

This way it looks&feels better, otherwise the `minlength` will stay detached from the rest of the fields.

password_security/models/res_company.py
astirpe: Implemented. Migration tests could be executed locally by adding `password_length = fields.Integer(default=12)` to res.company

astirpe:
No idea why OCB test now fails. Maybe a temporary glitch?

Abranes:
@astirpe Could you rebase from 16.0 to resolve the conflicts?

astirpe:
Still test in OCB failing, I cannot even reproduce the failure with OCB locally.

astirpe:
Rebased. And this is another surprise: now it's Odoo failing, not OCB anymore

password_security/models/res_users.py
```diff
@@ -0,0 +1,218 @@
...
+                raise UserError(
+                    _("Cannot use the most recent %d passwords")
+                    % user.company_id.password_history
+                )
+
+    def _set_encrypted_password(self, uid, pw):
+        """It saves password crypt history for history rules"""
+        res = super(ResUsers, self)._set_encrypted_password(uid, pw)
+
+        self.write({"password_history_ids": [(0, 0, {"password_crypt": pw})]})
```
kmagusiak: This method can be called as a (portal) user from `_check_credentials` with a replacement hash. In that case, the user does not have permissions to write.

```suggestion
        self.sudo().write({"password_history_ids": [(0, 0, {"password_crypt": pw})]})
```

astirpe: @kmagusiak thank you!
I made a fix in commit https://github.com/OCA/server-auth/pull/482/commits/4b1b2686d841b73f2594103d8218a17c989fa0c5 by setting the proper rights for portal user. Is that working for you?

kmagusiak: I haven't tested, but it looks like it should work.

jaenbox (changes requested):
I have tested the password_security module from the runboat. When enabling the module and enabling MFA, I encounter an incorrect behavior. MFA configuration is set up correctly. The problem arises when attempting a second login with this user. A 500 error occurs. In the log, it shows 'Login successful for xxxx,' but later, a singleton error is displayed: 'ValueError: Expected singleton: res.users()'.

astirpe:
@jaenbox thank you! I just reproduced the error and I'll fix it soon:

```
  File "/opt/odoo/odoo/models.py", line 5129, in ensure_one
    _id, = self._ids
ValueError: not enough values to unpack (expected 1, got 0)

During handling of the above exception, another exception occurred:

  File "/opt/odoo/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/opt/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 1725, in dispatch
    return self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/addons/website/models/ir_http.py", line 235, in _dispatch
    response = super()._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 697, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/website/controllers/main.py", line 186, in web_login
    return super().web_login(*args, **kw)
  File "/opt/odoo/odoo/http.py", line 697, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/mnt/data/odoo-addons-dir/password_security/controllers/main.py", line 31, in web_login
    if not request.env.user._password_has_expired():
  File "/mnt/data/odoo-addons-dir/password_security/models/res_users.py", line 144, in _password_has_expired
    self.ensure_one()
  File "/opt/odoo/odoo/models.py", line 5132, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: res.users()
```

astirpe:
@jaenbox done, would you give it a try?

jaenbox:
> @jaenbox done, would you give it a try?

@astirpe It's works!!!

astirpe:
@amh-mw @jaenbox @kmagusiak since you already reviewed and tested this old PR, do you mind to approve it? Thanks in advance!

kmagusiak (approved):

amh-mw (approved):

zuher83:
Any news? 🤔

thomaspaulb:
[test was failing](https://github.com/OCA/server-auth/actions/runs/5436772234/jobs/9886859848#step:8:361)

/ocabot rebase

thomaspaulb:
@astirpe Could you check out the failing test, not sure why it appeared

password_security/views/res_config_settings_views.xml
hwangh95: How about instead of removing the entire box that contains the label, we reuse it? There might have been some customization within the settings box that gets removed otherwise.

Something like this?
```
<!-- Remove label for uniform styling -->
<xpath expr="//label[@for='minlength']" position="replace"/>


<!-- Add an id for easier lookup -->
<xpath expr="//field[@name='minlength']/ancestor::div[hasclass('o_setting_box')]" position="attributes">
    <attribute name="id">password_policy</attribute>
</xpath>

<!-- Move the settings box to desired location -->
<xpath expr="//div[@id='enable_password_reset']" position="after">
    <xpath expr="//div[@id='password_policy']" position="move"/>
</xpath>


<!-- Ensure our settings will come before any former customization -->
<xpath expr="//div[@id='password_policy']//div[hasclass('o_setting_right_pane')]/*" position="before">
    <!-- Add all the settings here -->
    <span id="minlength">
         Minimum number of characters
    </span>
</xpath>


<!-- Add oe_inline attribute to minlength -->
<xpath expr="//field[@name='minlength']" position="attributes">
  <attribute name="class">oe_inline</attribute>
</xpath>

<!-- Move the minlength field to desired location -->
<xpath expr="//span[@id='minlength']" position="inside">
    <xpath expr="//field[@name='minlength']" position="move"/>
</xpath>
```

It might seem like a lot of work, but I personally feel like we should try to minimize making nodes disappear

password_security/tests/test_change_password.py
```diff
@@ -0,0 +1,136 @@
...
+    def test_04_change_password_check_password_history(self):
+        """It should fail when chosen password was previously used"""
+
+        # Set password history limit
+        user = self.env["res.users"].search([("login", "=", "admin")], limit=1)
+        user.company_id.password_history = 3
+        self.assertEqual(len(user.password_history_ids), 0)
+
+        # Change password: password history records created
+        user._change_password("!asdQWE12345_4")
```
hwangh95, line 100: Odoo stores datetime fields with time granularity of seconds. The tests are failing because the _change_password is processed fast enough for some of the history to be within the same second.

You need to add either sleeps or manually modify the 'date' field in the password_history_ids to be distinct, otherwise it will be ordered by id asc. You could also add id desc as a sort of fallback to the _order.

hwangh95 (changes requested):
Please change either the tests or the _order of 'res.users.pass.history' to have the tests pass.

cmarrero:
The reuse of previous passwords fails if user try with their last password, after the message that validate you are using an old password if you enter a new one gives you a invalid token error and you have to send again another reset to make it works @astirpe

password_security/tests/test_signup.py
```diff
@@ -0,0 +1,153 @@
...
+                    "login": "test",
+                    "csrf_token": http.Request.csrf_token(self),
+                },
+            )
+
+        # Ensure we stay in the signup page
+        self.assertEqual(response.request.path_url, "/web/signup")
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(
+            "Signup: no name or partner given for new user",
```
hwangh95, line 153: Because of auth_signup_verify_email, the error message will be "That does not seem to be an email address" rather than "Signup: no name or partner given for new user". I would try with a "valid" login format like "test@test.com"

hwangh95 (changes requested):
Tests are failing because of auth_signup_verify_email

password_security/tests/test_signup.py
hwangh95: Gaaaah auth_signup_verify_email just eats the error and throws a generic one. Maybe you could do some sort of check against the installed module? Or make a small PR to auth_signup_verify_email to add the thrown error to the error context.

password_security/views/res_config_settings_views.xml
astirpe: Yes thanks! Done in https://github.com/OCA/server-auth/pull/482/commits/b72dde148032d39043c5fd8b2bbe31461821da3f, would you check?

hwangh95: Looks mostly good to me! There's the comment that still says 

I don't know if you wanted the password_security settings to come before or after the previous nodes.

password_security/models/res_company.py
```diff
@@ -0,0 +1,57 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    password_expiration = fields.Integer(
+        "Days",
+        default=60,
```
hwangh95, line 13:
```suggestion
```
I don't know what the solution is unfortunately. I tried setting all the defaults to zero and setting the company values to the "correct" defaults on each of the tests but then the main company will have default 0 values.

I thought maybe we could use ir.defaults to put the defaults back, but the default values will only be set on a new company creation.

We can't use a post_install_hook since test suites are run with a post-install tag...

Do you have any suggestions?

hwangh95 (changes requested):
I've been doing some testing on an odoo.sh instance and unfortunately with the defaults set on company fields, anyone who installs this module on something with a stricter test suite will fail the base tests.

password_security/models/res_company.py
astirpe: I would go for setting `password_expiration` default to 0 and update the readme and the tests accordingly. What do you think?

Do you know if `password_expiration` is the only field to be fixed?

password_security/tests/test_signup.py
```diff
@@ -0,0 +1,160 @@
...
+        self.opener = Opener(self.env.cr)
+        self.opener.cookies.set("session_id", self.session.sid, domain=HOST, path="/")
+
+        with mock.patch("odoo.http.db_filter") as db_filter:
+            db_filter.side_effect = lambda dbs, host=None: [get_db_name()]
+            # Signup: no name or partner given for new user
+            response = self.url_open(
+                "/web/signup",
+                data={
+                    "login": "test@test.com",
```
hwangh95, line 142:
```suggestion
                    "login": "test@test.com",
                    "password": "test123",
```
Looking at auth_signup_verify_email, it seems like it only redirects the signup for passwordless signups. So If you set a password, it should be handled by web_auth_signup and your exception handler.

password_security/tests/test_signup.py
```diff
@@ -0,0 +1,160 @@
...
+        self.assertEqual(response.status_code, 200)
+        # Expected message would be "Signup: no name or partner given for new user",
+        # however it may be varying in case of extra module installed.
+        # For example with auth_signup_verify_email installed, expected message
+        # would be "Something went wrong, please try again later or contact us."
+        # So here we just check that a generic error message is displayed.
+        self.assertIn(
+            "alert alert-danger",
+            response.text,
+        )
```
hwangh95:
```suggestion
        self.assertIn(
            "Signup: no name or partner given for new user",
            response.text,
        )
```
By adding the "password" param, we can go back to the previous assertion

hwangh95, line 155: To cover the code in [password_security/controllers/main.py#L57-L61](https://github.com/OCA/server-auth/pull/482/files#diff-2c1c11f2ce9b2571602f9b0185b57a72e738dd34a5381a6478d1f27a2264ee87R57-R60), you can add some tests like these
```suggestion
        )
        self.assertIn("X-Frame-Options", response.headers)
        self.assertEqual(
            response.headers["X-Frame-Options"],
            "SAMEORIGIN"
        )
        
        self.assertIn("Content-Security-Policy", response.headers)
        self.assertEqual(
            response.headers["Content-Security-Policy"],
            "frame-ancestors 'self'"
        )
```

hwangh95 (changes requested):
Found some ways to get around the auth_signup_verify_email issue

password_security/models/res_company.py
hwangh95: Unfortunately it's all of them 🥲
base test errors
check password error

It's because the base test suite passwords don't fit the password requirements...

amh-mw: See https://github.com/odoo/odoo/pull/114858 for some commentary from Odoo developers on this problem.

password_security/tests/test_signup.py
astirpe: You are right, thank you!

password_security/models/res_company.py
astirpe: Regarding this problem, for the time being, I would revert https://github.com/OCA/server-auth/pull/482/commits/58adcf51fd527c1621b77b2d60842e89dd6d7ad9 and I would avoid to solve the problem within this PR. Later on we could open a separate PR in case https://github.com/odoo/odoo/pull/114858 will not go any further.
Is that ok with you all?

cmarrero:
After upgrade and enter Settings gives the following error with the refactor:

RPC_ERROR
Odoo Server Error
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/http.py", line 1584, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/http.py", line 1611, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/http.py", line 1815, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/Users/cmarrero/PycharmProjects/odoo/addons/website/models/ir_http.py", line 235, in _dispatch
    response = super()._dispatch(endpoint)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_http.py", line 154, in _dispatch
    result = endpoint(**request.params)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/http.py", line 697, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/Users/cmarrero/PycharmProjects/odoo/addons/web/controllers/dataset.py", line 42, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/Users/cmarrero/PycharmProjects/odoo/addons/web/controllers/dataset.py", line 33, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/api.py", line 457, in call_kw
    result = _call_kw_model(method, model, args, kwargs)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/api.py", line 430, in _call_kw_model
    result = method(recs, *args, **kwargs)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 2491, in get_views
    result['views'] = {
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 2492, in <dictcomp>
    v_type: self.get_view(
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 2676, in get_view
    result = dict(self._get_view_cache(view_id, view_type, **options))
  File "<decorator-gen-48>", line 2, in _get_view_cache
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/tools/cache.py", line 90, in lookup
    value = d[key] = self.method(*args, **kwargs)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 2636, in _get_view_cache
    arch, view = self._get_view(view_id, view_type, **options)
  File "/Users/cmarrero/PycharmProjects/odoo/addons/web/models/models.py", line 245, in _get_view
    arch, view = super()._get_view(view_id, view_type, **options)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 2583, in _get_view
    arch = view._get_combined_arch()
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 999, in _get_combined_arch
    arch = root.with_prefetch(tree_views._prefetch_ids)._combine(hierarchy)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 940, in _combine
    combined_arch = view.apply_inheritance_specs(combined_arch, arch)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 877, in apply_inheritance_specs
    self._raise_view_error(str(e), specs_tree)
  File "/Users/cmarrero/PycharmProjects/odoo/odoo/addons/base/models/ir_ui_view.py", line 754, in _raise_view_error
    raise err from from_exception
ValueError: Element '<xpath expr="//span[@id=&#39;minlength&#39;]">' cannot be located in parent view

The above server error caused the following client error:
RPC_ERROR: Odoo Server Error
    at makeErrorFromResponse (http://localhost:8069/web/assets/10737-0b43f28/web.assets_backend.min.js:985:163)
    at XMLHttpRequest.<anonymous> (http://localhost:8069/web/assets/10737-0b43f28/web.assets_backend.min.js:993:13)

astirpe:
@cmarrero if you are getting the error in a production server I would suggest you to restore the old code (before the config settings refactoring).

Instead if you are on a test server you can try this:

- Login as admin with enabled developer mode
- Go to Settings > Technical > User interface > Views
- Search view named "res.config.settings.form.password_security" and delete it
- Go to Apps
- Search module Password Security and click on Upgrade button
- Verify that the view named "res.config.settings.form.password_security" is re-created
- Check whether the error does not occur again

password_security/models/res_company.py
hwangh95: Makes sense. I think I'll end up having to make a custom module to prevent test failures on my end for now.

astirpe:
I just found out that when a new user is trying to signup by entering a very large password, there is a critical performance issue.
It's caused by the adoption of `zxcvbn` library, in particular the execution of this line https://github.com/dwolfhub/zxcvbn-python/blob/master/zxcvbn/__init__.py#L25 is very slow. That could potentially cause a DDOS attack in my opinion.

Probably that's also the reason of https://github.com/OCA/server-auth/issues/487 but I didn't investigate in deep that issue.

So I'm going to remove the password estimate check functionality from this module and also remove the dependency on `zxcvbn`, because the password estimation is already provided by "Password Meter" widget by standard Odoo.

I hope is ok with you all.

hwangh95:
Are there any updates on this PR? Or is it good to go?

astirpe:
Would be good for me if this PR gets merged soon

password_security/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,105 @@
...
+                    <div class="mt16">
+                        <span>
+                            Minimum number of special characters
+                            <field name="password_special" class="oe_inline" />
+                        </span>
+                    </div>
+                </div>
+                <span id="minlength">
+                     Minimum number of characters
+                </span>
```
hwangh95: I was doing some testing on my own instance and noticed this portion was slightly off
```suggestion
                      <div class="mt16">
                          <span id="minlength">
                               Minimum number of characters
                          </span>
                      </div>
                  </div>
```

hwangh95 (changes requested):
Small UI change

password_security/models/res_company.py
hwangh95: @astirpe How do you feel about having an `enable_password_security` field on res.config.settings that is related to company_id?

Then we can enable it for tests and have the setting_box attr be invisible depending on it

astirpe: @hwangh95 seems a good idea to me, however I'm very busy on a project these days. Is there any possibility for you to take over this PR for continuing the development? Would be really appreciated and I could review it later on.

zuher83:
Ready to migrate ?

antonag32:
Hello, is there anything left to do before merging? I could help with any tasks left.

PR #489, opened 2023-02-27
[14.0] [ADD] auth_pass_ip_restriction: A module to restrict password authentication to IP addresses.

ddejong-therp:
Here is a module that restricts password authentication to a whitelist of IP addresses. We think it should be useful to anyone who wishes to use another authentication module as the main authentication method, but still leave a sort of 'back door' open with password login.
It also hides the regular login 'screen', but leaves an element visible in which other authentication modules would/could add their login buttons.

We use it together with the auth_saml module.


auth_pass_ip_restriction/models/res_users.py
```diff
@@ -0,0 +1,15 @@
+from odoo import _, models
+from odoo.exceptions import AccessDenied
+from odoo.http import request
+from odoo.tools.config import config
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    def _check_credentials(self, password, env):
```
thomaspaulb, line 10: In our particular usecase there are some authentication methods which should be able to bypass this... so we should build in here that if `env` contains `bypass_ip_check: True` then we don't do the below check

auth_pass_ip_restriction/models/res_users.py
```diff
@@ -0,0 +1,15 @@
...
+        ip = request.httprequest.environ["REMOTE_ADDR"]
```
thomaspaulb, line 18: Unfortunately, [tests are failing because of this line](https://github.com/OCA/server-auth/actions/runs/4286094456/jobs/7465211683?pr=489#step:8:821), because we don't always have the `request` object available; only when this actually gets called from a HTTP context.

Odoo themselves seem to be doing this:

```
ip = request.httprequest.environ['REMOTE_ADDR'] if request else 'n/a'
```

which is a reasonable approach; so then what seems logical to me is if we can't find a request object and hence no IP, we do allow the authentication request to go through, since it doesn't come from an IP so it must be some internal check.

thomaspaulb (changes requested):

auth_pass_ip_restriction/models/res_users.py
ddejong-therp: ok, I've added a check for `request` together with the check for `bypass_ip_check`, which should have the same effect you propose I think.

auth_pass_ip_restriction/models/res_users.py
ddejong-therp: Added it.

auth_pass_ip_restriction/models/res_users.py
thomaspaulb: Yea, tests are now green. Now just fixing pre-commit and we're golden

thomaspaulb: Still some pre-commit issue

auth_pass_ip_restriction/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+When this module is installed, you can enable the regular password login form by
+adding `?debug=1` to your browser's url.
+
+Configuring the IP whitelist is done in the `odoo.conf` configuration file, with
+the `auth_pass_ip_whitelist` parameter like so:
+```
+auth_pass_ip_whitelist = ["127.0.0.1", "192.168.1.7"]
+```
+If this parameter is not set, it defaults to `["127.0.0.1"]`.
```
thomaspaulb: We need to add here some documentation of our new config parameter for whitelisting some URL's, plus an example of for which URL's this could be used (eg: auth_saml)

auth_pass_ip_restriction/models/res_users.py
```diff
@@ -0,0 +1,22 @@
...
+from odoo.exceptions import AccessDenied
+from odoo.http import request
+from odoo.tools.config import config
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    def _check_credentials(self, password, env):
+        ignored_endpoints = config.get("auth_pass_ignored_endpoints")
```
thomaspaulb, line 11: In particular I would like to know, as a README reader, in what way I should add these multiple endpoints in one configuration line. Are they comma-separated, or Python list-syntax, or...

thomaspaulb (changes requested):

thomaspaulb (approved):

NL66278:
@thomaspaulb A question on how / when to use this. I guess it is only effective if Odoo is connected directly to the internet and listening on all ports, that is, it is not behind a nginx or apache reverse proxy. Is that correct? And if correct, should we advise such a configuration, as basically users will be logging in over http, with their passwords in clear text on the local network. But maybe I misunderstand the usecase.

ddejong-therp:
A talked with @NL66278 to clarify the issue, but I think what he was concerned with was that the IP address that is tested against, would be the IP address of where the reverse proxy is running. But the IP address to be tested is (eventually) obtained from the WSGI variable `REMOTE_ADDR`, which I believe is set to be the web client's ip address. But I guess that may depend on your setup.

ddejong-therp:
I'll update the readme to clarify where the IP address is tested against.

PR #491, opened 2023-03-03
[15.0][FIX] password_security: constraint of password_estimate range

astirpe:
This fixes the computation of constraint of password_estimate range. Without this fix, the constraint never raises an error.
Backported from https://github.com/OCA/server-auth/pull/482


amh-mw (approved):
LGTM @OCA/tools-maintainers

thomaspaulb (approved):

Issue #54, opened 2018-11-08
[11] Password Security - Reset password


MathAndrews:
Hi,

I install the password_security, and I'm not sure that the test for the "resets inside of min" is working.
I'm on Odoo 11, and the reset password form send in "qcontext" a empty string for the token as so the first 'if' is never valid. (controller/main.py line 61)
Plus, if this 'if' is valid the queries "user_ids = request.env.sudo().search(...)" (line 74 and 79) are not valid and throw an exception "AttributeError: 'Environment' object has no attribute 'sudo'" and should be replace by "user_ids = request.env['res.users'].sudo().search(...)"

Best regards,

M.

pedrobaeza:
Can you express some steps for reproducing a problem?

MathAndrews:
Hi,

1. Token in qcontext variable:
- add logs to display _qcontext_ variable (controller/main.py line 66)
- go to "/web/reset_password" page
- submit the form
- in the logs you will see something like: "OrderedDict([('login', 'math@aaa.com'), ('redirect', ''), ('token', ''), ('reset_password_enabled', True), ('signup_enabled', True)])"

So the condition _'token' not in qcontext_ in the is statement is always False

2. Exception "AttributeError: 'Environment' object has no attribute 'sudo'"
- make sure the IF statement is True (controller/main.py line 67)
- go to "/web/reset_password" page
- submit the form
- you will get the Exception

pedrobaeza:
But I still don't see the problem in the standard module. Are you adding some custom code or something?

MathAndrews:
For me the issue is as the test **'token' not in qcontext** is always **False**, so it will not do the check done by **user_ids._validate_pass_reset()**

I try on a brand new database, and install the module "auth_signup", "base_setup", "contacts", "website", and "password_security"

MathAndrews:
Hi,
Any news about this? Do I misunderstand something?

raishak6:
I installed the module on Version 10 and it worked fine. However, I noticed that after password expires, based on the expiration period specified, no user (including the Administrator) would be able to login again and no one (including the Administrator) is redirected to the password reset page. If the Admin can be excluded from the password constrain and the users are redirected to the password reset page, then all will be fine. Otherwise, everyone, including the Admin might no longer have access after password expiration. Is there a way to resolve this?

evaldezb:
## I'm having the same (or similar) issue.

If the user password has expired:
```python
days = (today - write_date).days
[...]

PR #492, opened 2023-03-03
[15.0][FIX] password_security: Reset Password issue

astirpe:
Fixes the Reset Password issue described in https://github.com/OCA/server-auth/issues/54.
The fix is backported from https://github.com/OCA/server-auth/pull/482.
Replaces mocked tests for password reset, as they were not capable to test the code properly.

### DIFF
--- a/password_security/controllers/main.py
+++ b/password_security/controllers/main.py
@@ -58,29 +58,3 @@ class PasswordSecurityHome(AuthSignupHome):
             qcontext = self.get_auth_signup_qcontext()
             qcontext["error"] = str(e)
             return request.render("auth_signup.signup", qcontext)
-
-    @http.route()
-    def web_auth_reset_password(self, *args, **kw):
-        """It provides hook to disallow front-facing resets inside of min
-        Unfortuantely had to reimplement some core logic here because of
-        nested logic in parent
-        """
-        qcontext = self.get_auth_signup_qcontext()
-        if (
-            request.httprequest.method == "POST"
-            and qcontext.get("login")
-            and "error" not in qcontext
-            and "token" not in qcontext
-        ):
-            login = qcontext.get("login")
-            user_ids = request.env.sudo().search(
-                [("login", "=", login)],
-                limit=1,
-            )
-            if not user_ids:
-                user_ids = request.env.sudo().search(
-                    [("email", "=", login)],
-                    limit=1,
-                )
-            user_ids._validate_pass_reset()
-        return super(PasswordSecurityHome, self).web_auth_reset_password(*args, **kw)
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -164,11 +164,11 @@ class ResUsers(models.Model):
         :raises: UserError on invalidated pass reset attempt
         :return: True on allowed reset
         """
-        for rec_id in self:
-            pass_min = rec_id.company_id.password_minimum
+        for user in self:
+            pass_min = user.company_id.password_minimum
             if pass_min <= 0:
-                pass
-            write_date = rec_id.password_write_date
+                continue
+            write_date = user.password_write_date
             delta = timedelta(hours=pass_min)
             if write_date + delta > datetime.now():
                 raise UserError(
@@ -205,3 +205,13 @@ class ResUsers(models.Model):
 
         self.write({"password_history_ids": [(0, 0, {"password_crypt": pw})]})
         return res
+
+    def action_reset_password(self):
+        """Disallow password resets inside of Minimum Hours"""
+        if not self.env.context.get("install_mode") and not self.env.context.get(
+            "create_user"
+        ):
+            if not self.env.user._is_admin():
+                users = self.filtered(lambda user: user.active)
+                users._validate_pass_reset()
+        return super().action_reset_password()
--- a/password_security/tests/test_reset_password.py
+++ b/password_security/tests/test_reset_password.py
@@ -1,0 +1,87 @@
+# Copyright 2023 Onestein (<https://www.onestein.eu>)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import http
+from odoo.exceptions import UserError
+from odoo.tests.common import HttpCase, Opener, get_db_name, new_test_user, tagged
+
+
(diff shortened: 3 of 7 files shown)


password_security/controllers/main.py
```diff
@@ -70,17 +70,32 @@ def web_auth_reset_password(self, *args, **kw):
...
+                        limit=1,
+                    )
+                )
+            try:
+                user._validate_pass_reset()
+            except UserError as e:
+                qcontext["error"] = e.args[0]
+                response = request.render("auth_signup.reset_password", qcontext)
+                response.headers["X-Frame-Options"] = "DENY"
+                return response
```
amh-mw: What if the call to _validate_pass_reset moved into `res.users` action_reset_password instead? That would remove the need for this copy pasta.

password_security/models/res_users.py
```diff
@@ -164,11 +164,11 @@ def _validate_pass_reset(self):
         :raises: UserError on invalidated pass reset attempt
         :return: True on allowed reset
         """
-        for rec_id in self:
-            pass_min = rec_id.company_id.password_minimum
+        for user in self:
+            pass_min = user.company_id.password_minimum
             if pass_min <= 0:
                 pass
```
amh-mw: Shouldn't this be `continue`, not `pass`?

amh-mw (changes requested):

password_security/controllers/main.py
amh-mw: It might be necessary to check to see if the `request.uid` is the admin user, to allow for bulk password resets that ignore this setting.

password_security/models/res_users.py
astirpe: Right!

password_security/controllers/main.py
astirpe: In commit https://github.com/OCA/server-auth/pull/492/commits/8a5ead9447324199017fef6b984ba17302bcc2e6 I moved `_validate_pass_reset` into `action_reset_password`. It seems working fine to me. Would you also give it a look?

password_security/models/res_users.py
```diff
@@ -205,3 +205,12 @@ def _set_encrypted_password(self, uid, pw):
 
         self.write({"password_history_ids": [(0, 0, {"password_crypt": pw})]})
         return res
+
+    def action_reset_password(self):
+        """Disallow password resets inside of Minimum Hours"""
+        if self.env.context.get("install_mode"):
+            return
+        if not self.env.context.get("create_user"):
+            users = self.filtered(lambda user: user.active)
```
amh-mw: This active user check seems redundant with

https://github.com/odoo/odoo/blob/991d1c430f2c9bb0d2b2d7ac8f75338eafd2aced/addons/auth_signup/models/res_users.py#L166-L167

but maybe it is handling a case where a module could create an inactive user?

Good catch on the install_mode check. Maybe also add something like

```
if self.env.uid == SUPERUSER_ID:
```

as I just got `Passwords can only be reset every 24 hour(s). Please contact an administrator for assistance.` while logged in as admin. 😆 

Edit: Steps to reproduce as admin on a brand new instance with `--init=password_security`:

1. Home Menu > Settings
2. Users & Companies > Users
3. Check Administrator
4. Action > Send Password Reset Instructions
5. Error

amh-mw (changes requested):

password_security/models/res_users.py
astirpe: The check on active is redundant but I made it on purpose, so that in case an user is inactive, [this standard error](https://github.com/odoo/odoo/blob/991d1c430f2c9bb0d2b2d7ac8f75338eafd2aced/addons/auth_signup/models/res_users.py#L166-L167) will be always raised, instead of the eventual `_validate_pass_reset`.

I just added the SUPERUSER_ID check.

password_security/tests/test_reset_password.py
```diff
@@ -0,0 +1,74 @@
...
+        super().setUp()
+
+        # Create user with strong password: no error raised
+        new_test_user(self.env, "jackoneill", password="!asdQWE12345_3")
+
+    def reset_password(self, username):
+        """Reset user password"""
+        db = get_db_name()
+        if getattr(self, "session", None):
+            http.root.session_store.delete(self.session)
```
amh-mw: Delete this to appease codecov?

password_security/models/res_users.py
```diff
@@ -205,3 +205,13 @@ def _set_encrypted_password(self, uid, pw):
 
         self.write({"password_history_ids": [(0, 0, {"password_crypt": pw})]})
         return res
+
+    def action_reset_password(self):
+        """Disallow password resets inside of Minimum Hours"""
+        if self.env.context.get("install_mode"):
+            return
```
amh-mw: Maybe negate this and indent the following lines to remove the uncovered return? Falling through to the parent is fine, because it has its own install_mode checks.

amh-mw (changes requested):

password_security/models/res_users.py
```diff
@@ -205,3 +205,13 @@ def _set_encrypted_password(self, uid, pw):
 
         self.write({"password_history_ids": [(0, 0, {"password_crypt": pw})]})
         return res
+
+    def action_reset_password(self):
+        """Disallow password resets inside of Minimum Hours"""
+        if not self.env.context.get("install_mode") and not self.env.context.get(
+            "create_user"
+        ):
+            if self.env.uid != SUPERUSER_ID:
```
amh-mw: Whoops, this doesn't work, due to my bad suggestion. `SUPERUSER_ID` is the true root user, when really I should have said `self.env.ref('base.user_admin').id` for the admin user.

amh-mw (changes requested):
I'm not sure if there's anything to be done about the `codecov/project` red build. Yes, the lines of code covered have decreased, but that's because the new code is fewer lines of code...

password_security/models/res_users.py
amh-mw: Grepping the codebase a bit, I think `self.env.is_admin()` encapsulates this a bit better.

amh-mw (approved):
LGTM @OCA/tools-maintainers

thomaspaulb (approved):

PR #493, opened 2023-03-03
Add auth_oauth_ropc

adrienpeiffer:
This module add the possibility to login with OAuth Resource Owner Password Credentials Grant

https://datatracker.ietf.org/doc/html/rfc6749#section-4.3 

In most scenarios, more secure alternatives are available and recommended. This flow requires a very high degree of trust in the application, and carries risks that are not present in other flows. You should only use this flow when other more secure flows aren't viable.

This module is usefull for the Odoo mobile application, which only supports user/password authentication.


auth_azure_ropc/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module add the possibility to login with Microsoft Azure ROPC flow
+
+https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth-ropc
+
+Microsoft recommends you do not use the ROPC flow. In most scenarios, more secure alternatives are available and recommended. This flow requires a very high degree of trust in the application, and carries risks that are not present in other flows. You should only use this flow when other more secure flows aren't viable.
```
sbidoul: Actually, I think this is not specific to Azure. This looks like oauth2 Resource [Owner Password Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.3) (aka Direct Access Grant in keycloak).

So you could perhaps rename it to `auth_oauth_ropc`, and give the Azure config as an example?

Also it might be worth mentioning why it is necessary in the README (i.e. the Odoo mobile app as you explained to me :)

adrienpeiffer: @sbidoul Done

auth_oauth_ropc/models/res_users.py
```diff
@@ -0,0 +1,24 @@
...
+        try:
+            return super(ResUsers, self)._check_credentials(password, env)
+        except AccessDenied:
+            passwd_allowed = (
+                env["interactive"] or not self.env.user._rpc_api_keys_only()
+            )
+            if passwd_allowed and self.env.user.active:
+                ropc_providers = self.env["oauth.ropc.provider"].sudo().search([])
+                for conf in ropc_providers:
+                    if conf._authenticate(self.env.user.login, password):
```
sbidoul: Sending the user/password to several providers is going to leak the credential to providers that should not receive them? I tend to think we should allow only one?

auth_oauth_ropc/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module add the possibility to login with OAuth Resource Owner Password Credentials Grant
+
+https://datatracker.ietf.org/doc/html/rfc6749#section-4.3
+
+In most scenarios, more secure alternatives are available and recommended. This flow requires a very high degree of trust in the application, and carries risks that are not present in other flows. You should only use this flow when other more secure flows aren't viable.
+
+This module is usefull for the Odoo mobile application, which only supports user/password authentication.
```
sbidoul:
```suggestion
This module is useful for the Odoo mobile application, which only supports user/password authentication.
```

auth_oauth_ropc/models/res_users.py
adrienpeiffer: Good catch

adrienpeiffer:
@sbidoul Done

sbidoul (approved):

adrienpeiffer:
@OCA/server-environment-maintainers Could you reopen this one please ?

adrienpeiffer:
@sbidoul @lmignon

PR #495, opened 2023-03-08
[15.0][IMP] password_security: Speed up tests.

amh-mw:
Switched from TransactionCase to SavepointCase.
Removed sleeps.
Test runtime reduced by about six seconds.

### DIFF
--- a/password_security/tests/test_res_users.py
+++ b/password_security/tests/test_res_users.py
@@ -1,50 +1,48 @@
 # Copyright 2015 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
-import time
-
 from odoo.exceptions import UserError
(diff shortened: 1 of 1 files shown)


rvalyi (approved):
LGTM

pedrobaeza:
The sleep was introduce to avoid problems sometimes when the mark is the same on the several checks. Please do git blame to see why it was introduced.

amh-mw:
> The sleep was introduce to avoid problems sometimes when the mark is the same on the several checks. Please do git blame to see why it was introduced.

I surmised that the tests often created the user, then wrote `password_write_date` within the same second boundary, so I manually set `password_write_date` to the Unix epoch. Is there something else I'm missing?

pedrobaeza (approved):
OK then :)

/ocabot merge patch

PR #500, opened 2023-03-20
[IMP] password_security: add flag to enable/disable password security policy

ajaniszewska-dev:
image
image

### DIFF
--- a/password_security/models/res_company.py
+++ b/password_security/models/res_company.py
@@ -9,6 +9,7 @@ from odoo.exceptions import ValidationError
 class ResCompany(models.Model):
     _inherit = "res.company"
 
+    password_policy_enabled = fields.Boolean(default=False)
     password_expiration = fields.Integer(
         "Days",
         default=60,
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -55,6 +55,8 @@ class ResUsers(models.Model):
     def get_password_policy(self):
         data = super(ResUsers, self).get_password_policy()
         company_id = self.env.user.company_id
+        if not company_id.password_policy_enabled:
+            return data
         data.update(
             {
                 "password_lower": company_id.password_lower,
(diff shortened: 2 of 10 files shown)


password_security/models/res_users.py
```diff
@@ -55,16 +55,17 @@ def write(self, vals):
...
+                    "password_upper": company_id.password_upper,
+                    "password_numeric": company_id.password_numeric,
+                    "password_special": company_id.password_special,
+                    "password_length": company_id.password_length,
+                    "password_estimate": company_id.password_estimate,
+                }
+            )
         return data
 
     def _check_password_policy(self, passwords):
```
sebalix, line 72: This std method should check the new flag as well

sebalix:
NOTE: this flag is required do not get unrelated tests failing when a user is created with a weak password.
In general I think it's better to propose an option to enable a feature if this one could break existing features/flows.

password_security/models/res_users.py
thomaspaulb: @sebalix This comment seems to be open still?

sebalix: Hello @thomaspaulb it is now fixed

sebalix:
@ajaniszewska-dev @thomaspaulb I rebased the PR and rework a bit its content to reduce the diff to its minimum, ready for another review

sebalix (approved):

password_security/models/res_company.py
```diff
@@ -9,6 +9,7 @@
 class ResCompany(models.Model):
     _inherit = "res.company"
 
+    password_policy_enabled = fields.Boolean(default=False)
```
simahawk, line 12: I guess whoever installed this module already, expects it to be on by default, right?
If yes, we need a migration script to set the flag to True on upgrade.
N'est pas?

simahawk (commented):
Migration script probably needed.

password_security/models/res_company.py
sebalix: You are right! I pushed a migration script (tested locally on a customer project).

simahawk (approved):

PR #501, opened 2023-03-22
[14.0][IMP] auth_api_key: archive key when user is archived

TDu:
An archived user should not have an active api key anymore. But to stay backward compatible the migration script will keep all key active.

This is to follow up on :
* https://github.com/OCA/server-auth/pull/396

ref.: cos-3714

### DIFF
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -23,9 +23,11 @@ class AuthApiKey(models.Model):
         help="""The user used to process the requests authenticated by
         the api key""",
     )
-    # Not using related to stay backward compatible with having active key
-    # for an archived user (no need to be charged on active user for the api)
-    active = fields.Boolean(compute="_compute_active", readonly=False, store=True)
+    # Not using related to stay backward compatible with having active keys
+    # for archived users (no need being invoiced by Odoo for api request users)
+    active = fields.Boolean(
+        compute="_compute_active", readonly=False, store=True, default=True
+    )
 
     _sql_constraints = [("name_uniq", "unique(name)", "Api Key name must be unique.")]
 
@@ -52,10 +54,16 @@ class AuthApiKey(models.Model):
         self._retrieve_api_key_id.clear_cache(self.env[self._name])
         self._retrieve_uid_from_api_key.clear_cache(self.env[self._name])
 
-    @api.depends("user_id.active")
+    @api.depends(
+        "user_id.active", "user_id.company_id.archived_user_disable_auth_api_key"
+    )
     def _compute_active(self):
+        option_disable_key = self.user_id.company_id.archived_user_disable_auth_api_key
         for record in self:
-            record.active = record.user_id.active
+            if option_disable_key:
+                record.active = record.user_id.active
+            # To stay coherent if the option is disabled the active field is not
+            # changed. Because the field is stored, it should not be an issue.
 
     @api.model
     def create(self, vals):
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -23,6 +23,9 @@ class AuthApiKey(models.Model):
         help="""The user used to process the requests authenticated by
         the api key""",
     )
+    # Not using related to stay backward compatible with having active key
+    # for an archived user (no need to be charged on active user for the api)
+    active = fields.Boolean(compute="_compute_active", readonly=False, store=True)
 
     _sql_constraints = [("name_uniq", "unique(name)", "Api Key name must be unique.")]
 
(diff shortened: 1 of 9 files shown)


auth_api_key/models/auth_api_key.py
```diff
@@ -49,6 +52,11 @@ def _clear_key_cache(self):
         self._retrieve_api_key_id.clear_cache(self.env[self._name])
         self._retrieve_uid_from_api_key.clear_cache(self.env[self._name])
 
+    @api.depends("user_id.active")
+    def _compute_active(self):
+        for record in self:
+            record.active = record.user_id.active
```
simahawk: Wondering: shall we make this dependent on a company setting?
Although the manual action on the form should ease manual override...

auth_api_key/tests/test_auth_api_key.py
```diff
@@ -43,3 +43,12 @@ def test_cache_invalidation(self):
         )
         with self.assertRaises(ValidationError):
             self.env["auth.api.key"]._retrieve_uid_from_api_key("api_key")
+
+    def test_user_archived_unarchived(self):
+        demo_user = self.env.ref("base.user_demo")
+        self.env["auth.api.key"]._retrieve_uid_from_api_key("api_key"), demo_user.id
```
simahawk: can you use an assertion here?

simahawk (commented):
@lmignon @sebastienbeau ping

lmignon:
Be careful. We've UC where we use an archived user to process api calls. This should still be possible.

simahawk:
> Be careful. We've UC where we use an archived user to process api calls. This should still be possible.

The upgrade step sets the flag as active by default. Shouldn't be a problem in your case. 
However, I tend to say that making this behavior explicit and controlled via company setting is probably better.

auth_api_key/tests/test_auth_api_key.py
TDu: done

TDu:
I have added a setting that when unchecked (the default) the active field of api key records should never be changed.

auth_api_key/models/auth_api_key.py
```diff
@@ -23,6 +23,11 @@ class AuthApiKey(models.Model):
         help="""The user used to process the requests authenticated by
         the api key""",
     )
+    # Not using related to stay backward compatible with having active key
+    # for an archived user (no need to be charged on active user for the api)
```
simahawk:
```suggestion
    # for an archived user (no need to be changed on active user for api requests)
```

auth_api_key/models/auth_api_key.py
```diff
@@ -49,6 +54,15 @@ def _clear_key_cache(self):
         self._retrieve_api_key_id.clear_cache(self.env[self._name])
         self._retrieve_uid_from_api_key.clear_cache(self.env[self._name])
 
+    @api.depends("user_id.active")
+    def _compute_active(self):
+        option_disable_key = self.env.company.archived_user_disabled_key
```
simahawk: `self.user_id.company_id` maybe?

auth_api_key/models/res_company.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    archived_user_disabled_key = fields.Boolean(
```
simahawk:
```suggestion
    archived_user_disable_api_key = fields.Boolean(
```

auth_api_key/models/auth_api_key.py
simahawk:
```suggestion
        option_disable_key = self.env.company.archived_user_disable_api_key
```

auth_api_key/models/auth_api_key.py
```diff
@@ -49,6 +54,15 @@ def _clear_key_cache(self):
...
+        for record in self:
+            if option_disable_key:
+                record.active = record.user_id.active
```
simahawk, line 64:
```suggestion
            record.active = record.user_id.active if option_disable_key else True
```

auth_api_key/models/auth_api_key.py
```diff
@@ -49,6 +54,15 @@ def _clear_key_cache(self):
         self._retrieve_api_key_id.clear_cache(self.env[self._name])
         self._retrieve_uid_from_api_key.clear_cache(self.env[self._name])
 
+    @api.depends("user_id.active")
```
simahawk:
```suggestion
    @api.depends("user_id.active", "user_id.company_id.archived_user_disabled_key")
```

auth_api_key/models/res_company.py
lmignon: /s/archived_user_disabled_key/archived_user_disabled_auth_api_key

simahawk (approved):
LG, could you squash pls?

@lmignon fine for you?

PR #503, opened 2023-04-12
[IMP] auth_saml nicer handling of logout and autoredirect

gurneyalex:
When autoredirect is enabled, but some technical accounts have a normal
password authentication setup, it can be difficult to access them as
logging out of a session will trigger the autoredirect feature.

We change this by overriding the logout redirection so that it sets the
disable_autoredirect parameter -> this way the user will get redirected
to the normal signin screen.

### DIFF
--- a/auth_saml/controllers/main.py
+++ b/auth_saml/controllers/main.py
@@ -15,6 +15,7 @@ from odoo.http import request
 
 from odoo.addons.web.controllers.main import (
     Home,
+    Session as WebSession,
     ensure_db,
     login_and_redirect,
     set_cookie_and_redirect,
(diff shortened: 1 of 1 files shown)


auth_saml/controllers/main.py
```diff
@@ -270,3 +271,20 @@ def saml_metadata(self, req, **kw):
                 ),
                 [("Content-Type", "text/xml")],
             )
+
+
+class Session(WebSession):
+    @http.route("/web/session/logout", type="http", auth="none")
+    def logout(self, redirect="/web/login"):
```
gurneyalex, line 278: I'm wondering if I should not just say 
```suggestion
    def logout(self, redirect="/web/login?disable_autoredirect="):
```

and simply  `return super().logout(redirect)`

yankinmax: hello @gurneyalex , reading the doc https://github.com/OCA/server-auth/blob/15.0/auth_saml/README.rst#configuration I think we could try to test it and have an answer

yankinmax (approved):

ivantodorovich (approved):

auth_saml/controllers/main.py
yankinmax: I've tried this variant, seems we have the same expected result

PR #507, opened 2023-04-20
[16.0][MIG] auth_session_timeout

bosd:
Standard migration using odoo-module-migrator /ocabot migration auth_session_timeout .


gurneyalex (approved):

Kaliuta:
v 16.0
A message appears after a set time:
"Your Odoo session expired. The current page is about to be refreshed. "

When I click "OK", the page reloads, but I remain logged in.

bosd:
> A message appears after a set time: "Your Odoo session expired. The current page is about to be refreshed. "
> 
> When I click "OK", the page reloads, but I remain logged in.

I can replicate this behaviour.. 
Any ideas why it is not working?

auth_session_timeout/models/res_users.py
```diff
@@ -0,0 +1,103 @@
...
+                expired = True
+
+        # Try to terminate the session
+        terminated = False
+        if expired:
+            terminated = self._auth_timeout_session_terminate(session)
+
+        # If session terminated, all done
+        if terminated:
+            raise SessionExpiredException("Session expired")
```
Leonidas-VII: Hi @bosd , 
> I can replicate this behaviour..
> Any ideas why it is not working?

I see this line making the issue
Because you raise an exception here, then everything is reverted, the user can not log out

For solution, we should `return` here, not `raise`

CoraVanga: @bosd please help to check, thank you

bosd: > For solution, we should return here, not raise

I'm having some personal time. Later, in a couple of weeks I will get back to this.

bosd: > For solution, we should `return` here, not `raise`

@Leonidas-VII  Thanks for the suggestion, I made the commit. It does not seem to resolve the problem.

I have the impression it is related to these lines:

https://github.com/OCA/server-auth/pull/507/files#diff-1648a6bceb7d2da7ef9fc6326f0f894ee39f11f613d80c83ae53849fdb5698eaR51-R52

bosd:
It is ok now, please review

auth_session_timeout/models/res_users.py
thomaspaulb: For `return`, we can just return anything here, nothing is done with the result. So we don't need to return the exception.

Meanwhile, the issue of a refresh not leading to an expiration I think is caused by [this](https://github.com/OCA/server-auth/pull/535) - can you cherry-pick it in?

zuher83:
I think everything is ready now

auth_session_timeout/models/res_users.py
anhvu-sg: just a question: why don't we add the auto-reload function after 5/10 seconds instead of clicking on the button "OK"?

bosd:
> I think everything is ready now

I still did not include the cherry pick. Will do it later this week.

thomaspaulb:
When you do so, could you also check if [this](https://github.com/OCA/server-auth/pull/535) should be applied for 16.0 too?

bosd:
Excuse me for the delay. Cherry picks are now included.

thomaspaulb (approved):

thomaspaulb:
Code approved, but it's probably good if someone still tests this on runbot once more to be sure. @KKamaa maybe you can do a sanity check here since you've been busy with the module

auth_session_timeout/data/ir_config_parameter_data.xml
```diff
@@ -0,0 +1,14 @@
...
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+<odoo noupdate="1">
+    <record id="inactive_session_time_out_delay" model="ir.config_parameter">
+        <field name="key">inactive_session_time_out_delay</field>
+        <field name="value">7200</field>
+    </record>
+    <record id="inactive_session_time_out_ignored_url" model="ir.config_parameter">
+        <field name="key">inactive_session_time_out_ignored_url</field>
+        <field name="value">/calendar/notify,/longpolling/poll</field>
```
celm1990, line 12: I think this URL must be update to V16 for `web socket` instead of `longpolling`, what do you think?
Please see
https://github.com/odoo/odoo/blob/16.0/addons/bus/controllers/websocket.py#L13

bosd: Good Catch!
Will update the PR.
Related: https://github.com/odoo/odoo/pull/75510
https://github.com/odoo/odoo/commit/de6de48deb983be910876c6133c2869907c142f7

max3903 (approved):

SMaciasOSI (approved):

bosd:
@CasVissers-360ERP It's Merged

PR #511, opened 2023-05-10
[13.0][CHG] auth_oauth_autologin: redirect to the oauth in JS to keep track of the fragment

benwillig:
At the moment, if the user click on an Odoo mail notification message, he expects to be redirected on the record form view,. Using the autologin method, he will be redirected to the odoo home page instead after a successful login. This is due to the lost fragment of the original URI.
To correct this behavior, the redirect is done in javascript. Sending the original URI with its fragment to get the correct redirect URI.

Unfortunately, the login page has to be rendered first.


sbidoul (approved):

sbidoul:
@benwillig can you confirm this works in production?

benwillig:
@sbidoul Yes it's working in production.

sbidoul:
Thanks! 

/ocabot merge minor

PR #513, opened 2023-05-12
[16.0][MIG] auth_oauth_multi_token

mlaitinen:
- Applied English capitalization rules in field names
- Improved the `res.users` form by horizontally "unsqueezing" the list of auth tokens


mlaitinen:
I didn't touch `README.rst` because IIRC it gets automatically generated once the PR is merged. Correct me if I'm wrong.

auth_oauth_multi_token/models/auth_oauth_multi_token.py
```diff
@@ -10,7 +10,7 @@ class AuthOauthMultiToken(models.Model):
     """Define a set of tokens."""
 
     _name = "auth.oauth.multi.token"
-    _description = "OAuth2 token"
+    _description = "OAuth2 Token"
```
pedrobaeza, line 13: Don't change existing capitalization. It's not incorrect as it was. Same comment for the rest of this kind of changes.

pedrobaeza (changes requested):
The README is regenerated on merge, yes.

auth_oauth_multi_token/models/auth_oauth_multi_token.py
mlaitinen: Is this an OCA policy? If yes, could you please send me a link? Only remark about capitalization I could find is here: [CONTRIBUTING.rst](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#fields).

This is the way how titles in English are capitalized, and also how the names of all models & fields in Odoo are formatted.

PR #514, opened 2023-05-23
Fix web.base.url

sefirosweb:
Now is getting the current url of the request instead of using an environment variable of "web.base.url"

This causes an error when odoo is behind the proxy, and the proxy uses https, odoo thinks it is using http instead of https because it is behind the proxy, odoo sends to http instead https to serivce provider and cause fails auth

With this change now uses "web.base.url" correctly, and you can add manually your full domain including https

Note: maybe must be update to odoo 14 15 16

### DIFF
--- a/auth_saml/controllers/main.py
+++ b/auth_saml/controllers/main.py
@@ -170,7 +170,7 @@ class AuthSAMLController(http.Controller):
 
         provider = request.env["auth.saml.provider"].sudo().browse(provider_id)
         redirect_url = provider._get_auth_request(
-            self._get_saml_extra_relaystate(), request.httprequest.url_root.rstrip("/")
+            self._get_saml_extra_relaystate()
         )
         if not redirect_url:
             raise Exception(


sefirosweb:
I have another issue related with don't match web.base.url...

```
web_1  | 2023-05-23 21:24:06,046 1 ERROR odoo_db saml2.response: destination 'https://odoo.mydomain.com/auth_saml/signin' not in return addresses '['http://odoo.mydomain.com/auth_saml/signin', 'http://odoo.mydomain.com/auth_saml/signin']' 
web_1  | 2023-05-23 21:24:06,047 1 ERROR odoo_db odoo.addons.auth_saml.controllers.main: SAML2: failure - 'NoneType' object has no attribute 'text' 
web_1  | Traceback (most recent call last):
web_1  |   File "/mnt/extra-addons/auth_saml/controllers/main.py", line 217, in signin
web_1  |     env["res.users"]
web_1  |   File "/mnt/extra-addons/auth_saml/models/res_users.py", line 64, in auth_saml
web_1  |     validation = self._auth_saml_validate(provider, saml_response, base_url)
web_1  |   File "/mnt/extra-addons/auth_saml/models/res_users.py", line 29, in _auth_saml_validate
web_1  |     return provider._validate_auth_response(token, base_url)
web_1  |   File "/mnt/extra-addons/auth_saml/models/auth_saml_provider.py", line 293, in _validate_auth_response
web_1  |     matching_value = response.name_id.text
web_1  | AttributeError: 'NoneType' object has no attribute 'text'
```

I create new PR when i fix this

Issue #515, opened 2023-05-23
auth_saml not working behind proxy with https


sefirosweb:
## Module
auth_saml

## Describe the bug
When you setup odoo with proxy reversed to add https the module auth_saml don't works propelly,
It is expected that if you configure the web.base.url parameter it should take this into account when redirecting and creating the route for "XXX/auth_saml/signin", but this is not the case.

What it does is use the route of the sent request, and the problem is due to the proxy, since the client requests the proxy through https, and then the proxy connects to odoo at http, and odoo interprets that it is http and not https ignoring web.base.url

## To Reproduce
**Affected versions**: Checked on 16 but i'm sure happens on 14 + 15

- Do standard installatión with http
- Install ngrok, this provider you a "public" url with HTTPS (https://ngrok.com/)
- Now configure your service provider with this public url
- DONT use Override Base URL,

Now when you try to login you must have this errors:
![image](https://github.com/OCA/server-auth/assets/20754836/0d6eeb0b-416e-49c2-85ae-e422bac4292b)
![image](https://github.com/OCA/server-auth/assets/20754836/737b7cf7-fa0a-4073-b1da-fe13fbfc4a36)

jcolley:
I'll refer you to the Odoo documentation here: https://www.odoo.com/documentation/16.0/administration/install/deploy.html#https

You're looking for the `proxy_mode` configuration parameter.

sefirosweb:
I think I already tried this, but I don't remember it well, I'll try again, I'll let you know if it was this error

PR #516, opened 2023-05-23
Fix issue when have reverse proxy with https

sefirosweb:
With that changes I fixed the issue:
https://github.com/OCA/server-auth/issues/515


sefirosweb:
I dont understood the error of "codecov/project — 80.30% (-0.39%) compared to 302a3a8" anything to do here?

PR #520, opened 2023-06-07
[16.0] auth_jwt: add cookie mode

sbidoul:
Add a cookie mode on JWT validators. In this case, the JWT payload obtained
from the ``Authorization`` header is returned as a Http-Only cookie. This mode is
sometimes simpler for front-end applications which do not then need to store and protect
the JWT token across requests and can simply rely on the cookie management mechanisms of
browsers. When both the ``Authorization`` header and a cookie are provided, the cookie
is ignored in order to let clients authenticate with a different user by providing a new
JWT token.

### DIFF
--- a/auth_jwt/models/ir_http.py
+++ b/auth_jwt/models/ir_http.py
@@ -58,14 +58,6 @@ class IrHttpJwt(models.AbstractModel):
         return super()._authenticate(endpoint)
 
     @classmethod
-    def _get_jwt_cookie_secret(cls):
-        secret = request.env["ir.config_parameter"].sudo().get_param("database.secret")
-        if not secret:
-            _logger.error("database.secret system parameter is not set.")
-            raise ConfigurationError()
-        return secret
-
-    @classmethod
     def _get_jwt_payload(cls, validator):
         """Obtain and validate the JWT payload from the request authorization header or
         cookie."""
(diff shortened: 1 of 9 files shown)


sbidoul:
I keep this draft until the design of `auth_jwt_fastapi` in oca/rest-framework is ready, as we may need some refactoring.

hparfr (approved):
code read, didn't test

sbidoul:
This should now be ready.

hparfr (approved):

auth_jwt/models/ir_http.py
```diff
@@ -54,12 +52,25 @@ def _authenticate(cls, endpoint):
                 raise UnauthorizedSessionMismatch()
         return super()._authenticate(endpoint)
 
+    @classmethod
+    def _get_jwt_payload(cls, validator):
+        """Obtain and validate the JWT payload from the request authorization header or
+        cookie."""
+        try:
+            token = cls._get_bearer_token()
+            assert token
```
lmignon, line 61:
```suggestion
            assert token, "JWT token not found into request header nor as cookie"  
```

lmignon (approved):
Great. LGTM (Code review)

sbidoul:
Thanks for the review, @lmignon @hparfr. 

/ocabot merge minor

PR #521, opened 2023-06-11
[16.0][ADD] auth_oauth_signup

mlaitinen:
Enable sign up for users logging in using OAuth2 on a per-provider basis, overriding the global setting.

Without this module, your options are:
 1. allow sign up for everyone who can access the login page
 2. import users (and provide values for `oauth_provider_id` and `oauth_uid`) before they log in for the first time.


jaenbox:
In the event that the variable 'provider' has the value 'false,' the exception error is not caught, and the error is not displayed. 
You can fix it with this code. `if provider and provider.allow_signup:`

mlaitinen:
> In the event that the variable 'provider' has the value 'false,' the exception error is not caught, and the error is not displayed. You can fix it with this code. `if provider and provider.allow_signup:`

Well spotted, thanks. I fixed it and added a test to make sure that `SignupError` gets raised when sign up fails and `provider` is `False`.

jaenbox:
LGTM Functional Review

jaenbox (approved):

PR #522, opened 2023-06-14
[14.0][FIX] vault: version not set when migrate from 13.0

CarlosRoca13:
When we migrate from 13.0 the values of the version field take null value, this causes that when we introduce the user keys to see the vaults, the system throws the following error:
image

cc @Tecnativa TT43889

ping @pedrobaeza @fkantelberg

### DIFF
--- a/vault/migrations/14.0.1.8.0/post-migration.py
+++ b/vault/migrations/14.0.1.8.0/post-migration.py
@@ -1,0 +1,14 @@
+# Copyright 2023 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from openupgradelib import openupgrade
+
+
+def set_users_key_version(env):
+    openupgrade.logged_query(env.cr, "UPDATE res_users_key SET version = 1")
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    set_users_key_version(env)


fkantelberg:
It's difficult to say without knowing the 13.0 state. Normally `version < 1` means that the `session.username` is used to construct the key for decrypting the database. Currently 13.0 doesn't have the version => null/0 would be correct because username is used to derive the key. Reading the error message it also points to something different (`session` seems null)

CarlosRoca13:
On version 13.0 we don't have the field version https://github.com/OCA/server-auth/blob/dfb4e45030eb6df53a8e09e3104a4b4faaf89d72/vault/models/res_users_key.py#L21-L30

That's why I'm filling the valuewhen migrating. If it isn't filled we get the error that I have pasted on the description.

You can test it by going to an instance and via psql set the value of version on your active key to null.

fkantelberg:
I can't reproduce it like that and the test would be invalid because the correct value after a migration from 13.0 would be 0/null. In 13.0 I still see [this](https://github.com/OCA/server-auth/blob/dfb4e45030eb6df53a8e09e3104a4b4faaf89d72/vault/static/src/js/vault.js#L45). The same way the migration must be done [see here](https://github.com/OCA/server-auth/blob/14.0/vault/static/src/js/vault.js#L360). This is why the correct version would be 0 because the same decryption must be used. If you set version to null on a 14.0 this might be wrong because on 14.0 you can only generate keys with version 1 scheme.

If I read the error message correctly it's pointing to a different thing. Cannot read `username` of null inside of `_import_from_database` points to `session.username` and `session = null`. session comes from `var session = require("web.session");` I think you are looking on the wrong spot. Your fix only reverts your manual psql change but isn't a valid migration and it looks fixed to you because if version >= 1 the if branch isn't hit.

CarlosRoca13:
Thanks for the explanation and sorry for the inconvenience. I will then see how to solve the error for versions 0/null at js

CarlosRoca13:
Sorry, the problem is not on 14.0 it is only in 15.0

I will open new PR with the correct solution

Thank you for your attention

PR #530, opened 2023-06-22
[15.0][FIX] vault, vault_share: Error on test

CarlosRoca13:
With the changes introduced on https://github.com/odoo/odoo/commit/a251b14f017b35899663d30a1b38cb75a41cdb1b#diff-99de14b6e0a411718ac3adaaf779395155841c7ca9d95fa258369656dae63d18 the tests of vault and vault_share are failing.

cc @Tecnativa 

@fkantelberg please, can you review this. I'm not sure if this solution is the correct one, but we need to solve this.

Sorry for the inconvenience

### DIFF
--- a/vault/tests/test_controller.py
+++ b/vault/tests/test_controller.py
@@ -55,7 +55,7 @@ class TestController(TransactionCase):
             return context
 
         with MockRequest(self.env) as request_mock:
-            request_mock.render.side_effect = return_context
+            request_mock.render = return_context
             response = self.controller.vault_inbox("")
             self.assertIn("error", response)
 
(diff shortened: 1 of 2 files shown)


CarlosRoca13:
BTW... Why are we using `odoo.addons.website.tools -> MockRequest` if we have no dependency on website? https://github.com/OCA/server-auth/blob/4e970c0164565ac310754bb312db9c1497ade686/vault/__manifest__.py#L13

fkantelberg (approved):

pedrobaeza (approved):
/ocabot merge patch

fkantelberg:
@CarlosRoca13 Well yes there is no real dependency to website and I have not found a better way to mock it without copying the function nearly 1:1. Either way adding the dependency wouldn't be my favorite way. Suggestions are welcome

PR #533, opened 2023-06-26
[14.0] auth_api_key: allow to update api key from form view and hide …

petrus-v:
…keys in UI

While using with server `auth_api_key_server_env` and `server_environment_data_encryption` modules it makes hard to edit values for different environments.

Also I fill it's better practice to hide the key value on user interface so I've had the `password="True"` (I'm not considering this as security improvement but worth of it)


damdam-s (approved):

Koudja (approved):

petrus-v:
Dear @OCA/tools-maintainers,

Is it possible to get this PR re-open and merged? it get 2 review and approved label ?

Thanks for your time

PR #540, opened 2023-07-14
[FIX] vault: Vault re-encryption in a single transaction

fkantelberg:
This PR should solve #539 because it puts the storing of the newly encrypted values and master keys in a single transaction preventing the user from reloading/closing the JS before the process finished.

It also adds copy support for vault.entry because it was useful to reproduce the bug.

### DIFF
--- a/vault/controllers/main.py
+++ b/vault/controllers/main.py
@@ -126,3 +126,21 @@ class Controller(http.Controller):
 
             if isinstance(master_key, str):
                 right.sudo().key = master_key
+
+    @http.route("/vault/replace", auth="user", type="json")
+    def vault_replace(self, data):
+        """Replace the master keys and values within a single transaction"""
+        if not isinstance(data, list):
+            return
+
+        vault = request.env["vault"].with_context(vault_skip_log=True)
+        for changes in data:
+            record = vault.env[changes["model"]].browse(changes["id"])
+            vault |= record.vault_id
+            if record._name in ("vault.field", "vault.file"):
+                record.write({k: v for k, v in changes.items() if k in ["iv", "value"]})
+            elif record._name == "vault.right":
+                record.write({k: v for k, v in changes.items() if k in ["key"]})
+
+        for v in vault:
+            v._log_entry("Replaced the keys", "info")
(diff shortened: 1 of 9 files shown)


CarlosRoca13:
When you have not yet entered the key, and you try to delete a right, the interface is blocked and does not allow you to continue because it is not possible to enter the key.
blocked_vault

Also if there is corrupted data from the previous version, it will not be saved in the database because there will be null values, I think this change should be added: 
https://github.com/OCA/server-auth/pull/539/files#diff-f702debd4ca746a875148e54328f635aea387987af6e74f934251d94392539b4

fkantelberg:
@CarlosRoca13 I'm a bit conflicted with converting the nulls to empty strings. Yes it runs smoother when you do it and the user gets no exceptions but this can also be tricky because you never notice when entries are broken. In the end I want all entries to be valid at every time and eliminating the causes. If you want the change I'll push it.

I also try to add the missing unittests today.

CarlosRoca13:
I think the user should at least receive a more descriptive warning. Because if you have 600 entries in the same vault, you won't know where the exception comes from or if it is produced by more than one entry.

In addition to this, if an exception occurs when saving, the change of permissions is saved but the re-encryption is not saved. So I think we should add a button to launch the re-encryption manually and in the exception that indicates that entries are corrupted, suggest to launch it manually. What do you think about this approach?

fkantelberg:
Interresting approach to add the re-encrypt as extra feature. It could be simpler to do it like that. I have to think about it a bit. Maybe even 2 buttons to verify (check if everything is decryptable) and re-encrypt the vault.

fkantelberg:
I changed the following things:

- Button to verify the integrity of a vault with a helpful output
- Button to re-encrypt the vault (generate new master keys, decrypt and encrypt the fields and files)
- Deletion of a right only flags the vault (reencrypt_required and highlights the button)
- Added unittests for the controller and flagging

This helps identifying the problems for a recovery a backup and `vault-recovery` or an export + import is needed.

CarlosRoca13 (approved):

CarlosRoca13:
Thank you so much!

pedrobaeza (approved):
/ocabot merge major

PR #541, opened 2023-07-21
[BACKPORT][FIX][14.0] vault: Rework of re-encryption process

fkantelberg:
This is the backport of #540 to 14.0

PR #545, opened 2023-08-02
[15.0][IMP] vault: Add reload when finish the reencriptation to display updated data

CarlosRoca13:
Before this improvement when validate data after reencriptation, all the registries seems to be broken and the button does not recover it's initial state:
vaulr_reenc_verify1

After this change, the data is updated and the problems will not appear if they don't exists:
vaulr_reenc_verify2

cc @Tecnativa TT44183
ping @pedrobaeza @fkantelberg

### DIFF
--- a/vault/static/src/legacy/vault_controller.js
+++ b/vault/static/src/legacy/vault_controller.js
@@ -306,6 +306,7 @@ odoo.define("vault.controller", function (require) {
                         route: "/vault/replace",
                         params: {data: changes},
                     });
+                    this.reload();
                 }
             } finally {
                 framework.unblockUI();


fkantelberg (approved):
LGTM

vault/i18n/es.po
```diff
@@ -1347,6 +1347,8 @@ msgstr ""
 msgid ""
 "You will loose access to all vaults and your inbox. Do you want to continue?"
 msgstr ""
+"Perderás el acceso a todos tus vaults y a tu bandeja de entrada. ¿Desea "
```
pedrobaeza: Always use `usted` treatment:

```suggestion
"Perderá el acceso a todos sus vaults y a su bandeja de entrada. ¿Desea "
```

And this should be done on Weblate, not here, as there's chance of conflicting changes. Please discard them and translate on the proper platform.

CarlosRoca13:
Done @pedrobaeza

pedrobaeza (approved):
/ocabot merge patch

fkantelberg:
I cherry picked it for #541 and #542

PR #550, opened 2023-08-25
[IMP] added option for TOTP/2FA bypass for admin passkey

codeagencybe:
I have added a new option to bypass TOTP/2FA for admin passkey. 
In the odoo.conf one needs to add the new parameter as below. 
By default, this new option is always false, so it will not bypass TOTP/2FA if it has been set on a user login. 

New odoo.conf parameter:
`
auth_admin_passkey_ignore_totp = True
` 

I have tested this new feature in several production setups and it's working perfectly.

### DIFF
--- a/auth_admin_passkey/models/res_users.py
+++ b/auth_admin_passkey/models/res_users.py
@@ -8,6 +8,7 @@ from datetime import datetime
 
 from odoo import SUPERUSER_ID, _, api, exceptions, models
 from odoo.tools import config
+from odoo.http import request
 
 logger = logging.getLogger(__name__)
 
@@ -74,6 +75,12 @@ class ResUsers(models.Model):
                 password = hashlib.sha512(password.encode()).hexdigest()
 
             if password and file_password == password:
+                request.session['ignore_totp'] = config.get("auth_admin_passkey_ignore_totp", False)
                 self._send_email_passkey(users[0])
             else:
                 raise
+
+    def _mfa_url(self):
+        if request.session.get('ignore_totp'):
+            return None
+        return super()._mfa_url()
--- a/auth_admin_passkey/README.rst
+++ b/auth_admin_passkey/README.rst
@@ -51,6 +51,8 @@ following keys in your ``odoo.cfg`` configuration file.
 * ``auth_admin_passkey_password``. The password that allows user to logging in
   with any login. If not set, the feature is disabled.
 
+* ``auth_admin_passkey_ignore_totp`` (default False), if enabled, then 2FA will be ignored.
+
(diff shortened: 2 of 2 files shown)


legalsylvain:
Hi @codeagencybe could you take a look on red CI ?

bofiltd:
@codeagencybe Thanks for this contribution. Makes sense.
Can you please re-trigger tests?

bosd:
@codeagencybe Can you fix pre-commit?

astirpe:
@codeagencybe will you attend the comments above, to make pre-commit and tests green?

auth_admin_passkey/models/res_users.py
```diff
@@ -74,6 +75,12 @@ def _check_credentials(self, password, env):
                 password = hashlib.sha512(password.encode()).hexdigest()
 
             if password and file_password == password:
+                request.session['ignore_totp'] = config.get("auth_admin_passkey_ignore_totp", False)
                 self._send_email_passkey(users[0])
             else:
                 raise
+
+    def _mfa_url(self):
```
CasVissers-360ERP, line 83: Module should inherit auth_totp_mail_enforce to bypass.

codeagencybe:
Hello all. 

Apologies for late reply, seems like I don't receive any notifications from GitHub except from the last message from Cas. 

I will review everything shortly and update. 
Thanks!

auth_admin_passkey/models/res_users.py
```diff
@@ -74,6 +75,12 @@ def _check_credentials(self, password, env):
                 password = hashlib.sha512(password.encode()).hexdigest()
 
             if password and file_password == password:
+                request.session['ignore_totp'] = config.get("auth_admin_passkey_ignore_totp", False)
```
astirpe, line 78: @codeagencybe 

I ported your PR to V17, here: https://github.com/OCA/server-auth/pull/624

To make the existing tests passing with success, I added this change https://github.com/OCA/server-auth/pull/624/commits/f138da23c70db9dad8c0f095309857cc5e3a5b33 
This way we avoid that the session is being written if it's not existing.

I think you need to implement the same here as well.

astirpe (changes requested):

auth_admin_passkey/models/res_users.py
astirpe: As an alternative to inheriting `auth_totp_mail_enforce`, we could make use of a glue module to combine this PR and `auth_totp_mail_enforce`. Here is a proposal of a glue module for V17: https://github.com/OCA/server-auth/pull/625

PR #552, opened 2023-09-18
[FIX] auth_saml: decrypt matching_value response to compare output response.name_id.text

### LINKED ISSUE #553: [16.0] auth_saml: Error decrypt matching_value response to compare output response.name_id.text
Error trying to decrypt matching_value response to compare output response with response.name_id.text.

It shows AttributeError: 'NoneType' object has no attribute 'text'.

```
2023-09-17 02:02:54,015 4049 ERROR odoo odoo.addons.auth_saml.controllers.main: SAML2: failure - 'NoneType' object has no attribute 'text'
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/controllers/main.py", line 219, in signin
    .auth_saml(
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/models/res_users.py", line 64, in auth_saml
    validation = self._auth_saml_validate(provider, saml_response, base_url)
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/models/res_users.py", line 29, in _auth_saml_validate
    return provider._validate_auth_response(token, base_url)
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/models/auth_saml_provider.py", line 296, in _validate_auth_response
    matching_value = response.name_id.text
AttributeError: 'NoneType' object has no attribute 'text'
```

## Module
auth_saml

## Describe the bug

In the `class SecurityContext` in  `__init__` we need an `encryption_keypairs` to decrypt matching_value and compare with `response.name_id.text` on `[auth_saml_provider.py](https://github.com/OCA/server-auth/blob/16.0/auth_saml/models/auth_saml_provider.py#L209) according to the [saml2 library](https://github.com/IdentityPython/pysaml2/blob/master/tests/test_40_sigver.py#L126)

## To Reproduce
**Affected versions**:
16.0 - May affect 15.0

randall-vx:
Decrypt matching_value response to compare output response with response.name_id.text. It shows AttributeError: 'NoneType' object has no attribute 'text'.

```python
2023-09-17 02:02:54,015 4049 ERROR odoo odoo.addons.auth_saml.controllers.main: SAML2: failure - 'NoneType' object has no attribute 'text'
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/controllers/main.py", line 219, in signin
    .auth_saml(
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/models/res_users.py", line 64, in auth_saml
    validation = self._auth_saml_validate(provider, saml_response, base_url)
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/models/res_users.py", line 29, in _auth_saml_validate
    return provider._validate_auth_response(token, base_url)
  File "/home/odoo/instance/extra_addons/server-auth/auth_saml/models/auth_saml_provider.py", line 296, in _validate_auth_response
    matching_value = response.name_id.text
AttributeError: 'NoneType' object has no attribute 'text'
```

Check [class SecurityContext in saml2 lib](https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/sigver.py#L1186)

Fixes #553

### DIFF
--- a/auth_saml/models/auth_saml_provider.py
+++ b/auth_saml/models/auth_saml_provider.py
@@ -232,6 +232,9 @@ class AuthSamlProvider(models.Model):
             "cert_file": self._get_cert_key_path("sp_pem_public"),
             "key_file": self._get_cert_key_path("sp_pem_private"),
         }
+        settings["encryption_keypairs"] = [
+            {"key_file": settings["key_file"], "cert_file": settings["cert_file"]}
+        ]
         sp_config = Saml2Config()
         sp_config.load(settings)
         sp_config.allow_unknown_attributes = True


vincent-hatakeyama:
Hi,

There’s no issue opened on the subject so in which case do you get the message?

Can it be reproduced by adding a test as it is not raised in the current tests.

Regards.

randall-vx:
@vincent-hatakeyama What are you waiting for? Add a test where an error is shown without those parameters?

cc: @rolandojduartem

randall-vx:
@vincent-hatakeyama also coverage is increased.

cc: @rolandojduartem

vincent-hatakeyama:
> What are you waiting for? Add a test where an error is shown without those parameters?

That would be best if possible.
Otherwise, a simple scenario to reproduce the error would help. I didn’t see the error when testing with a local keycloak, but it might be linked with an option I did not test or something similar.

The version 15.0 of the module is in production without having such a fix, and the error is not raised.

randall-vx:
@vincent-hatakeyama 

> That would be best if possible. Otherwise, a simple scenario to reproduce the error would help. I didn’t see the error when testing with a local keycloak, but it might be linked with an option I did not test or something similar.

On AD server side, we have a 256 encryption key, so a `_validate_auth_response(token, base_url)` returns a `matching_value` doesn't match with a `response.name_id.text`. Without `encryption_keypairs` cannot decrypt `/tmp/file1.xml` to compare with  `/tmp/file2.xml`.

Also, (I hope, in a short time) to make an MR where the metadata can be signed with other certificate than CA Certificate.

> The version 15.0 of the module is in production without having such a fix, and the error is not raised.

Now I'm running on v16.0 production server. 

cc: @rolandojduartem

randall-vx:
Hi @vincent-hatakeyama do you have any news?

regards,

vincent-hatakeyama:
I’ve tested auth_saml in odoo 16 again with a production keycloak server. It works correctly, so this can not be a fix.

There’s encryption configuration in keycloak. When enabled, the authentication does not work anymore, but that’s to be expected as the module does not handle encryption.

The changes can not be merged as is because they will break all installations of auth_saml. This needs to be configurable by SAML provider like other options are, and ideally, the keys used need for the encryption should be configurable too.

PR #555, opened 2023-09-23
[16.0][IMP] auth_session_timeout adapt to use websocket instead of longpolling

bosd:
Migration leftover see: 
https://github.com/OCA/server-auth/pull/507#pullrequestreview-1603549270

### DIFF
--- a/auth_session_timeout/data/ir_config_parameter_data.xml
+++ b/auth_session_timeout/data/ir_config_parameter_data.xml
@@ -9,6 +9,6 @@
     </record>
     <record id="inactive_session_time_out_ignored_url" model="ir.config_parameter">
         <field name="key">inactive_session_time_out_ignored_url</field>
-        <field name="value">/calendar/notify,/longpolling/poll</field>
+        <field name="value">/calendar/notify,/bus/im_status</field>
     </record>
 </odoo>


thomaspaulb (approved):

thomaspaulb:
@bosd Any reason why this was still on draft? Otherwise I will just merge, since it should have been part of https://github.com/OCA/server-auth/pull/507

bosd:
> Any reason why this was still on draft?

It's been a while. Actually I don't remember. :)
Likely I was not sure of these changes.


In general I think the tests of this module needs to be changed.
As before the module was not working. Yet tests where still passing.

thomaspaulb:
Hm, I looked into it for a bit and in 16.0, `/bus/im_status` does not exist, rather things seem to go through a [websocket](https://github.com/odoo/odoo/blob/16.0/addons/bus/controllers/websocket.py#L13), and as such they are not regular polling calls, but persistent websocket connections through which data can flow.

Hence I'm not even sure if "ignored_urls" are needed at all.

If we wanted to add a unit test, one ideais to do a naive "tour" test - open a headless Chrome browser, set timeout to 1 minute, do nothing but let any browser background calls do their thing, then check whether session has indeed timed out, or if it has been woken up by some websocket call.

Of course we could actively simulate such websocket calls as well, that's another option.

As for functional experience - are you using this somewhere in production, and does it work?

bosd:
>As for functional experience - are you using this somewhere in production, and does it work? 

Well, I know a few places where it is installed in production.
But I'm currently not associated with that partner.

Side note: Currently in the process of becoming an offical partner.

Long time ago I did a successfull functional test "Manually" on V16.

bosd:
@bizzappdev Can you please review?

thomaspaulb:
Fasttracking since it was already approved for 17.0 and merged there; treating that like a positive review here also.

Issue #556, opened 2023-09-25
auth_saml : fix saml login with no providers 


FlorianDesp:
## Module
auth_saml

## Describe the bug
Traceback when trying to access /web/login without auth_oauth module installed

## To Reproduce
**Affected versions**:  15.0

1. Install auth_saml and ensure that auth_oauth is not installed
2. Configure a saml provider
3. Go back to web/login
![image](https://github.com/OCA/server-auth/assets/81736540/f7ba31ab-7741-4095-90d0-dfb02cef8742)

Related commit in which variables were renamed :  https://github.com/OCA/server-auth/commit/0dd018b5eb682c5eba861e77e743249caaed77f1

Fixed this on my locahost by changing this : 
```xml
<em
  t-attf-class="d-block text-center text-muted small my-#{len(providers) if len(providers) &lt; 3 else 3}"
>- or -</em>
```
by 
```xml
<em
  t-attf-class="d-block text-center text-muted small my-#{len(saml_providers) if len(saml_providers) &lt; 3 else 3}"
>- or -</em>
```

vincent-hatakeyama:
Reproduced on runboat.

vincent-hatakeyama:
Just checked, that does not apply to 16.0, the merged fix is already present for version 16.0.

I don’t have the right to change the issue, but it should be changed to solved.

FlorianDesp:
THX

PR #559, opened 2023-09-29
[FIX] auth_saml : fix provider connection without auth_oauth module

FlorianDesp:
Related issue : https://github.com/OCA/server-auth/issues/556


vincent-hatakeyama:
Tested on runboat, that fixes the issue in the ticket and the provider is visible.
/ocabot merge patch

PR #561, opened 2023-10-03
[16.0] Add auth_oidc_environment

adrienpeiffer:
From : https://github.com/OCA/server-env/pull/143


adrienpeiffer:
@AnizR @sbidoul @simahawk

AnizR (approved):
Great work! 
Code LGTM and already tested on the field (from https://github.com/OCA/server-env/pull/143)

auth_oidc_environment/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,14 @@
+To configure this module, you need to:
+
+Create a module server_environment_file with a cfg file or set the environment variable SERVER_ENV_CONFIG with the following section:
+
+[auth_oauth_provider.<name>]
```
simahawk, line 5: I always recommend to use tech_name instead which is more reliable

simahawk (approved):
LG

auth_oidc_environment/readme/CONFIGURE.rst
sbidoul: Agreed. There is no such field on oauth providers, though.

simahawk: that's why we have the techname.mixin (which I wonder if we shouldn't make the default :thinking: )

sbidoul: Ah, interesting. Thanks for the tip!

PR #572, opened 2023-11-14
[16.0][ADD] auth_oidc_portal: Link OAuth provider on portal user create

CRogos:
Adds OAuth Provider field to grant portal access wizard

image

feedback highly welcome.


CRogos:
Anyone an idea why this unit test is failing after adding my module? 

```
2023-11-14 15:26:59,666 831 ERROR odoo odoo.addons.password_security.tests.test_signup: FAIL: TestPasswordSecuritySignup.test_02_signup_user_success
  File "/__w/server-auth/server-auth/password_security/tests/test_signup.py", line 63, in test_02_signup_user_success
    self.assertEqual(
AssertionError: '/my' != '/web/login_successful?account_created=True'
- /my
+ /web/login_successful?account_created=True
```

CRogos:
@sbidoul: do you have an hint, how to fix this test compatibility issue?

CRogos:
The test fails, when the "portal" module is installed. I've changed the test to be compatible with the changed return url.

CRogos:
@dreispt what do you think of this addon?

PR #573, opened 2023-11-14
[16.0][IMP] auth_oidc: prompt for account on AAD login

CRogos:
Added the possibility to add additional parameters to the auth_link. 
When using the AzureAD auth, this gives the ability to set the "prompt" parameter to "select_account" which allowes to display the  select account dialog during the login process. This is helpful, when you have multiple AzureAD accounts.

image


auth_oidc/controllers/main.py
```diff
@@ -43,7 +44,17 @@ def list_providers(self):
...
+                # append provider specific auth link params
+                params_upd = literal_eval(provider["auth_link_params"])
+                params.update(params_upd)
+
                 # auth link that the user will click
+                provider["auth_link"] = "%s?%s" % (
+                    provider["auth_endpoint"],
+                    url_encode(params),
+                )
+
```
sbidoul:
```suggestion
```

This is already done just below.

auth_oidc/views/auth_oauth_provider.xml
```diff
@@ -18,6 +18,7 @@
             <field name="validation_endpoint" position="after">
                 <field name="token_endpoint" />
                 <field name="jwks_uri" />
+                <field name="auth_link_params" />
```
sbidoul: Maybe place this field next to the `auth_endpoint` field from the parent view?

sbidoul (commented):
Nice!

auth_oidc/views/auth_oauth_provider.xml
sbidoul: What I meant here is something like this, so the auth_link_params field is next to the auth_endpoint field on screen, as they are so closely related:

```
<field name="auth_endpoint" position="after">
  <field name="auth_link_params" />
</field>
```

sbidoul (approved):
Still a minor ergnonomy issue, otherwise looks good.

sbidoul:
Ah, can you rebase and resolve the conflict on the CONFIGURE readme fragment?
There were rendering issues with the images, and while fixing that I took the opportunity to convert them to markdown.

lucuswolfius (approved):
Code lgtm, nice work

CRogos:
@sbidoul can you merge?

dreispt:
@CRogos Can you resolve the conflict please?

CRogos:
@dreispt it is rebased. can you merge?

sbidoul:
@CRogos do you plan to port this to 17?

PR #578, opened 2023-11-21
[17.0][MIG] auth_admin_passkey

amkarthik:
cc: @SodexisTeam


pedrobaeza:
`[IMP] auth_admin_passkey: black, isort, prettier` commit contains a .deb file that shouldn't be there, and the version should be changed in the other commit. Another extra thing is that the commit message is not correct. See https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

amkarthik:
@pedrobaeza 
Done with the changes. 
Thanks

legalsylvain (approved):
Diff review. LGTM. thanks !

atchuthan (approved):
Code review :+1:

PR #582, opened 2023-11-29
[16.0] auth_jwt_demo: add test

sbidoul:
Test behaviour when we pass a valid token for a non-existent partner.


sbidoul:
/ocabot merge patch

Fast tracking this simple additional test.

PR #587, opened 2023-12-15
Small bugfix, RelayState can be empty

tb-flyt:


vincent-hatakeyama:
The title of the commit should be “[FIX] auth_saml: RelayState can be empty”.

This should backported too.

PR #591, opened 2023-12-16
[17.0][MIG] auth_api_key: Migration to 17.0.

chrisb-c01:


pedrobaeza:
Isn't this a native feature since the integration of TOTP?

chrisb-c01:
> Isn't this a native feature since the integration of TOTP?

Do you mean the ability for a user to generate an API key from their profile page? There is some overlap indeed, but imho there is still use for this addon. 

* You can (as an admin) centrally manage the API keys assigned to different users from the settings page. 
* Also having the `auth='api_key'` param on the controller allows you to permit authenticating using the API key but not just any user/password combination.

chrisb-c01:
Unit tests were failing before because clearing the orm cache using `clear_cache()` is not available anymore in 17.0 on the cache object. It was removed in this [commit](https://github.com/odoo/odoo/commit/595aa248433246959a5fa9288e477091701c6a35#diff-3d5ccf8623cd4456c775fef15f0cc973b6f88f05c0ce3b04ab113324e589b851L60) with the introduction of multiple caches. Needed to call `env.registry.clear_cache()` instead. Might be useful for other migrations.

auth_api_key/models/auth_api_key.py
```diff
@@ -0,0 +1,62 @@
...
+                return api_key.id
+        raise ValidationError(_("The key %s is not allowed") % key)
+
+    @api.model
+    @tools.ormcache("key")
+    def _retrieve_uid_from_api_key(self, key):
+        return self._retrieve_api_key(key).user_id.id
+
+    def _clear_key_cache(self):
+        self.env.registry.clear_cache()
```
chrisb-c01, line 49: Unit tests were failing before because clearing the orm cache using `clear_cache()` is not available anymore in 17.0 on the cache object. It was removed in this [commit](https://github.com/odoo/odoo/commit/595aa248433246959a5fa9288e477091701c6a35#diff-3d5ccf8623cd4456c775fef15f0cc973b6f88f05c0ce3b04ab113324e589b851L60) with the introduction of multiple caches. Needed to call `env.registry.clear_cache()` instead. Might be useful for other migrations.

pedrobaeza:
OK, IMO you should add that explanations to the README.

gurneyalex (approved):

FrancoMaxime (approved):
LGTM

auth_api_key/models/auth_api_key.py
FrancoMaxime: @chrisb-c01 Thanks for the details!

PR #593, opened 2023-12-20
[16.0][FIX] *_ropc: walrus operator not on python 3.7

NL66278:
Remove an assignment expression (walrus operator) from the code, as these are introduced with python 3.8, whereas the minimum python version for Odoo 16.0 is python 3.7.


ddejong-therp (approved):
LGTM.
Have run into the same problem. Code is able to compile again in python 3.7.

PR #594, opened 2023-12-20
[15.0][FIX] password_security: Allow password hash updates

JonathanStein:
Updating the password hash fails, if "_set_encrypted_password" is called as a regular user.

The core _set_encrypted_password method executes a direct query without any access checks, so we assume it's safe to operate with sudo() here.

The problem was seen after https://github.com/odoo/odoo/pull/146865 where hash strength is increased, resulting in password hash updates without a normal password change.

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -215,7 +215,7 @@ class ResUsers(models.Model):
         res = super(ResUsers, self)._set_encrypted_password(uid, pw)
         if not self.env.user.company_id.password_policy_enabled:
             return res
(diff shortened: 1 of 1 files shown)


fkantelberg (approved):
Tested it and works. I do agree that the sudo is ok.

I have seen the PR too late but closed mine with the same fix. If you want to copy the unittest from #609 because it's an interesting case we have here.

Also related to #604

JonathanStein:
Closing this PR because https://github.com/OCA/server-auth/pull/609 has the same fix but also includes test code

PR #595, opened 2023-12-23
[DOC]Add option to configure keycloak

flotho:


auth_oidc/readme/CONFIGURE.md
```diff
@@ -57,6 +57,7 @@ In Odoo, create a new Oauth Provider with the following parameters:
 - Provider name: Keycloak (or any name you like that identify your
   keycloak provider)
 - Auth Flow: OpenID Connect (authorization code flow)
+- Token Map : `preferred_username:user_id`
```
sbidoul: Perhaps this should be presented as a possible alternative. `sub:user_id, or `email:user_id` are other common possibilities.

flotho: Hi @sbidoul , actually I added this by checking your demo datas because I didn't succeed without this option.
but you're right, other mapping are possible !

sbidoul:
Retrying as the previous attempt was aborted due to codecov problems.

/ocabot merge patch

sbidoul:
It looks like `password_security` tests now fail.

sbidoul:
16.0 seems to be green now, let's try again.

/ocabot merge patch

PR #596, opened 2023-12-25
[16.0][FIX] users_ldap_groups: safe LDAP decode

oh2fih:
The group mapping in query mode fails if LDAP returns binary data in any of the fields. This adds a function that handles such situation by base64 encoding it.

### DIFF
--- a/users_ldap_groups/models/res_company_ldap_operator.py
+++ b/users_ldap_groups/models/res_company_ldap_operator.py
@@ -1,6 +1,7 @@
 # Copyright 2012-2018 Therp BV <https://therp.nl>
 # Copyright 2018 Brainbean Apps <https://brainbeanapps.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+import base64
 from logging import getLogger
 from string import Template
 
@@ -32,10 +33,27 @@ class ResCompanyLdapOperator(models.AbstractModel):
 
     def _query(self, ldap_entry, mapping):
         query_string = Template(mapping.value).safe_substitute(
-            {attr: ldap_entry[1][attr][0].decode() for attr in ldap_entry[1]}
+            {
+                attr: self.safe_ldap_decode(ldap_entry[1][attr][0])
+                for attr in ldap_entry[1]
+            }
         )
 
         results = mapping.ldap_id._query(mapping.ldap_id.read()[0], query_string)
         _logger.debug('Performed LDAP query "%s" results: %s', query_string, results)
 
         return bool(results)
+
+    def safe_ldap_decode(self, attr):
+        """Safe LDAP decode; Base64 encode attributes containing binary data.
+        Binary data can be stored in Active Directory, e.g., thumbnailPhoto is
+        stored as binary. As Str.decoce() cannot handle binary, this method
+        Base64 encodes the data in the attribute, instead, if decode fails.
+        Using Base64 should be a suitable fallback, because the use cases of
+        users_ldap_groups do not really require comparing binary attributes.
+        """
+
+        try:
+            return attr.decode()
+        except UnicodeDecodeError:
+            return base64.b64encode(attr)
--- a/users_ldap_groups/tests/test_users_ldap_groups.py
+++ b/users_ldap_groups/tests/test_users_ldap_groups.py
@@ -225,3 +225,70 @@ class TestUsersLdapGroups(TransactionCase):
                 self.env["res.users"].sudo().authenticate(
                     self.env.cr.dbname, "users_ldap_groups-username", "password", {}
                 )
+
+    def test_users_ldap_groups_ldap_returns_binary_data(self):
+        self._create_ldap_config(
+            groups=[
+                {
+                    "ldap_attribute": "name",
+                    "operator": "contains",
+                    "value": "hello3",
+                    "group_id": self.group_system.id,
+                },
+                {
+                    "ldap_attribute": "name",
+                    "operator": "contains",
+                    "value": "hello",
+                    "group_id": self.group_user.id,
+                },
+                {
+                    "ldap_attribute": "name",
+                    "operator": "contains",
+                    "value": "hello2",
+                    "group_id": self.group_contains.id,
+                },
+                {
+                    "ldap_attribute": "name",
+                    "operator": "equals",
+                    "value": "hello",
+                    "group_id": self.group_equals.id,
+                },
+                {
+                    "ldap_attribute": "",
+                    "operator": "query",
+                    "value": "is not run because of patching",
+                    "group_id": self.group_query.id,
+                },
+            ],
+            only_ldap_groups=True,
+        )
+        with mock.patch(
+            _company_ldap_class + "._connect",
+            return_value=FakeLdapConnection(
+                {
+                    "dc=users_ldap_groups,dc=example,dc=com": {
+                        "cn": [b"User Name"],
+                        "name": [b"hello", b"hello2"],
+                        "thumbnailPhoto": [
+                            b"GIF89a\x01\x00\x01\x00\x00\xff\x00,"
+                            b"\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x00;"
+                        ],
+                    }
+                }
+            ),
+        ), mock_cursor(self.cr):
+            user_id = (
+                self.env["res.users"]
+                .sudo()
+                .authenticate(
+                    self.env.cr.dbname, "users_ldap_groups-username", "password", {}
+                )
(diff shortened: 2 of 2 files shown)


oh2fih:
This same bug affects all other versions, too.

I'm about to give up writing the test coverage for this change because codecov keeps complaining about everything I try. (Especially I didn't understand why it demands test coverage for the tests, too.) My idea is to add binary data to the LDAP response, e.g., a minimal GIF image as the LDAP attribute `thumbnailPhoto`, as this is a common case where LDAP returns binary data.

```
{
    "dc=users_ldap_groups,dc=example,dc=com": {
        "cn": [b"User Name"],
        "name": [b"hello", b"hello2"],
        "thumbnailPhoto": [
            b"GIF89a\x01\x00\x01\x00\x00\xff\x00,"
            b"\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x00;"
        ],
    }
}

**Comment by oh2fih** [comment]
Tested with Odoo 16.0.20240102 on Ubuntu 22.04.

**Comment by oh2fih** [comment]
This patch now has 100% test coverage after adding a fully functioning test where LDAP returns binary data. I'm leaning towards the conclusion that the `codecov/project` pipeline is simply broken...

**Comment by oh2fih** [comment]
The patches have now been rearranged into coherent commits that introduce one fix at a time and adds the required test coverage in parallel, so this pull request should be easy to review.

Both the patches have been tested with Odoo 16.0.20240328 on Ubuntu 22.04. All unit tests are also passing with 100% test coverage (`100.00% of diff hit (target 91.95%)`). The pull request patches a vulnerability involving [Unsafe Public Methods](https://www.odoo.com/documentation/17.0/developer/reference/backend/security.html#unsafe-public-methods) & fixes a bug that renders the entire module useless whenever LDAP returns binary data in any of the fields.

I can see that `users_ldap_groups` is missing a named maintainer, but not fixing this kind of bugs and vulnerabilities for months surely seems like this @OCA is dead. Could, e.g., @joao-p-marques take a look at this as `users_ldap_mail` & `users_ldap_populate` are pretty close to this. Thank you!

**Review by rvalyi — APPROVED** [review]
I'm no ldap specialist at all, but I understand the vulnerability and your fix seems both reasonable and safe so I approve it. Let's give it a couple more days to see if we get another reviewer. If not I will be able to merge it.

But yeah the ldap module isn't very actively maintained and possibly the maintainers just missed your issue. You could eventually propose yourself as a module maintainer (adding your GitHub user to the maintainers value of the module manifest file in a PR will be enough to enable to let you merge and publish improvements in the module).

Thank you for the fix.

File: users_ldap_groups/models/res_company_ldap_operator.py [hunk 1]
```diff
@@ -20,22 +20,33 @@ def operators(self):
...
+            }
         )
 
         results = mapping.ldap_id._query(mapping.ldap_id.read()[0], query_string)
         _logger.debug('Performed LDAP query "%s" results: %s', query_string, results)
 
         return bool(results)
+
+    def safe_ldap_decode(self, attr):
+        import base64
```
sbidoul: Why not importing at the top?

users_ldap_groups/models/res_company_ldap_operator.py
```diff
@@ -20,22 +20,33 @@ def operators(self):
...
+
+        try:
+            return attr.decode()
+        except UnicodeDecodeError:
+            return base64.b64encode(attr)
```
sbidoul, line 59: A comment explaining why a base 64 decoding fallback makes sense would be helpful.

sbidoul (commented):
Thanks for submitting this fix.

For next time, note security@odoo-community.org may be contacted for reporting such issues.

users_ldap_groups/models/res_company_ldap_operator.py
oh2fih: Good point! As it is imported anyway, and not just if the exception occurs, it doesn't make technical difference to be at the top; moved there.

users_ldap_groups/models/res_company_ldap_operator.py
oh2fih: A docstring explaining the decision made added for the method.

sbidoul:
This decode-that-encodes-in-base64 looks a bit weird to me. Why is the `decode` necessary in the first place?

Perhaps, to move faster, you could do the security fix in a PR, so I can merge this, and do the decode improvement in another PR and work on that one with folks who are more familiar with this module?

oh2fih:
> This decode-that-encodes-in-base64 looks a bit weird to me. Why is the `decode` necessary in the first place?

This simply replaces the original `ldap_entry[1][attr][0].decode()` doing the fix with the least possible modifications to the module. The `Str.decode()` was added in ad15e3aa96d7d6f04aba840cd1499ad2c3af78ea when migrating the module from Odoo 10 to Odoo 12 and seems quite necessary, because it did not work without it on the first attempts to address this bug. The `contains` & `equals` uses the `Str.decode()` (in `x.decode()`), too, so not using it here does not seem coherent, either. Although not explained in the commit, I guess the root cause might be support for non-ASCII names.

So, why not use this `safe_ldap_decode()` in `contains` & `equals` then? Because only the requested LDAP attributes are compared and they should never contain binary data. This is a special case for the `query` operator for which LDAP returns the entire Active Directory object with all the attributes.

Does this satisfactorily explain the situation here?

oh2fih:
> Perhaps, to move faster, you could do the security fix in a PR

@sbidoul: There's now another PR https://github.com/OCA/server-auth/pull/659 that only contains the commit fixing the vulnerability. After merging that this PR is only the second commit ahead of upstream `16.0` branch.

oh2fih:
Rebased to fix the checks that didn't pass; the problem was in another module.

Could someone please review & merge this already?

oh2fih:
Rebased over `fc906889df964a3f8b50013389dbc585aee0080a` after workflow fixed in https://github.com/OCA/server-auth/issues/693.

oh2fih:
@sbidoul It seems difficult to get anyone that actually knows something about LDAP to review this fix. However, this has been running on an instance for months and has fixed a real-life problem there. The problematic LDAP response has been demonstrated in a new test this PR adds, and all the previous tests are passing, too. Could this finally be merged?

pedrobaeza (approved):
/ocabot merge patch

users_ldap_groups/models/res_company_ldap_operator.py
sbidoul: I'm not familiar with LDAP, but decoding without specifying the encoding and this kind of fallback seem suspicious to me. Does LDAP provide a mechanism to know the encoding of a field in advance, or to know if a field is effectively a binary field?

pedrobaeza: This is encoding, not decoding, so I saw it safe enough, checking also the test.

oh2fih: The `x.decode()` is there in the `_contains()` & `_equals()`, too, and has been part of this module since it was migrated to 12.0 in [[MIG] users_ldap_groups: Migration to 12.0](https://github.com/OCA/server-auth/commit/ad15e3aa96d7d6f04aba840cd1499ad2c3af78ea) ([blame](https://github.com/OCA/server-auth/blame/4501d8a8772e90b99be07ff3a093f4036e6a1a72/users_ldap_groups/models/res_company_ldap_operator.py)). This PR never added it, but only the `base64.b64encode(attr)` as a fallback in case it fails.

Maybe @alexey-pelykh could shed some light on the background behind this decision.

alexey-pelykh: I concur with @oh2fih that such fix should be consistent across the codebase. First and foremost - do we have actual usecase when we need binary data and don't skip the non-UTF8-decodable string?

alexey-pelykh: > So, why not use this safe_ldap_decode() in contains & equals then? Because only the requested LDAP attributes are compared and they should never contain binary data. This is a special case for the query operator for which LDAP returns the entire Active Directory object with all the attributes.

Okay, that explains why only `_query` yet the part "do we actually have any need for binary fields like photos" remain

oh2fih: No, we do not **need** binary fields, but we need to expect to receive them. Another option would be to disregard this data and return an empty value. However, the base64 approach likely makes it easier to debug what data was returned that caused the problem.

alexey-pelykh: I agree on debugging

PR #599, opened 2024-01-03
[14.0][FIX] auth_saml: password reset when deactivating the config settings

leemannd:
**issue**
- have users with password
- install the module `auth_saml`
- activate the paramter that allows to login with the password
- setup of the user with the saml provider
- update the settings by removing the possibility to log with the password

This will delete the ir.config parameter, but not trigger the removal of the password.

**Fix**
Now when we update the settings and unlink the related ir.config.parameter. We trigger the removal of passwords

Once this PR is merged, I'll forward port it to next versions


Ricardoalso (approved):

santostelmo (approved):

leemannd:
Hello @vincent-hatakeyama as you recently worked on this module, do you have time for a review and eventually a merge?

PR #611, opened 2024-02-01
[MIG][16.0] Migration of vault

fkantelberg:
The migration needed quite a rework because of the JS framework changes. (I hope the next are easier) I tested every functionality during migration but plan to do another round again soon.

There are some minor fixes which were possible to do:
- Better file handling. Inside of an entry you can send files to an inbox now
- Adjusting the iterations for PBKDF2 to the OWASP recommendations >= 600,000. The user keys should be automatically migrated if the user uses it the first time [backportable]


pedrobaeza:
Thanks for the migration.

Can you please leave this one only for the module `vault` and o later `vault_share` with its story. Another thing you can do is to compact a bit the commit story, as it contains some non useful commits, like the first 11.

fkantelberg:
Done and `vault_share` is now in #612

vault/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Vault",
     "summary": "Password vault integration in Odoo",
     "license": "AGPL-3",
-    "version": "15.0.2.1.0",
+    "version": "16.0.2.1.0",
```
CarlosRoca13:
```suggestion
    "version": "16.0.1.0.0",

**Review by CarlosRoca13 — CHANGES_REQUESTED** [review]
It’s a bit annoying that to press the button for generating secure keys, we have to search for where the button allows us to press. In the following gif, I’ll show you an example.

Button example


I'm getting an error adding more than a field to the vault. This is the error:
```
Error: couldn't find x2many field name
    at Record.update (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:2368:437)
    at async Object.update (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:3004:263)
    at async VaultField.storeValue (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:14166:105)
    at async VaultField.commitChanges (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:14132:286)
    at async Promise.all (index 2)
    at async Record.checkValidity (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:2324:41)
    at async Record.onRecordWillSwitchMode [as _onWillSwitchMode] (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:2395:214)
    at async Record.switchMode (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:2335:17)
    at async Record.onRecordWillSwitchMode [as _onWillSwitchMode] (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:2395:262)
    at async Record.switchMode (http://localhost:16069/web/assets/684-f75f43c/web.assets_backend.min.js:2335:17)
```

I paste you a gif with the steps too:
error vaulr

The bug can be seen when entering fields, which makes it impossible to add the second one.

File: vault/__manifest__.py (hunk 1 above)
Comment (fkantelberg) [reply vault/__manifest__.py]: Which browser have you tested with? (especially regarding the first error)

Comment (CarlosRoca13) [reply vault/__manifest__.py]: I'm using chrome

Comment (fkantelberg) [reply vault/__manifest__.py]: Thanks.

First problem I was able to fix with some CSS.

Second problem was within the JS of a vault_field and is fixed too. There was an endless loop.

Furthermore the encrypted values aren't show as tooltip anymore.

**Review by CarlosRoca13 — COMMENTED** [review]
I had another error when I tried to display a field already created without entering the key.
```
TypeError: Cannot read properties of undefined (reading 'privateKey')
    at Vault.get_private_key (http://localhost:16069/web/assets/686-878a514/web.assets_backend.min.js:14223:68)
    at async Vault.unwrap (http://localhost:16069/web/assets/686-878a514/web.assets_backend.min.js:14238:63)
    at async VaultField._getMasterKey (http://localhost:16069/web/assets/686-878a514/web.assets_backend.min.js:14172:177)
    at async VaultField._decrypt (http://localhost:16069/web/assets/686-878a514/web.assets_backend.min.js:14173:95)
    at async VaultField._onShowValue (http://localhost:16069/web/assets/686-878a514/web.assets_backend.min.js:14124:139)
```
ERROR

I think in this case it should ask for the key

fkantelberg:
I was able to reproduce the issue. Seems like a hidden bug in all versions. I reproduced it by clicking away the first password dialog when I clean all browser caches. If I try to read `this.time` was null which causes an early return instead of reimporting of the keys.

CarlosRoca13:
I'm giving access to demo user but he can not see any secret... See gif below

Access_error

fkantelberg:
In my tests I had the issue that the key in vault.right wasn't generated correctly because of the if in the controller. I suspect that this is also the issue for your case. Not 100% sure because the right was marked red in my case but isn't in your case.

bosd:
image
The buttons in the topbar, don't seem to be working.
Import file, Export file, Verify, Re-encrypt



Also the smart button "Entries" is no longer working.

fkantelberg:
@bosd I can't reproduce it and every button is working. Maybe a module update gone wrong on your local instance? At least the smart button doesn't do anything special just uses a normal action.

CarlosRoca13:
> In my tests I had the issue that the key in vault.right wasn't generated correctly because of the if in the controller. I suspect that this is also the issue for your case. Not 100% sure because the right was marked red in my case but isn't in your case.

I think the problem was that, I have deleted and re-added the user and there has been no problem.

I think it's ready :+1: :smile:

CarlosRoca13 (approved):

bosd:
> Maybe a module update gone wrong on your local instance?

I was using the runboat.. Will reset it and try again.

fkantelberg:
@bosd Please check that you aren't using `http`. `https://` is required for the module because the browser enforces it for the secure context and browser side encryption.

pedrobaeza:
@fkantelberg maybe a clarification in the ROADMAP of the module mentioning explicitly runboat there?

bosd:
Thanks,with  `https`  The import window is working now.
I'm getting more clickable icons on the screenlike in above gif's.

Export throws an error.. 
image

Viryfy buttons as well:
UncaughtPromiseError > OperationError
Uncaught Promise
OperationError

The smartbutton:
image

fkantelberg:
There is a line in the readme `This modules requires a secure context for the browser to work properly.` but I guess I'm just biased and know what it means. Roadmap sounds like that I want to work on "fixing" it but http will never be supported beside localhost.

Is there a technical reason why the runboat links to http instead of the possible https per default?

@bosd 
Regarding the error: I assume it just happens because it was possible to create a vault without https. The vault and user keys aren't initialized correctly and can be thrown away. I have to check if I can further prevent anything when the context isn't secured.

pedrobaeza:
> Roadmap sounds like that I want to work on "fixing" it but http will never be supported beside localhost.

The file is called ROADMAP.rst, but the section is "Known issues / Roadmap".

> Is there a technical reason why the runboat links to http instead of the possible https per default?

@sbidoul can tell you, but if the certificate is self-signed, it's for not having the browser alert and have more questions about that.

sbidoul:
> Is there a technical reason why the runboat links to http instead of the possible https per default?

It's not a technical reason, but to avoid the burden of maintaining a trusted wildcard certificate.

bosd:
Thanks for the clarifications. 
Did'nt spot that line in the readme. Even if i did I don't know what it means :zipper_mouth_face: 

I'm used to test stuff on runboat. Did'nt occur to me that is not possible for this module.
Guess this is the exception of the rule.

fkantelberg:
You can just swap to https and confirm the self signed certificate of the runboat to test it.

I understand that a wildcard cert can be a burden. I see how I can make it a bit more error prune when not using a secure context and check if I can rephrase the hint in the readme.

fkantelberg:
The latest patch does some things when the secure browser context isn't available:

- Disables the buttons for vault list and form view (no create, save, discard)
- Throws a descriptive error when hitting an action that https should be tried or the admin contacted

Additionally I extended the readme and added it also to the roadmap. I hope this makes it visible enough :smile:

bosd (approved):
> The latest patch does some things when the secure browser context isn't available:

Hehehe, That's one feature I could test succesfuly 😀

PR #626, opened 2024-03-15
[14.0] [FIX] auth_jwt: Unhandled PyJWKClientError in _get_key

paradoxxxzero:
This commit https://github.com/OCA/server-auth/commit/5e726ac0084de377a046308d21bf39fa05ad3f4c breaks the next_validator_id loop.

`_get_key` calls `jwks_client.get_signing_key(kid)` which can raise PyJWKClientError which is not caught by the validation loop (only catching Unauthorized)

Putting `_get_key` in the preceding try except fix this problem since it catches all Exception

### DIFF
--- a/auth_jwt/models/auth_jwt_validator.py
+++ b/auth_jwt/models/auth_jwt_validator.py
@@ -189,10 +189,10 @@ class AuthJwtValidator(models.Model):
         else:
             try:
                 header = jwt.get_unverified_header(token)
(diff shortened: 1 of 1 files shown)


sbidoul (approved):
Looks good.

sbidoul:
/ocabot merge patch

@paradoxxxzero do you plan to forward port?

PR #628, opened 2024-03-22
[16.0][MIG] password_security: Add flag to toggle password security policy

hwangh95:
This is a forward-ported commit of #500. Some portions were modified to handle multi-company environments


hwangh95:
There's codecov issues, but other than that it should be good to go

hwangh95:
@simahawk Could I get you to look at this commit?

thomaspaulb (approved):

password_security/migrations/16.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    companies = env["res.company"].with_context(active_test=False).search([])
+    _logger.info("Enable the password policy on %s companies", len(companies))
+    _logger.info(companies.mapped("name"))
+    companies.write({"password_policy_enabled": True})
```
thomaspaulb, line 17: My only comment would be, that in case people migrate via openupgrade from 15 to 16, and they have the password policy switched off, then it would get enabled again.

Not a big problem, if that happens people can just switch it off again, but perhaps it makes sense to look at the version being migrated from, and take the correct approach.

PR #629, opened 2024-03-29
[14.0] [FIX] auth_api_key: return 401 instead of 403

renda-dev:
`AccessDenied()` has status code `403` while, [according to the docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) the status code `401` [suits better in this case](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) so I would like to propose to use `werkzeug.exceptions.Unauthorized` instead, which returns `401`.

### DIFF
--- a/auth_api_key/models/ir_http.py
+++ b/auth_api_key/models/ir_http.py
@@ -5,8 +5,9 @@
 
 import logging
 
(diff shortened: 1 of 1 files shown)


HekkiMelody (approved):
Code review, LGTM

PR #635, opened 2024-04-10
[16.0][IMP] remove python libs included in odoo requirements

CRogos:
same story as https://github.com/OCA/reporting-engine/pull/865

https://github.com/odoo/odoo/blob/16.0/requirements.txt


sbidoul (changes requested):
This topic comes over and over again.

I have never seen a good justification to remove such dependencies.
And there is a good justification to keep them: they document that they are direct dependencies of the module.

See for instance https://github.com/OCA/queue/pull/530#issuecomment-1755968900 and the surrounding thread.

CRogos:
@sbidoul Thanks you for sharing this link. It was very interesting to read and all arguments are already mentioned. I can understand both sites. I've added my comment there, to have it in the existing thread.

PR #636, opened 2024-04-12
[ADD][14.0] Authentication Show Hide Password

FrancescoBallerini:
Add buttons to hide/show password in login, signup and reset password form. It seems like an useful missing functionality to add in OCA.

I have found some free modules doing this in the odoo app store but they all use `replace` xpath attributes on password inputs. 

This implementation is a bit more verbose but avoid using `replace` on pw inputs.


FrancescoBallerini:
I have just realized by testing on runbot that this has some incompatibility with the "password meter" element introduced in signup by module `auth_password_policy_signup`. 

I will work on a solution to this.

FrancescoBallerini:
Update: 
- added some basic scss to avoid "password meter" element to overlap with the "show password" button
- added compatibility with Edge: the browser will display its own button, we disable the feature in extended form in order to avoid redundant buttons while this modue is installed 

The only "issue" remaining is the fact that the "password meter" item will temporary disappear while user focuses the "password" input (see images)

Input Password has focus (password meter disappear)

Screenshot 2024-04-14 162000


Move focus to another element (password meter appear again)

Screenshot 2024-04-14 161948

 I have made an attempt to solve this last one (you can find it on scss file) but I didn't manage to solve the issue

PR #637, opened 2024-04-17
[14.0][ADD] impersonate_login

Kev-Roche:
This module allows to login as another user, its purpose is mainly (only?) for testing and the support team.
As an alternative to auth_admin_passkey, for a more French/European data security complience. 

No tests wrote in this module yet, any hints / help will be appreciated.
After adding tests, I will migrate the module in v16.


impersonate_login/models/model.py
```diff
@@ -0,0 +1,23 @@
...
+        if request and request.session.impersonate_from_uid:
+            user = self.env["res.users"].browse(request.session.impersonate_from_uid)
+            return super(BaseModel, self.with_user(user))._create(data_list)
+        return super()._create(data_list)
+
+    def write(self, vals):
+        if request and request.session.impersonate_from_uid:
+            user = self.env["res.users"].browse(request.session.impersonate_from_uid)
+            return super(BaseModel, self.with_user(user)).write(vals)
+        return super().write(vals)
```
rvalyi: (same review as in the [the server-tools proposal](https://github.com/OCA/server-tools/pull/2900) )

May be I am mistaken but I have big doubts about this implementation: what if my module B overrides create or write (super common) and does something, (potentially even like writing or creating some other record) on the behalf of the current self.env.user before calling super().create or super().write?

Because impersonate_login will not depend on my module B, your _create and write overrides can be arbitrarily called before or after the ones in B... So I fear I have no guarantee your impersonate_login module will have impersonated the user yet in my overrides in B...

What about renaming the module base_schizophrenia or something like that?

Shouldn't this user switch be done during the login phase in res.users instead or in https://github.com/OCA/server-auth/blob/16.0/auth_admin_passkey/models/res_users.py#L57 ?

cc @sebastienbeau @legalsylvain

rvalyi (changes requested):

impersonate_login/models/model.py
sebastienbeau: In fact overwritting the write and create was done for setting the right create_uid, write_uid. @Kev-Roche will change it

CasVissers-360ERP:
@Kev-Roche Did you already forward port this to v16/v17?

impersonate_login/models/mail_message.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class Message(models.Model):
+    _inherit = "mail.message"
+
+    impersonated_author_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="Impersonated Author",
+        compute="_compute_impersonated_author_id",
+        store=True,
```
CasVissers-360ERP, line 16: This will take forever to compute in active databases.

CasVissers-360ERP: Maybe just ALTER TABLE mail_message ADD COLUMN impersonated_author_id INTEGER;

CasVissers-360ERP:
@Kev-Roche We are working on a v17 port. Like the module.

Kev-Roche:
> @Kev-Roche We are working on a v17 port. Like the module.

Hi @CasVissers-360ERP , thanks for your interest. I did not work on other versions yet because tests are missing and I m not sure if the module is working properly.

CasVissers-360ERP:
Well the module will make our lives a lot easier. I will ask more team members to review.

impersonate_login/models/res_users.py
```diff
@@ -0,0 +1,87 @@
...
+                logger.info(
+                    f"IMPERSONATE: {self._get_partner_name(self._uid)} "
+                    f"Login as {self._get_partner_name(self.id)}"
+                )
+
+                request.env["res.users"].clear_caches()
+                request.session.session_token = security.compute_session_token(
+                    request.session, request.env
+                )
+                return {
```
CasVissers-360ERP, line 63: If we go to a user without access to the current URL we get an access error, might be better to redirect to main menu?

astirpe:
@Kev-Roche 
Thank you for this module! I'm porting it to V17 with a few changes here: https://github.com/OCA/server-auth/pull/680

Please feel free to include some of the changes I did there, for example the revised `def _create()` and `def write()` that I changed a bit to increase efficiency and robustness

Kev-Roche:
@CasVissers-360ERP , @astirpe thank you both for your time and help on this module, I bring some of your changes from V16/17 here.

PR #638, opened 2024-04-18
[14.0] Add module base_user_empty_password

grindtildeath:
This module provides a wizard to allow emptying users password field.

This could be useful to force the user to use another sign on method than Odoo.

### DIFF
--- a/base_user_empty_password/models/res_users.py
+++ b/base_user_empty_password/models/res_users.py
@@ -1,0 +1,25 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    has_password = fields.Boolean(compute="_compute_has_password")
+
+    def _compute_has_password(self):
+        # Bypass ORM as password is always empty in cache
+        self.env.cr.execute(
+            """SELECT id, password FROM res_users WHERE id IN %s;""", (tuple(self.ids),)
+        )
+        res = {row[0]: row[1] for row in self.env.cr.fetchall()}
+        for user in self:
+            user.has_password = bool(res.get(user.id))
+
+    def _empty_password(self):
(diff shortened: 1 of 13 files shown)


base_user_empty_password/models/res_users.py
```diff
@@ -0,0 +1,25 @@
...
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    has_password = fields.Boolean(compute="_compute_has_password")
+
+    def _compute_has_password(self):
+        # Bypass ORM as password is always empty in cache
+        self.env.cr.execute(
+            """SELECT id, password FROM res_users WHERE id IN %s;""", (tuple(self.ids),)
+        )
```
yankinmax, line 15: for security reasons maybe we can execute query like this:
```
from psycopg2 import sql
query = sql.SQL("""
    SELECT {field} FROM res_partner WHERE ID IN %(partner_ids)s FOR UPDATE NOWAIT;
    UPDATE res_partner SET {field} = {field} + %(n)s
    WHERE id IN %(partner_ids)s
""").format(field=sql.Identifier(field))
self.env.cr.execute(query, {'partner_ids': tuple(self.ids), 'n': n})
```
it's just a copy paste from v14 source code

yankinmax (approved):
Interesting module, thanks

sebalix (approved):

base_user_empty_password/models/res_users.py
simahawk: @yankinmax I'm not sure to understand what's the problem w/ security?

simahawk (approved):
LG

yankinmax:
@simahawk to avoid SQL injection if we need to execute a query on a database directly we incapsulate it with SQL.Identifier tools. Am I missing something here?

simahawk:
> @simahawk to avoid SQL injection if we need to execute a query on a database directly we incapsulate it with SQL.Identifier tools. Am I missing something here?

There's no injection here as it relies on psygopg doing proper interpolation of the data.

PR #639, opened 2024-04-19
[15.0] password_security: Fix login with MFA enabled

grindtildeath:
When MFA is enabled, the web_login function returns a redirect to /web/login/totp, but the session is not fully authenticated.

Therefore, the check done for expired password on request.env.user was executed on the public user instead of the user trying to log in.

Once the public user password is expired, it was then redirecting to the password reset form of the public user blocking the user from logging in.

Instead of overriding web_login, it's safer to override _login_redirect as this function is meant to customize the user redirection on login.

By checking if the user is fully logged before checking if its password is expired, we ensure it will not mess with any other authentication process.


grindtildeath:
ping @astirpe since you seem to have worked on similar issue in 16.0

sebalix (approved):
LG

JuMiSanAr (approved):

dreispt:
@grindtildeath Please squash your commits

grindtildeath:
@dreispt done.

PR #640, opened 2024-04-22
[15.0][FIX] auth_saml: do not force using vulnerable cryptography module

vincent-hatakeyama:
Without this fix, it is impossible to use an upgraded version of cryptography with this module.

It is still necessary to check for compatibility, so that information is added to the install instructions.

### DIFF
--- a/auth_saml/__manifest__.py
+++ b/auth_saml/__manifest__.py
@@ -12,14 +12,7 @@
     "license": "AGPL-3",
     "depends": ["base_setup", "web"],
     "external_dependencies": {
-        # Place an upper bound on cryptography version to be compatible with
-        # pyopenssl 19 mentioned in Odoo 15's requirements.txt. If we don't do
-        # this, installing this module will try to upgrade cryptography to the latest
-        # version because the minimum required version in pysaml2 (>=3.1) is greater than
-        # version 2.6 (from Odoo's requirement.txt). Since cryptography/pyopenssl don't
-        # declare minimum supported versions, this lead to inconsistencies.
-        # https://github.com/OCA/server-auth/issues/424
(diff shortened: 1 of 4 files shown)


gurneyalex (approved):

thomaspaulb:
@vincent-hatakeyama Could you rebase or allow access

vincent-hatakeyama:
> @vincent-hatakeyama Could you rebase or allow access

I’ve juste rebased.

thomaspaulb:
@sbidoul I need your input here.. this goes against [this](https://github.com/OCA/server-auth/pull/429/files). Is `test-requirements.txt` OK to edit directly or is there some `setup.py` magic that we should do to override the module dependency for the test, but not for the module itself?

Or, come to think of it, could we override it there, to let the requirements.txt include the limit but to leave the module's manifest free for users of the module to decide?

sbidoul:
I agree we should not place an upper bound on cryptography in the module. My initial approach was not optimal. This PR is better.

Last time I checked, the pin on pyopenssl 19 on Odoo's requirements.txt was not really necassary?
So another approach could be putting a minimum pyopenssl version in auth_saml2's external dependencies.
I think we have done that somewhere, but I can't find it right now.

Longer term, I would like to try and stop using Odoo's requirements.txt in oca/oca-ci, in favor of a custom set of upper bounds on dependencies that are known not to work with Odoo (things like `xlrd<2`, etc...).

vincent-hatakeyama:
> @vincent-hatakeyama Could you rebase or allow access

I’ve rebased again.

sbidoul:
Sounds like a good solution.

/ocabot merge patch

PR #643, opened 2024-05-06
[17.0] [MIG] auth_ldaps: Migration to 17.0

rven:


auth_ldaps/models/res_company_ldap.py
```diff
@@ -0,0 +1,50 @@
...
+import ldap
+
+from odoo import fields, models
+
+_logger = logging.getLogger(__name__)
+
+
+class CompanyLDAP(models.Model):
+    _inherit = "res.company.ldap"
+    _description = "Company LDAP configuration"
```
acsonefho: Not needed as it's an `_inherit`.

acsonefho (approved):

ryanc-me:
@rven I'm in the process of testing this at a client site - seems to be working well so far. Thanks for your efforts here!

Do you have some time to solve the review changes?

I would be happy to take over this PR if you're short on time. :)

auth_ldaps/models/res_company_ldap.py
ryanc-me: Agreed that it's better not to do this, but this was also present in 16.0, 15.0, etc.

Maybe it's better to fix it after upgrade with a "fix" PR, so the fix can be backported.

rven:
@ryanc-me Changes implemented, also added a separate commit for the removal of the _description attribute.

ryanc-me:
Awesome, thanks @rven! I'll re-test this today then do a review, so hopefully it can be merged :)

If you have a moment, maybe you can check out #675, #674, #673 - if one of them is useful for you, you could provide a review.

ryanc-me (approved):

nojusZeta (approved):

PR #645, opened 2024-05-08
[DRAFT][15.0][IMP] auth_password_pwned: check passwords against haveibeenpwned.com

ap-wtioit:
Adding a module that checks password hashes against the haveibeenpwned.com API to make sure publicly known passwords can not be set in Odoo and not used to login anymore. The user is forced to use password reset upon login if it is enabled, an email is sent to him (if auth_signup is installed) or he has to contact an admin to ensure a safe password change.

This is still work in progress. Feedback welcome.

Planned tasks (next week):
* [ ] adding tests
* [ ] remove password_security changes and move them into a seperate module (to be added later)
* [ ] adding translation
* [ ] fix license badge in README.rst

Further reading:
* https://www.troyhunt.com/authentication-and-the-have-i-been-pwned-api/
* https://haveibeenpwned.com/API/v3#SearchingPwnedPasswordsByRange


auth_password_pwned/controllers/main.py
```diff
@@ -0,0 +1,82 @@
...
+                [("name", "=", "auth_signup"), ("state", "in", ["installed"])], limit=1
+            )
+        )
+
+    def _reset_password_enabled(self):
+        return (
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param("auth_signup.reset_password")
+            == "True"
```
amh-mw, line 28: Seems aberrant to check for a string value to exactly equal True, but that does seem to match what the auth_signup module does.

auth_password_pwned/models/res_users.py
```diff
@@ -0,0 +1,59 @@
...
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    def _set_password(self):
+        self._passswordshavebeenpwned(self.mapped("password"))
+
+        super()._set_password()
+
+    def _passswordshavebeenpwned(self, passwords):
```
amh-mw: Misspelled _passsword with third s?

amh-mw (changes requested):
Commit should be prefixed [ADD] not [IMP], since a new module has been created.

PR #649, opened 2024-05-14
Fix: replace tenant id for single tentant

janikvonrotz:
Moved `-  replace {tenant_id} in urls with your Azure tenant id` to the correct section in the readme.

PR #650, opened 2024-05-14
[Fix] auth_oidc: move tenant id config step in readme

janikvonrotz:


janikvonrotz:
Supersedes: https://github.com/OCA/server-auth/pull/649

sbidoul:
Makes sense, thanks!

/ocabot merge patch

PR #652, opened 2024-05-17
[16.0][IMP] auth_saml: Allow admin/managers to edit User SAML information

vincent-hatakeyama:
As user in that group can already edit users, so it make sense to allow them to see and edit that information rather than restrict it to admin/system.

### DIFF
--- a/auth_saml/security/ir.model.access.csv
+++ b/auth_saml/security/ir.model.access.csv
@@ -1,6 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_auth_saml_provider,auth_saml_provider,model_auth_saml_provider,base.group_system,1,1,1,1
-access_auth_res_users_saml,auth_res_users_saml,model_res_users_saml,base.group_system,1,1,1,1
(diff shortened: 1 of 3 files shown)


auth_saml/security/ir.model.access.csv
```diff
@@ -1,6 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_auth_saml_provider,auth_saml_provider,model_auth_saml_provider,base.group_system,1,1,1,1
-access_auth_res_users_saml,auth_res_users_saml,model_res_users_saml,base.group_system,1,1,1,1
+access_auth_res_users_saml,auth_res_users_saml,model_res_users_saml,base.group_erp_manager,1,1,1,1
 access_auth_saml_attribute_mapping,auth_saml_attribute_mapping,model_auth_saml_attribute_mapping,base.group_system,1,1,1,1
-auth_saml_provider_erp_manager_access,auth_saml_provider_erp_manager_access,model_auth_saml_provider,base.group_erp_manager,1,0,0,0
```
PCatinean, line 5: I think we might still need this given that res.users.saml has a many2one to auth.saml.provider and one needs to select the provider when making/editing an entry

PCatinean (approved):

bizzappdev (approved):

PR #641, opened 2024-04-24
[17.0][MIG] auth_user_case_insensitive

CRogos:
Follow-up of: https://github.com/OCA/server-auth/pull/607

Commit history cleaned.

PR #657, opened 2024-05-30
[17.0][MIG] auth_jwt: Migration to 17.0

MikeAelbrecht:
Changed the manifest to version 17.0.1.0.0.
Updated the XML file to be supported by Odoo version 17.0.


MikeAelbrecht:
The tests fail because the response is not equal to the expected response. The reason is because the expected string is 95 character long instead of 88.  So I formatted the file so it would fit the pre-commit check. But this causes the tests to fail. How can I solve this issue?

auth_jwt/tests/test_auth_jwt.py
```diff
@@ -0,0 +1,406 @@
...
+        validator6.next_validator_id = validator7
+
+        authorization = "Bearer " + self._create_token()
+        with self._mock_request(authorization=authorization):
+            with self.assertRaises(UnauthorizedCompositeJwtError) as composite_error:
+                self.env["ir.http"]._auth_method_jwt_validator()
+            self.assertEqual(
+                str(composite_error.exception),
+                "401 Unauthorized: "
+                + "Multiple errors occurred during JWT chain validation:\n"
```
sbidoul, line 204:
```suggestion
                "Multiple errors occurred during JWT chain validation:\n"
```

Was this `+` added automatically by ruff ?

pandoc-3.2-1-amd64.deb
sbidoul, line 1: Yikes, what's this pandoc-3.2.1-amd64.deb doing here?

rvalyi: it already landed in the 17.0 branch! https://github.com/OCA/server-auth/tree/17.0 Here was the culprit https://github.com/OCA/server-auth/pull/641

sbidoul:
@MikeAelbrecht sorry I had lost track of this. Could you rebase, and get rid of the `.deb` file (it was added in [f7e9643](https://github.com/OCA/server-auth/pull/657/commits/f7e9643e4c75a95c71cea8e74bfe42b46382fb99))?

Otherwise the migration looks good to me.

StefanRijnhart:
Merged in https://github.com/OCA/server-auth/pull/751. Thank you for your contribution!

PR #659, opened 2024-06-06
[16.0][FIX] users_ldap_groups JSON RPC vulnerability

### LINKED ISSUE #617: Vulnerability in users_ldap_groups allows arbitrary LDAP queries via JSON RPC API
## Module
users_ldap_groups

## Describe the bug
Model `res.company.ldap.operator` operators should be private methods; public methods allow arbitrary LDAP queries via JSON RPC API.

## To Reproduce
**Affected versions**: all

**Steps to reproduce the behavior:**

Send `POST /web/dataset/call_kw/res.company.ldap.operator/query HTTP/1.1`, e.g., with data `{"id":5,"jsonrpc":"1.0","method":"call","params":{"model":"res.company.ldap.operator","method":"query","args":["","",""],"kwargs":{}}}`

**Expected behavior**
Methods should not be available via JSON RPC API but only from other Python classes.

**Additional context**
See https://www.odoo.com/documentation/17.0/developer/reference/backend/security.html#unsafe-public-methods

**Attempted fix**
A possible fix for this vulnerability is now in https://github.com/OCA/server-auth/pull/659 (https://github.com/OCA/server-auth/pull/659/commits/62d064f07f4a1d87fb949d1ff6dad9df6139cedd).

oh2fih:
Fix https://github.com/OCA/server-auth/issues/617

res.company.ldap.operator operators should be private methods; public methods allow arbitrary LDAP queries via JSON-API

### DIFF
--- a/users_ldap_groups/models/res_company_ldap_operator.py
+++ b/users_ldap_groups/models/res_company_ldap_operator.py
@@ -20,17 +20,17 @@ class ResCompanyLdapOperator(models.AbstractModel):
         """Return names of function to call on this model as operator"""
         return ("contains", "equals", "query")
 
-    def contains(self, ldap_entry, mapping):
+    def _contains(self, ldap_entry, mapping):
         return mapping.ldap_attribute in ldap_entry[1] and mapping.value in map(
             lambda x: x.decode(), ldap_entry[1][mapping.ldap_attribute]
         )
 
-    def equals(self, ldap_entry, mapping):
+    def _equals(self, ldap_entry, mapping):
         return mapping.ldap_attribute in ldap_entry[1] and mapping.value == str(
             list(map(lambda x: x.decode(), ldap_entry[1][mapping.ldap_attribute]))
         )
 
-    def query(self, ldap_entry, mapping):
+    def _query(self, ldap_entry, mapping):
         query_string = Template(mapping.value).safe_substitute(
             {attr: ldap_entry[1][attr][0].decode() for attr in ldap_entry[1]}
         )
--- a/users_ldap_groups/models/res_company_ldap_operator.py
+++ b/users_ldap_groups/models/res_company_ldap_operator.py
@@ -17,7 +17,7 @@ class ResCompanyLdapOperator(models.AbstractModel):
 
     @api.model
     def operators(self):
-        """Return names of function to call on this model as operator"""
+        """Return names (without '_') of function to call on this model as operator"""
         return ("contains", "equals", "query")
 
     def _contains(self, ldap_entry, mapping):
--- a/users_ldap_groups/models/res_company_ldap.py
+++ b/users_ldap_groups/models/res_company_ldap.py
@@ -48,7 +48,7 @@ class ResCompanyLdap(models.Model):
             _logger.debug("deleting all groups from user %d", user_id)
             groups.append((5, False, False))
         for mapping in this.group_mapping_ids:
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Please do a responsible disclosure of any vulnerability. You can't put `JSON RPC vulnerability` in big in the title to call those who want to exploit it.

oh2fih:
This information has been out there since Feb 23 as #617 and this fix available since Mar 30 as a part of #596, though...

pedrobaeza:
Yes, and that was also a non responsible disclosure. It's not your fault though, as there's no mechanism established for this at OCA (cc @vdewulf), as there's for Odoo: https://www.odoo.com/es_ES/security-report

oh2fih:
I agree. Well, https://odoo-community.org/contactus does have security contact information, but the information is a few clicks away from the repository and, therefore, slightly hard to find. Now that I have it, I'll definitely approach the security team first in the future.

sbidoul:
> I agree. Well, https://odoo-community.org/contactus does have security contact information, but the information is a few clicks away from the repository and, therefore, slightly hard to find. Now that I have it, I'll definitely approach the security team first in the future.

Good point, @oh2fih. I added a security policy on our GitHub organization. It is now visible on each repo.

users_ldap_groups/models/res_company_ldap_operator.py
```diff
@@ -20,17 +20,17 @@ def operators(self):
         """Return names of function to call on this model as operator"""
         return ("contains", "equals", "query")
```
hparfr, line 21: why not: 

```suggestion
        return ("_contains", "_equals", "_query")
```
to be constitent with the docstring ?

or adapt this docstring (l20)

hparfr (commented):
minor remark

users_ldap_groups/models/res_company_ldap_operator.py
oh2fih: Those are the operators and the `_` simply makes the function private. The operator is a bit different concept despite having a corresponding operator, isn't it?

This function is called in `class ResCompanyLdapGroupMapping` for `fields.Selection()`. Changing the format of the operator naming would affect on how the value is saved in the database and, therefore, brake every existing setting on instances updating the module.  You could suggest a better docstring, though.

etobella (approved):

users_ldap_groups/models/res_company_ldap_operator.py
```diff
@@ -20,17 +20,17 @@ def operators(self):
         """Return names of function to call on this model as operator"""
```
hparfr:
```suggestion
        """Return names (without '_') of function to call on this model as operator"""
```

hparfr (changes requested):

users_ldap_groups/models/res_company_ldap_operator.py
hparfr: Thanks for the explaination 
I just suggested a docstring change

oh2fih: Thanks! That seems ok even without further explanations on the docstring.

hparfr (approved):

pedrobaeza:
Let's merge it although the CI is red.

sbidoul:
I opened backport prs for 15, 14, 13. 

12 will require manual invervention.

PR #665, opened 2024-06-11
[16.0][FIX] vault: Stop propagation when clicking "enter"

CarlosRoca13:
Before this changes, when the password was required some other click events can be thrown.

To reproduce this problem:

1. Create a vault with an entry and a field
2. Clean session coockies
3. Without entering the private key, go to try change the field value
4. Set the private key and press enter

The system will ask for the private key once again.

cc @Tecnativa TT49556

ping @fkantelberg @pedrobaeza

### DIFF
--- a/vault/static/src/common/utils.esm.js
+++ b/vault/static/src/common/utils.esm.js
@@ -191,7 +191,8 @@ function askpass(title, options = {}) {
                 {
                     text: _t("Enter"),
                     classes: "btn-primary",
-                    click: async function () {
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Isn't the same problem in v15?

fkantelberg (approved):
I can't fully reproduce it now but had it some times in the past and the patch sounds reasonable.

Both recent issues could be cherry picked down to 13.0.

PR #666, opened 2024-06-11
[14.0][IMP] auth_oidc: Logout the user from the auth provider

wluyima:
Added a mechanism to logout the user from the auth provider when

### DIFF
--- a/auth_oidc/controllers/main.py
+++ b/auth_oidc/controllers/main.py
@@ -6,10 +6,15 @@ import base64
 import hashlib
 import logging
 import secrets
+from urllib.parse import parse_qs, urljoin, urlparse
 
 from werkzeug.urls import url_decode, url_encode
 
+from odoo import http
+from odoo.http import request
+
 from odoo.addons.auth_oauth.controllers.main import OAuthLogin
+from odoo.addons.web.controllers.main import Session
 
 _logger = logging.getLogger(__name__)
 
@@ -48,3 +53,26 @@ class OpenIDLogin(OAuthLogin):
                     provider["auth_endpoint"], url_encode(params)
                 )
         return providers
+
+
+class OpenIDLogout(Session):
+    @http.route("/web/session/logout", type="http", auth="none")
+    def logout(self, redirect="/web/login"):
+        # https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            provider = (
+                request.env["auth.oauth.provider"]
+                .sudo()
+                .browse(user.oauth_provider_id.id)
+            )
+            if provider.end_session_endpoint:
+                redirect_url = urljoin(request.httprequest.url_root, redirect)
+                components = urlparse(provider.end_session_endpoint)
+                params = parse_qs(components.query)
+                params["client_id"] = provider.client_id
+                params["post_logout_redirect_uri"] = redirect_url
+                logout_url = components._replace(query=url_encode(params)).geturl()
+                return super().logout(redirect=logout_url)
+        # User has no account with any provider or no logout URL is configured for the provider
+        return super().logout(redirect=redirect)
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -46,6 +46,12 @@ class AuthOauthProvider(models.Model):
         string="Token URL", help="Required for OpenID Connect authorization code flow."
     )
     jwks_uri = fields.Char(string="JWKS URL", help="Required for OpenID Connect.")
+    end_session_endpoint = fields.Char(
+        string="End Session URL",
+        help="If set, the user is logged out in the authorization provider upon logout "
+        "in the client, should be the value of end_session_endpoint specified by "
+        "the authorization provider.",
+    )
 
     @tools.ormcache("self.jwks_uri", "kid")
     def _get_keys(self, kid):
--- a/auth_oidc/tests/test_auth_oidc_logout.py
+++ b/auth_oidc/tests/test_auth_oidc_logout.py
@@ -1,0 +1,130 @@
+# Copyright 2021 ACSONE SA/NV <https://acsone.eu>
+# License: AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+import contextlib
+import logging
+from urllib.parse import parse_qsl, urljoin, urlparse
+
+from werkzeug.urls import url_encode
+
+import odoo
+from odoo.tests import common
+from odoo.tools.misc import DotDict
+
+from odoo.addons.website.tools import MockRequest
+
+from ..controllers.main import OpenIDLogout
+
+BASE_URL = "http://localhost:%s" % odoo.tools.config["http_port"]
+CLIENT_ID = "auth_oidc-test"
+LOGIN_PATH = "/web/login"
+OIDC_BASE_LOGOUT_URL = "http://keycloak"
+OIDC_LOGOUT_PATH = "/logout"
+
+logger = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def create_request(env, user_id, user_logout_func):
+    with MockRequest(env) as request:
+        request.httprequest.url_root = BASE_URL + "/"
+        request.session = DotDict(uid=user_id, logout=user_logout_func)
+        yield request
+
+
+class TestOpenIDLogout(common.HttpCase):
+    @staticmethod
+    def mock_logout_user(keep_db):
+        logger.info("Logging out user in Odoo")
+
+    def setUp(self):
+        super().setUp()
+        # search our test provider and bind the demo user to it
+        self.provider = self.env["auth.oauth.provider"].search(
+            [("client_id", "=", CLIENT_ID)]
+        )
+        self.assertEqual(len(self.provider), 1)
+
+    def _prepare_login_test_user(self, provider_id):
+        user = self.env.ref("base.user_demo")
+        user.write({"oauth_provider_id": provider_id, "oauth_uid": user.login})
+        return user
+
+    def _set_test_oidc_logout_url(self, end_session_endpoint):
+        self.provider.write({"end_session_endpoint": end_session_endpoint})
+
+    def test_skip_oidc_logout_for_user(self):
+        """Test that oidc logout is skipped if user is not associated to a provider"""
+        user = self._prepare_login_test_user(None)
+        with create_request(self.env, user.id, self.mock_logout_user):
+            resp = OpenIDLogout().logout()
+            self.assertEqual(LOGIN_PATH, resp.location)
+
+    def test_skip_oidc_logout_for_all_users(self):
+        """Test that oidc logout is skipped for all users if provider has no logout url"""
+        self.assertFalse(self.provider.end_session_endpoint)
+        user = self._prepare_login_test_user(self.provider)
+        with create_request(self.env, user.id, self.mock_logout_user):
+            resp = OpenIDLogout().logout()
+            self.assertEqual(LOGIN_PATH, resp.location)
+
+    def test_oidc_logout(self):
+        """Test that oidc logout"""
+        self._set_test_oidc_logout_url(urljoin(OIDC_BASE_LOGOUT_URL, OIDC_LOGOUT_PATH))
+        user = self._prepare_login_test_user(self.provider)
+        with create_request(self.env, user.id, self.mock_logout_user):
+            resp = OpenIDLogout().logout()
+            self.assertTrue(resp.location.startswith(OIDC_BASE_LOGOUT_URL))
+            actual_components = urlparse(resp.location)
+            self.assertEqual(OIDC_LOGOUT_PATH, actual_components.path)
+            actual_params = dict(parse_qsl(actual_components.query))
+            self.assertEqual(CLIENT_ID, actual_params["client_id"])
+            self.assertEqual(
+                urljoin(BASE_URL, LOGIN_PATH), actual_params["post_logout_redirect_uri"]
+            )
+
+    def test_oidc_logout_with_params(self):
+        """Test that params both in the logout and redirect urls are preserved"""
+        logout_url_params = {"param_1": 1, "param_2": 2}
+        oidc_logout_path = "{}?{}".format(
+            OIDC_LOGOUT_PATH, url_encode(logout_url_params)
+        )
+        logout_url = urljoin(OIDC_BASE_LOGOUT_URL, oidc_logout_path)
+        self._set_test_oidc_logout_url(logout_url)
+        user = self._prepare_login_test_user(self.provider)
+        with create_request(self.env, user.id, self.mock_logout_user):
+            redirect_path = "{}?{}".format(
+                LOGIN_PATH, url_encode({"param_3": 3, "param_4": 4})
+            )
+            params = {}
+            params.update(logout_url_params)
+            params["client_id"] = CLIENT_ID
+            params["post_logout_redirect_uri"] = urljoin(BASE_URL, redirect_path)
+            resp = OpenIDLogout().logout(redirect_path)
+            self.assertTrue(resp.location.startswith(OIDC_BASE_LOGOUT_URL))
+            actual_components = urlparse(resp.location)
+            self.assertEqual(OIDC_LOGOUT_PATH, actual_components.path)
+            actual_params = dict(parse_qsl(actual_components.query))
+            self.assertEqual(CLIENT_ID, actual_params["client_id"])
+            self.assertEqual("1", actual_params["param_1"])
+            self.assertEqual("2", actual_params["param_2"])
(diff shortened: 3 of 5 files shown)


auth_oidc/controllers/main.py
```diff
@@ -48,3 +51,18 @@ def list_providers(self):
...
+class OpenIDLogout(Session):
+
+    @http.route("/web/session/logout", type="http", auth="none")
+    def logout(self, redirect="/web/login"):
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            p = request.env["auth.oauth.provider"].sudo().browse(user.oauth_provider_id.id)
+            if p.logout_endpoint:
+                r = request.env['ir.config_parameter'].sudo().get_param('web.base.url') + redirect
+                logout_url = p.logout_endpoint + "?client_id=" + p.client_id + "&post_logout_redirect_uri=" + r
```
ap-wtioit: the logout_endpoint may already contain query parameters:
> 
    end_session_endpoint
        REQUIRED. URL at the OP to which an RP can perform a redirect to request that the End-User be logged out at the OP. This URL MUST use the https scheme and MAY contain port, path, and query parameter components. 

https://openid.net/specs/openid-connect-rpinitiated-1_0.html

So it would be better to not add a hardcocded questionmark here but use a more dynamic solution to adding parameters.

Also it seems suspicious that the redirect url `r` is not encoded here. (But i have not tested it yet)

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -46,6 +46,10 @@ class AuthOauthProvider(models.Model):
         string="Token URL", help="Required for OpenID Connect authorization code flow."
     )
     jwks_uri = fields.Char(string="JWKS URL", help="Required for OpenID Connect.")
+    logout_endpoint = fields.Char(
+        string="Logout URL",
+        help="Required for OpenID Connect to logout the user in the authorization provider upon logout in the client."
```
ap-wtioit: I guess this is expected to be the same value as `end_session_endpoint` from the provider discovery metadata, can you add a remark to the documentation that this is the value to be used.

ap-wtioit (changes requested):
Thanks for your efforts to make this module better. Could you run the OCA pre-commit or incoperate the suggestions from the log output: https://github.com/OCA/server-auth/actions/runs/9479497854/job/26118147534?pr=666

auth_oidc/controllers/main.py
ap-wtioit: Yes `r` should be encoded when added as post_logout_redirect_uri. (e.g. when redirect is `/web/login?db=devel&something=different` it would possibly end up as parameter something on the logout_endpoint)

wluyima: Makes sense, I will update the PR

auth_oidc/models/auth_oauth_provider.py
wluyima: Sounds good, I could also rename the property to `end_session_endpoint`

auth_oidc/controllers/main.py
```diff
@@ -48,3 +51,26 @@ def list_providers(self):
...
+
+class OpenIDLogout(Session):
+
+    @http.route("/web/session/logout", type="http", auth="none")
+    def logout(self, redirect="/web/login"):
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            p = request.env["auth.oauth.provider"].sudo().browse(user.oauth_provider_id.id)
+            if p.logout_endpoint:
+                r = request.env['ir.config_parameter'].sudo().get_param('web.base.url') + redirect
```
wluyima: @ap-wtioit is redirect expected to always be relative to the base URL? If full URLs are supported then the line above would need to be updated to be optional if redirect starts with `http` or `https`

ap-wtioit: i think only relative redirects are used in odoo but there is no technical limit for that and the redirect paramter is sent directly to werkzeug: https://github.com/odoo/odoo/blob/14.0/addons/web/controllers/main.py#L1319

so yes that's a good catch and web.base.url should only be used when the url is not starting with a protocol (i guess http(s) will suffice), also there is the possibility that the redirect is relative to /web/session/logout (e.g. `../login`, or `destroy`) but i think this is also not used currently (and i tested it only locally for checking security)

wluyima: You mentioned `destroy`, I see it also logs out the user, do we need to override it too to logout the user in the IDP?

ap-wtioit: i can only see `destroy` being used in one Odoo Enterprise application "project_timesheet_synchro / Synchronization with the external timesheet application" and that code upon login does not support the things added by auth_oidc so i would skip this usecase here.

wluyima: So is this good to go?

ap-wtioit: I'd like to test it again (most likely tomorrow) but yes this looks good to me now.

auth_oidc/controllers/main.py
```diff
@@ -48,3 +51,28 @@ def list_providers(self):
...
+class OpenIDLogout(Session):
+
+    @http.route("/web/session/logout", type="http", auth="none")
+    def logout(self, redirect="/web/login"):
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            p = request.env["auth.oauth.provider"].sudo().browse(user.oauth_provider_id.id)
+            if p.logout_endpoint:
+                r = redirect
+                if r.find('http') == -1 and r.find('https') == -1:
```
wluyima: I have sneaked in support for full URLs

ap-wtioit: Please use `not r.startswith("http:") and not r.startswith("https:")` here. the check you used would skip adding the base url for all redirects containing http.

ap-wtioit:
Reviewed again, did you see my comment regarding oca-precommit? The check is still failing.

wluyima:
@ap-wtioit when do you plan to have this merged?

auth_oidc/controllers/main.py
wluyima: Done

ap-wtioit (approved):

ap-wtioit:
> @ap-wtioit when do you plan to have this merged?

to me it looks good now. to get it merged we need a second approval, ideally from a PSC (Project Steering Committee member) or the maintainer of the module. then it can be merged (by the maintainer or PSC).

wluyima:
@sbidoul do you mind becoming the second reviewer/approver on this PR?

sbidoul:
I will look into it. Not had time yet.

wluyima:
I see that this was added to the 14.0 milestone, it would be great to forward port the changes to later milestones too i.e. 15, 16 and 17

wtaferner (approved):

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -46,6 +46,12 @@ class AuthOauthProvider(models.Model):
         string="Token URL", help="Required for OpenID Connect authorization code flow."
     )
     jwks_uri = fields.Char(string="JWKS URL", help="Required for OpenID Connect.")
+    logout_endpoint = fields.Char(
```
sbidoul: I'd suggest naming this field `end_session_endpoint` to match the field in the openid configuration document. This will make the code easier to understand.

auth_oidc/controllers/main.py
```diff
@@ -48,3 +52,36 @@ def list_providers(self):
...
+                .browse(user.oauth_provider_id.id)
+            )
+            if p.logout_endpoint:
+                r = redirect
+                if not r.startswith("http:") and not r.startswith("https:"):
+                    r = (
+                        request.env["ir.config_parameter"]
+                        .sudo()
+                        .get_param("web.base.url")
+                        + r
```
sbidoul: Let's use `urllib.parse.urljoin` instead of `+` here. It will be better in case `web.base.url` ends with a `/`.

auth_oidc/controllers/main.py
```diff
@@ -48,3 +52,36 @@ def list_providers(self):
...
+                    )
+                logout_base_url = p.logout_endpoint
+                params = {}
+                if "?" in p.logout_endpoint:
+                    url_parts = p.logout_endpoint.split("?")
```
sbidoul: I'm not too comfortable with parsing a URL this way. What if, for instance, there is more than 1 `?` in the url. It *should* not happen of course, but... It would be more robust to use `urllib.parse.urlparse` and `parse_qs`.

auth_oidc/controllers/main.py
```diff
@@ -48,3 +52,36 @@ def list_providers(self):
...
+                    logout_base_url = url_parts[0]
+                    params = url_decode(url_parts[1])
+                params["client_id"] = p.client_id
+                params["post_logout_redirect_uri"] = r
+                logout_url = f"{logout_base_url}?{url_encode(params)}"
```
sbidoul: It might be nice to extract this url manipulation logic in a little class method that could be unit tested.

auth_oidc/controllers/main.py
```diff
@@ -48,3 +52,36 @@ def list_providers(self):
...
                 )
         return providers
+
+
+class OpenIDLogout(Session):
+    @http.route("/web/session/logout", type="http", auth="none")
+    def logout(self, redirect="/web/login"):
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            p = (
```
sbidoul: A little detail: let's use explicit variable names?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -46,6 +46,12 @@ class AuthOauthProvider(models.Model):
...
+        string="Logout URL",
+        help="Required for OpenID Connect to logout the user in the authorization provider "
+        "upon logout in the client, should be the value of end_session_endpoint specified by "
+        "the authorization provider",
```
sbidoul: Shall we mention that when this field is set, logging out of Odoo also causes a complete logout of the identify provider?

auth_oidc/controllers/main.py
```diff
@@ -48,3 +52,36 @@ def list_providers(self):
                     provider["auth_endpoint"], url_encode(params)
                 )
         return providers
+
+
+class OpenIDLogout(Session):
+    @http.route("/web/session/logout", type="http", auth="none")
+    def logout(self, redirect="/web/login"):
```
sbidoul, line 60: Can you add a comment or docstring referencing the spec that this implements? 
https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

sbidoul (changes requested):
Hi,

Thanks for this contrib.

I made a few comments, the most important being the URL parsing logic that should use urllib.parse for robustness. I think this is important is such a security sensitive module.

Could you also add a test? The existing tests should give inspiration.

Head up about forward porting: I tested this on 16.0 and it will require some work to adapt, because Odoo now forces a local redirect (requests.redirect has a local=True argument):

https://github.com/odoo/odoo/blob/e889bd2c993a456d3077beaa3271f73e3510665e/addons/web/controllers/session.py#L85-L87

auth_oidc/controllers/main.py
wluyima: Is there documentation on how to run the test suite?

sbidoul: @wluyima This [part of the odoo documentation](https://www.odoo.com/documentation/17.0/developer/reference/backend/testing.html?highlight=test%20suite#running-tests) should get you started. Let me know if you hit any issue.

wluyima: Thanks @sbidoul is there a way I can find documentation to run Odoo 14.0 locally from source? I was following the [15.0 documentation](https://www.odoo.com/documentation/15.0/administration/on_premise/source.html) and I can't seem to install the packages in the requirements.txt file, I get [this error](https://dpaste.com/8A7SVGMGA) when I try to install the packages, do you have any clues? I'm using Python **3.12.4**. I'm not sure if is not compatible, is there documentation on Odoo and Python versions compatibility?

sbidoul: For Odoo 14 I generally use python 3.8. It will for sure not work with 3.12.

I recommend installing the Odoo dependencies in a virtual environment:

```
python3.8 -m venv ./venv
source ./venv/bin/activate
pip install -r https://raw.githubusercontent.com/odoo/odoo/14.0/requirements.txt
```

But if you don't use Odoo 14, why do you do this PR to the 14.0 branch?

wluyima: Thanks @sbidoul! I did say we're using Odoo 14 which is why this PR is issued against the 14.0 branch and I'm installing the dependencies in a virtual environment but I was using python 3.12, I'll try with python 3.8 and see if it works. I was actually able to later find the documentation for Odoo 14 [here](https://www.odoo.com/documentation/14.0/administration/install/source.html).

auth_oidc/controllers/main.py
```diff
@@ -48,3 +53,33 @@ def list_providers(self):
...
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            provider = (
+                request.env["auth.oauth.provider"]
+                .sudo()
+                .browse(user.oauth_provider_id.id)
+            )
+            if provider.end_session_endpoint:
+                red_url = redirect
+                if not red_url.startswith("http:") and not red_url.startswith("https:"):
```
sbidoul: This test is now unnecessary, I think, because urljoin will do the right thing.

auth_oidc/controllers/main.py
```diff
@@ -48,3 +53,33 @@ def list_providers(self):
...
+        # https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            provider = (
+                request.env["auth.oauth.provider"]
+                .sudo()
+                .browse(user.oauth_provider_id.id)
+            )
+            if provider.end_session_endpoint:
+                red_url = redirect
```
sbidoul:
```suggestion
                post_logout_redirect_uri = redirect
```

wluyima:
@sbidoul tests have been added

wluyima:
@sbidoul @ap-wtioit tests were added, did you get a chance to take a look?

gbrito (approved):

auth_oidc/controllers/main.py
```diff
@@ -48,3 +53,26 @@ def list_providers(self):
...
+        # https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout
+        user = request.env["res.users"].sudo().browse(request.session.uid)
+        if user.oauth_provider_id.id:
+            provider = (
+                request.env["auth.oauth.provider"]
+                .sudo()
+                .browse(user.oauth_provider_id.id)
+            )
+            if provider.end_session_endpoint:
+                red_url = urljoin(request.httprequest.url_root, redirect)
```
sbidoul: When I see `red_url` my brain wonders why that url is not blue ;)
Let's not abuse abbreviations.

auth_oidc/models/auth_oauth_provider.py
sbidoul: Can you also finish that sentence with a `.` as for other help texts.

auth_oidc/tests/test_auth_oidc_logout.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_skip_oidc_logout_for_user(self):
+        """Test that oidc logout is skipped if user is not associated to a provider"""
+        user = self._prepare_login_test_user(None)
+        with create_request(self.env, user.id, self.mock_logout_user):
+            resp = OpenIDLogout().logout()
+            self.assertEqual(LOGIN_PATH, resp.location)
+
+    def test_skip_oidc_logout_for_all_users(self):
+        """Test that oidc logout is skipped for all users if provider has no logout url"""
+        self.assertFalse(None, self.provider.end_session_endpoint)
```
sbidoul: This assert is useless as None is always falsy, and if you remove it, this test looks identical to the previous one?

sbidoul (commented):
A couple of minor remarks, otherwise looks good now.

Can you squash commits?

auth_oidc/tests/test_auth_oidc_logout.py
wluyima: Good catch! It was an oversight.

The tests are not identical, the previous test is for the scenario where a user is not associated to a provider, while this test is for the scenario where no end_session_endpoint is configured for the provider, the difference is on lines 58 and 66.

wluyima:
Done and squashed commits

sbidoul (approved):
Thanks!

wluyima:
Hi @ap-wtioit, is this waiting on anything to be merged?

ap-wtioit:
@wluyima a maintainer needs to merge it, preferably @sbidoul. but since he approved it already maybe he just forgot.

sbidoul:
Indeed.

/ocabot merge minor

Thanks for the work @wluyima

PR #671, opened 2024-07-23
[17.0][REM] remove bad pandoc .deb package at the repo root

rvalyi:
this pandoc .deb package was added in the root of the 17.0 branch by mistake by @CRogos in https://github.com/OCA/server-auth/pull/641

cc @CRogos @pedrobaeza


pedrobaeza:
Doing this is worse, as there's 60 MB of diff instead of 30 MB. The solution is to forced push the branch. I plan to do it massively next week.

PR #679, opened 2024-08-20
[16.0][FIX] vault_share: Button to share value not showed and not working

CarlosRoca13:
To show the button up, it is requiring a getter that is not defined on vault field. So the getter is added and make the button works correctly.

Before:
image

After:
image

Steps to reproduce the problem:

1. Go to Vault
2. Enter/Create a Vault
3. Press on entries
4. Access/Create a entry with a field

The button with external link is not showed

cc @Tecnativa TT50557

ping @pedrobaeza @chienandalu @fkantelberg


chienandalu (approved):
LGTM :+1:

pedrobaeza (approved):
/ocabot merge patch

PR #680, opened 2024-08-23
[17.0][MIG] impersonate_login

astirpe:
Porting `impersonate_login` to V17.

Differences compared to original V14 module:

- when switching to different user, redirect to main menu to avoid possible access rights error
- on the top-right menu, the button `Back To Origin Login` is displayed in case the current user is being impersonated
- method `impersonate_login()` can call `back_to_origin_login()` in case the impersonated user is trying to impersonate the original user back
- the `def _create()` and `def write()` are a bit revised to increase efficiency and robustness
- added tests


impersonate_login/models/mail_message.py
```diff
@@ -0,0 +1,51 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.http import request
+
+
+class Message(models.Model):
+    _inherit = "mail.message"
+
+    impersonated_author_id = fields.Many2one(
```
StefanRijnhart, line 13: Please precreate the column in the pre-init hook, so that the compute method is not run unnecesssarily on what might be a huge amount of messages.

impersonate_login/models/mail_message.py
```diff
@@ -0,0 +1,51 @@
...
+            additional_info = ""
+            if (
+                request
+                and request.session.impersonate_from_uid
+                and rec.impersonated_author_id
+            ):
+                current_partner = (
+                    self.env["res.users"].browse(request.session.uid).partner_id
+                )
+                additional_info = _(f"Logged as {current_partner.name}")
```
StefanRijnhart: The string substitution has to take place before the translation. Also, maybe apply markupsafe.escape to the substituted value.

impersonate_login/models/model.py
```diff
@@ -0,0 +1,30 @@
...
+                vals["create_uid"] = request.session.impersonate_from_uid
+        return super()._create(data_list)
+
+    def write(self, vals):
+        if (
+            request
+            and request.session.impersonate_from_uid
+            and "write_uid" in self._fields
+        ):
+            vals["write_uid"] = request.session.impersonate_from_uid
```
StefanRijnhart: You could consider putting the same in a similar overwrite for `_write` (https://github.com/odoo/odoo/blob/71b6eee/odoo/models.py#L4492)

impersonate_login/models/model.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @api.model
+    def _create(self, data_list):
+        if (
+            request
+            and request.session.impersonate_from_uid
+            and "create_uid" in self._fields
+        ):
+            for vals in data_list:
+                vals["create_uid"] = request.session.impersonate_from_uid
```
StefanRijnhart, line 20: It would be nice to have a test to check the create_uid and write_uid of records created during an impersonated session.

impersonate_login/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,6 @@
+This module allows to login as another user. In the chatter, the user
+who is logged as another user is displayed. The mails and messages are
+sent from the orignal user. A table diplays the impersonated logins in
+technical. The user can return to his own user by clicking on the button
+"Return to my user". This module is very useful for the support team. An
+alternative module will be auth_admin_passkey.
```
StefanRijnhart: I think it would be useful to emphasize that this module takes all the measures below for auditing and governance purpose. Something like the following:

```suggestion
This module allows one user (for example, a member of the support team) to log in as another user. The impersonation session can be exited by clicking on the button "Back to Original User".

To ensure that any abuse of this feature will not go unnoticed, the following measures are in place:

* In the chatter, it is displayed who is the user that is logged as another user.
* Mails and messages are sent from the original user.
* Impersonated logins are logged and can be consulted through the Settings -> Technical menu.

There is an alternative module to allow logins as another user (auth_admin_passkey), but it does not support these security mechanisms.
```

impersonate_login/models/mail_message.py
```diff
@@ -0,0 +1,51 @@
...
+        store=True,
+    )
+
+    body = fields.Html(
+        compute="_compute_message_body",
+        store=True,
+    )
+
+    @api.depends("author_id")
+    def _compute_impersonated_author_id(self):
```
StefanRijnhart, line 25: This works for notes, but tracking messages still have the impersonated author:

image

Actually, maybe the create_uid overwrite does not work. I am logged in as demo, impersonating a user called demo2. I created a crm.lead, but demo2 is still mentioned everywhere:

image

impersonate_login/models/mail_message.py
StefanRijnhart:
```suggestion
                additional_info = _(f"Logged in as {current_partner.name}")
```

impersonate_login/readme/USAGE.md
```diff
@@ -0,0 +1,2 @@
+1.  On the top right corner or in the res.users list, click "Switch Login" to impersonate another user
```
StefanRijnhart:
```suggestion
1.  In the menu that is displayed when clicking on the user avatar on the top right corner, or in the res.users list, click "Switch Login" to impersonate another user.
```

impersonate_login/README.rst
```diff
@@ -0,0 +1,100 @@
...
+technical. The user can return to his own user by clicking on the button
+"Return to my user". This module is very useful for the support team. An
+alternative module will be auth_admin_passkey.
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
```
StefanRijnhart, line 58: Please add a `readme/CONFIGURE.rst` file that mentions the Impersonate Logins user group.

impersonate_login/models/res_users.py
```diff
@@ -0,0 +1,118 @@
...
+                request.session.session_token = security.compute_session_token(
+                    request.session, request.env
+                )
+                logger.info(
+                    f"IMPERSONATE: {self._get_partner_name(from_uid)} "
+                    f"Logout as {self._get_partner_name(self._uid)}"
+                )
+            return {
+                "type": "ir.actions.client",
+                "tag": "reload",
```
StefanRijnhart, line 118: It is possible to get an access error when going back to the original user as well! Would you consider redirecting to the main menu here as well? 

image

StefanRijnhart (changes requested):

impersonate_login/models/model.py
astirpe: The `_write` takes the vals as frozendict, so it requires some workarounds to make it working. Is it ok to keep it as it is?

impersonate_login/models/model.py
astirpe: Tests added, thanks!

impersonate_login/README.rst
astirpe: CONFIGURE.rst added

impersonate_login/models/res_users.py
astirpe: Done!

impersonate_login/models/mail_message.py
astirpe: The tracking messages should be fixed now, would you check?

Regarding the create_uid, it seems to me correct that demo2 is mentioned there. Am I missing something?

StefanRijnhart: isn't it the case that the code attempts to inject the original user as the create_uid? Demo2 is the impersonated user.

impersonate_login/models/model.py
StefanRijnhart: OK

impersonate_login/models/mail_message.py
StefanRijnhart: The note message now also has the create_uid and author_id of the impersonated user. I would expect both messages to have the create_uid of the original user.

image

impersonate_login/readme/CONFIGURE.md
```diff
@@ -0,0 +1 @@
+Both impersonating and impersonated users must belong to group "Impersonate Users".
```
StefanRijnhart: No, only the impersonating user.

impersonate_login/models/mail_message.py
astirpe: Ok sorry, I misunderstood your initial request. Reverted.

The note message is telling us: I'm Marc Demo (original user) and I'm impersonating Demo2 (impersonated user).
Do you mean that, when creating the crm.lead, the create_uid should say I'm Marc Demo (not Demo2)? Otherwise it seems correct to me.

impersonate_login/readme/CONFIGURE.md
astirpe: Right! Updated.

impersonate_login/models/mail_message.py
astirpe: OK, I see the problem. Working on it.

astirpe: @StefanRijnhart 
With this commit https://github.com/OCA/server-auth/pull/680/commits/77f85f03eb504b319f500d2963cbdfe8bf4ccddb the create_uid should be solved. I was trying to do the same with the write_uid but I didn't manage to fix it.

StefanRijnhart: Indeed, it is not possible to override the write_uid in the write values, but it is possible to overwrite it using
```
    def write(self, vals):
        """Overwrite the write_uid with the impersonating user"""
        res = super().write(vals)
        if (
            request
            and request.session.impersonate_from_uid
            and "write_uid" in self._fields
        ):
            self._fields["write_uid"].write(self, request.session.impersonate_from_uid)
        return res
```

StefanRijnhart:
Can you catch the log output in the tests so that no warnings are logged?

```
2024-08-29 06:55:55,780 509855 WARNING 17imp odoo.http: It's you. 
2024-08-29 06:55:55,817 509855 WARNING 17imp odoo.http: You are already Logged as another user. 
```

impersonate_login/models/mail_message.py
astirpe: It Works! Thank you very much!

StefanRijnhart (approved):
Thanks for the updates, looking good!

CasVissers-360ERP (approved):
Functional review

PR #592, opened 2023-12-18
[17.0][MIG] auth_signup_verify_email

augusto-weiss:

PR #681, opened 2024-08-23
[17.0][MIG] auth_signup_verify_email: Migration to 17.0

pedrobaeza:
Resumed from #592, squashing some commits, removing offending pandoc*.deb file, and fixing pre-commit.

@Tecnativa TT50109


mac-adhoc:
@pedrobaeza would you mind to update the current Readme text by the following one in order to make ir clearer?

This module add an email verification process during new users Sign up process from /web/signup page, admiting "valid" email addresses only. In this verification process the module tries to send an invitation email to the new user with a unique access link which allows the new user to create a password (same email template than "reset password" mail).
This module also hides the "password" and "confirm password" fields from sign up form because the password would be set for the very first time when user clicks on the unique access link recieved by email.

Thanks in advanced!

pedrobaeza:
@mac-adhoc you can go to https://github.com/OCA/server-auth/blob/16.0/auth_signup_verify_email/readme/DESCRIPTION.rst, press `.` (the dot) to access to a VSCode environment, perform the modifications, and propose yourself a PR. Please give it a try to see if it's difficult.

PR #682, opened 2024-08-28
[17.0][IMP] auth_oidc: add groups' handling

OdyX:
This allows groups' handling via a token's attributes as passed by either a Keycloak or a Gitlab instance serving as IdP.

* This is the 17.0-version of #350, then #372 , by @26houseRobot and @hbrunn , with just minor fixes

@sbidoul : I'd be happy to make any necessary changes!

### DIFF
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -130,8 +130,12 @@ class AuthOauthProviderGroupLine(models.Model):
         for this in self:
             try:
                 this._eval_expression(self.env.user, {})
-            except (AttributeError, KeyError, NameError) as e:
-                raise exceptions.ValidationError("\n".join(e.args)) from e
+            except (AttributeError, KeyError, NameError, ValueError) as e:
+                # AttributeError: user object can be accessed via attributes: user.email
+                # KeyError: token is a dict of dicts
+                # NameError: only user and token can be used
+                # ValueError: for inexistant variables or attributes
+                raise exceptions.ValidationError(e) from e
 
     def _eval_expression(self, user, token):
         self.ensure_one()
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -2,12 +2,13 @@
 # Copyright 2021 ACSONE SA/NV <https://acsone.eu>
 # License: AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
+import collections
 import logging
 import secrets
 
 import requests
 
-from odoo import fields, models, tools
+from odoo import api, exceptions, fields, models, tools
 
 try:
     from jose import jwt
@@ -51,6 +52,11 @@ class AuthOauthProvider(models.Model):
         "For example: {'prompt':'select_account'}"
     )
     end_session_endpoint = fields.Char(string="End Session URL")
+    group_line_ids = fields.One2many(
+        "auth.oauth.provider.group_line",
+        "provider_id",
+        string="Group mappings",
+    )
 
     @tools.ormcache("self.jwks_uri", "kid")
     def _get_keys(self, kid):
@@ -109,3 +115,36 @@ class AuthOauthProvider(models.Model):
         if error:
             raise error
         return {}
+
+
+class AuthOauthProviderGroupLine(models.Model):
+    _name = "auth.oauth.provider.group_line"
+
+    provider_id = fields.Many2one("auth.oauth.provider", required=True)
+    group_id = fields.Many2one("res.groups", required=True)
+    expression = fields.Char(required=True, help="Variables: user, token")
+
+    @api.constrains("expression")
+    def _check_expression(self):
+        for this in self:
+            try:
+                this._eval_expression(self.env.user, {})
+            except (AttributeError, KeyError, NameError) as e:
+                raise exceptions.ValidationError("\n".join(e.args))
+
+    def _eval_expression(self, user, token):
+        self.ensure_one()
+
+        class Defaultdict2(collections.defaultdict):
+            """Class keeping yielding defaultdicts"""
+
+            def __init__(self, *args, **kwargs):
+                super().__init__(Defaultdict2, *args, **kwargs)
+
+        return tools.safe_eval.safe_eval(
+            self.expression,
+            {
+                "user": user,
+                "token": Defaultdict2(token),
+            },
+        )
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -119,6 +119,7 @@ class AuthOauthProvider(models.Model):
 
 class AuthOauthProviderGroupLine(models.Model):
     _name = "auth.oauth.provider.group_line"
+    _description = "OAuth mapping between an Odoo group and an expression"
 
     provider_id = fields.Many2one("auth.oauth.provider", required=True)
     group_id = fields.Many2one("res.groups", required=True)
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -131,7 +131,7 @@ class AuthOauthProviderGroupLine(models.Model):
             try:
                 this._eval_expression(self.env.user, {})
             except (AttributeError, KeyError, NameError) as e:
-                raise exceptions.ValidationError("\n".join(e.args))
+                raise exceptions.ValidationError("\n".join(e.args)) from e
 
     def _eval_expression(self, user, token):
         self.ensure_one()
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -69,6 +69,7 @@ class ResUsers(models.Model):
             data = requests.get(
                 oauth_provider.data_endpoint,
                 headers={"Authorization": "Bearer %s" % access_token},
+                timeout=10,
             ).json()
             validation.update(data)
         # required check
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -64,6 +64,12 @@ class ResUsers(models.Model):
             _logger.error("No id_token in response.")
             raise AccessDenied()
         validation = oauth_provider._parse_id_token(id_token, access_token)
+        if oauth_provider.data_endpoint:
+            data = requests.get(
+                oauth_provider.data_endpoint,
+                headers={"Authorization": "Bearer %s" % access_token},
+            ).json()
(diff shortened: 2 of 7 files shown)


hbrunn:
the v14 PR is based on my v12 PR which was merged - why didn't you just forward port this?

OdyX:
@hbrunn thanks for asking! As I'm quite fresh in the Odoo ecosystem, I did not see the v12 PR. Care to share a link?

As you can see from the code, my patch works a bit differently; as it appeared that what I needed for group mapping was directly in the access token, there's no usage of the data_endpoint. But I'm also likely not fluent enough in OAuth2 to know if that is really a correct way too.

Well; in any case, I'm happy to work towards merging either this or your v12 PR (or a mix of both) for v17. We need @sbidoul 's input, right?

hbrunn:
you find the v12 PR [here](https://github.com/OCA/server-auth/pull/350)

OdyX:
@hbrunn Great. Thanks for the pointer to the v12 PR. I've now understood the code much better, and did a mostly-straightforward port, with just two minor additions as separate commits. Could you perhaps review?

OdyX:
As the codecov warnings seem critical, I've now added some more tests around the safe_eval call of the expressions.

Edit: and now also added some groups' assignment/deassignment checks, pushing the codecov bar above the needed limits.

auth_oidc/models/res_users.py
```diff
@@ -80,3 +88,22 @@ def auth_oauth(self, provider, params):
...
+        login = super()._auth_oauth_signin(provider, validation, params)
+        user = self.search([("login", "=", login)])
+        if user:
+            group_updates = []
+            for group_line in (
+                self.env["auth.oauth.provider"].browse(provider).group_line_ids
+            ):
+                if group_line._eval_expression(user, validation):
+                    if group_line.group_id not in user.groups_id:
+                        group_updates.append((Command.LINK, group_line.group_id.id))
```
hbrunn:
```suggestion
                        group_updates.append(Command.link(group_line.group_id.id))
```

seems more canonical to me

auth_oidc/models/res_users.py
```diff
@@ -80,3 +88,22 @@ def auth_oauth(self, provider, params):
...
+                else:
+                    if group_line.group_id in user.groups_id:
+                        group_updates.append((Command.UNLINK, group_line.group_id.id))
```
hbrunn:
```suggestion
                        group_updates.append(Command.unlink(group_line.group_id.id))
```

hbrunn (approved):
thanks, just minor stylistic things

hbrunn:
@OdyX please rebase your branch. If you allow edits by maintainers, I could say

/ocabot rebase

and it would do this automatically

OdyX:
@hbrunn integrated your suggestions in https://github.com/OCA/server-auth/pull/682/commits/afd10ea74a6ad31bcb60c1b258fab2da95bc65e5; and I have rebased this on top of 17.0.

OdyX:
Rebased and pushed. It would be nice to have @sbidoul 's input and get this merged (or commented upon).

auth_oidc/models/res_users.py
```diff
@@ -64,6 +65,13 @@ def auth_oauth(self, provider, params):
             _logger.error("No id_token in response.")
             raise AccessDenied()
         validation = oauth_provider._parse_id_token(id_token, access_token)
+        if oauth_provider.data_endpoint:
+            data = requests.get(
+                oauth_provider.data_endpoint,
+                headers={"Authorization": "Bearer %s" % access_token},
+                timeout=10,
+            ).json()
```
sbidoul: Error handling for this request is missing?

sbidoul: I see there is a _auth_oauth_validate in auth_oauth. Does it make sense to reuse it / get inspiration from it ?

auth_oidc/models/res_users.py
```diff
@@ -80,3 +88,22 @@ def auth_oauth(self, provider, params):
...
+            ):
+                if group_line._eval_expression(user, validation):
+                    if group_line.group_id not in user.groups_id:
+                        group_updates.append(Command.link(group_line.group_id.id))
+                else:
+                    if group_line.group_id in user.groups_id:
+                        group_updates.append(Command.unlink(group_line.group_id.id))
+            if group_updates:
+                user.write({"groups_id": group_updates})
+        return login
```
sbidoul, line 115: For comparison, do we have other modules that do a similar thing ?

sbidoul (commented):
Sorry folks for the delay.

I've not had time to do a deep review yet, but I have a couple a questions from a first read.

auth_oidc/models/res_users.py
hbrunn: https://github.com/OCA/server-auth/tree/16.0/users_ldap_groups

auth_oidc/models/res_users.py
OdyX: I've pushed a refactoring of this part towards an _auth_oauth_validate function override. It's not possible to reuse it as-is, but at least it becomes really similar

OdyX:
@sbidoul Could you perhaps take another look?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -108,3 +114,39 @@ def _decode_id_token(self, access_token, id_token, kid):
...
+                # AttributeError: user object can be accessed via attributes: user.email
+                # KeyError: token is a dict of dicts
+                # NameError: only user and token can be used
+                # ValueError: for inexistant variables or attributes
+                raise exceptions.ValidationError(e) from e
+
+    def _eval_expression(self, user, token):
+        self.ensure_one()
+
+        class Defaultdict2(collections.defaultdict):
```
sbidoul, line 143: This comment that @hbrunn added in the original PR was helpful:

```suggestion
        class Defaultdict2(collections.defaultdict):
```
```suggestion
        class Defaultdict2(collections.defaultdict):
            """Class keeping yielding defaultdicts"""
```

auth_oidc/models/res_users.py
```diff
@@ -44,6 +45,45 @@ def _auth_oauth_get_tokens_auth_code_flow(self, oauth_provider, params):
         # https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse
         return response_json.get("access_token"), response_json.get("id_token")
 
+    @api.model
+    def _auth_oauth_validate(self, provider, id_token, access_token):
```
sbidoul: Does this completely replace the upstream `_auth_oauth_validate`? Why is this necessary?

OdyX: Yes. It's mostly a copy-paste from https://github.com/odoo/odoo/blob/17.0/addons/auth_oauth/models/res_users.py#L47-L72 here's the difference:

```diff
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -46,16 +46,19 @@ class ResUsers(models.Model):
         return response_json.get("access_token"), response_json.get("id_token")
 
     @api.model
-    def _auth_oauth_validate(self, provider, access_token):
-        """return the validation data corresponding to the access token"""
+    def _auth_oauth_validate(self, provider, id_token, access_token):
+        """
+        return the validation data corresponding to the access token
+        Mostly the same as auth_oauth ResUsers._auth_oauth_validate, minus the
+        validation_endpoint
+        """
         oauth_provider = self.env["auth.oauth.provider"].browse(provider)
-        validation = self._auth_oauth_rpc(
-            oauth_provider.validation_endpoint, access_token
-        )
-        if validation.get("error"):
-            raise Exception(validation["error"])
+
+        # Parse the token to get validation data
+        validation = oauth_provider._parse_id_token(id_token, access_token)
+
         if oauth_provider.data_endpoint:
-            data = self._auth_oauth_rpc(oauth_provider.data_endpoint, access_token)
+            data = super()._auth_oauth_rpc(oauth_provider.data_endpoint, access_token)
             validation.update(data)
         # unify subject key, pop all possible and get most sensible. When this
         # is reworked, BC should be dropped and only the `sub` key should be
@@ -75,7 +78,8 @@ class ResUsers(models.Model):
             None,
         )
         if not subject:
-            raise AccessDenied("Missing subject identity")
+            _logger.error("Access Denied: missing subject identity")
+            raise AccessDenied()
         validation["user_id"] = subject
 
         return validation
```

It seemed clearer this way:
* mimick what's done upstream in terms of methods
* rely on `super()._auth_oauth_rpc(oauth_provider.data_endpoint, access_token)` instead of doing a `requests.get` in the code removed below
* it's not clearer to parse the `id_token` before or separately.

I kept the two `_logger.error` + `raise AccessDenied` instead of copying upstream's, but I can certainly live with a closer copy.

Does this make it clearer?

I'm open to suggestions!

OdyX:
@sbidoul Force-pushed a rebase with the comment suggestion included in the first commit. (twice because I forgot about `pre-commit`).

sbidoul:
I'm not comfortable with replacing upstream method completely. Is there no other way? It is not clear to me why it is necessary for this PR but I have unfortunately no time to dig into this until October.

OdyX:
@sbidoul I managed to dive back into this code again. 

> I'm not comfortable with replacing upstream method completely. Is there no other way?

It turns out, there is. See https://github.com/OCA/server-auth/pull/682/commits/77c5c2d935906436d9c9f80853911ef4832b99b6 . In there, we just delete the duplicated code from `auth_oauth`, and merge the id_token and userinfo data into `validation`.

I'd be thrilled if we could merge this in 17.0, and I'd then be happy with forward porting these to 18.0 and 19.0 as needed.

(edited: sorry for the multiple force-pushes; tests turned out to be… relevant :-) )

OdyX:
Note for future me; KeyCloak needs the system parameter `auth_oauth.authorization_header` set to 1, see https://github.com/odoo/odoo/blob/17.0/addons/auth_oauth/models/res_users.py#L33

OdyX:
> I'm not comfortable with replacing upstream method completely. Is there no other way?

It turns out, after more tests and a (yet) better understanding of the code, doing so is actually *worse* than using the upstream code: [the "Authorization: Bearer" is behind a system-level toggle](https://github.com/odoo/odoo/blob/17.0/addons/auth_oauth/models/res_users.py#L33), which makes things unnecessarily confusing and hard to debug.

I finally went with just a copy of the response checker in https://github.com/OCA/server-auth/commit/89a2a5f2f2159c9fcaa6b505f72fd5c5951659f4 (and I could live without that last commit).

OdyX:
I would really love this being merged. As far as I can tell, this is ready and just missing validation from @sbidoul .

So please remove this `stale` label.

PR #683, opened 2024-08-29
[16.0][MIG] impersonate_login: backport from V17 to V16

astirpe:


StefanRijnhart (approved):
Thanks!

CasVissers-360ERP (approved):

CasVissers-360ERP:
Tested / functional review.

impersonate_login/models/impersonate_log.py
```diff
@@ -0,0 +1,25 @@
...
+
+from odoo import fields, models
+
+
+class ImpersonateLog(models.Model):
+    _name = "impersonate.log"
+    _description = "Impersonate Logs"
+
+    user_id = fields.Many2one(
+        comodel_name="res.partner",
```
Kev-Roche:
```suggestion
        comodel_name="res.users",
```

Kev-Roche (changes requested):
@astirpe, sincerely thank you to bring this module alive and fix it from the PR in V14 !
I have a question, in my original PR, I had css to visually show when we are logged with an impersonate user. Why not keep it ?

impersonate_login/models/impersonate_log.py
astirpe: Thank you!

astirpe:
@Kev-Roche Thank you!
Porting the css needed also the porting of the rest of the js logic, it was taking too much time to me. I was more focused to make this module just working in V17. But I think the css part is an important usability feature. Is it maybe something you can implement?

Kev-Roche (approved):

Kev-Roche:
> @Kev-Roche Thank you! Porting the css needed also the porting of the rest of the js logic, it was taking too much time to me. I was more focused to make this module just working in V17. But I think the css part is an important usability feature. Is it maybe something you can implement?

Ok, I'll work on this part.

rvalyi (approved):

PR #684, opened 2024-08-30
[17.0][FIX] impersonate_login: fix computed body field

astirpe:
Fix bug in compute method + computed body should be writable


astirpe:
@CasVissers-360ERP @StefanRijnhart

CasVissers-360ERP:
@dreispt @Kev-Roche I think this fix should be fast tracked. We missed this in our tests, the fixed bug has a big impact.

CasVissers-360ERP (approved):
Functional review

Kev-Roche (approved):

PR #686, opened 2024-09-01
[17.0][FIX] impersonate_login: fix user_id comodel_name

astirpe:


Kev-Roche (approved):

CasVissers-360ERP (approved):

rvalyi (approved):

impersonate_login/models/impersonate_log.py
```diff
@@ -11,7 +11,7 @@ class ImpersonateLog(models.Model):
     _description = "Impersonate Logs"
 
     user_id = fields.Many2one(
-        comodel_name="res.partner",
+        comodel_name="res.users",
```
simahawk, line 14: does it require any migration for existing records?

astirpe: @simahawk this is the fix already merged in V16 https://github.com/OCA/server-auth/commit/51eccf805efebcdd1f3c0141619a76a88991bcea

The integer written in `user_id` is a `res.users` anyway, so no need of migration

PR #690, opened 2024-09-18
[FIX] password_security: Fix reset password when user doesn't have password_write_date value.

keylor2906:
[FIX] password_security: Fix reset password when user doesn't have password_write_date value.

It happens that when you install the module, all the users that previously existed before that won't have a value set for the new field password_write_date. And when the validation of minimal hours has passed between the last password write date and now is done, it gets an error of unsupported operands between a bool and timedelta. With this change, it won't be necessary to validate the last password date if it's the first time reseting password, and when the password is changed it will correctly save the value on password_write_date and will be working normally afterwards.

image

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -144,6 +144,8 @@ class ResUsers(models.Model):
             if pass_min <= 0:
                 continue
             write_date = user.password_write_date
+            if not write_date:
+                continue
             delta = timedelta(hours=pass_min)
             if write_date + delta > datetime.now():
                 raise UserError(


maneandrea:
@rolandojduartem and @moylop260 Could you review? Thanks!

moylop260 (approved):

maneandrea (approved):

password_security/models/res_users.py
```diff
@@ -144,6 +144,8 @@ def _validate_pass_reset(self):
             if pass_min <= 0:
                 continue
             write_date = user.password_write_date
+            if not write_date:
+                continue
             delta = timedelta(hours=pass_min)
             if write_date + delta > datetime.now():
```
moylop260, line 150: IMHO it could be only in one line since that the variable `delta` is used only one time

But it is only a nitpick related to style not change if you disagree

```suggestion
            if write_date and write_date + timedelta(hours=pass_min) > datetime.now():
```

PR #695, opened 2024-09-25
[17.0][IMP] auth_saml: user provisioning on login

vincent-hatakeyama:
- custom message when response is too old
- avoid using werkzeug.urls method, they are deprecated
- add missing ondelete cascade when user is deleted
- attribute mapping is now also duplicated when the provider is duplicated
- factorize getting SAML attribute value, allowing using subject.nameId in mapping attributes too

This is inspired by #655.

### DIFF
--- a/auth_saml/models/auth_saml_provider.py
+++ b/auth_saml/models/auth_saml_provider.py
@@ -1,5 +1,5 @@
 # Copyright (C) 2020 Glodo UK <https://www.glodo.uk/>
-# Copyright (C) 2010-2016, 2022 XCG Consulting <https://xcg-consulting.fr/>
+# Copyright (C) 2010-2016, 2022, 2025-2026 XCG SAS <https://orbeet.io/>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import base64
@@ -81,6 +81,7 @@ class AuthSamlProvider(models.Model):
         "auth.saml.attribute.mapping",
         "provider_id",
         string="Attribute Mapping",
+        copy=True,
     )
     active = fields.Boolean(default=True)
     sequence = fields.Integer(index=True)
@@ -135,6 +136,28 @@ class AuthSamlProvider(models.Model):
     sign_metadata = fields.Boolean(
         default=True,
         help="Whether metadata should be signed or not",
+    )
+    # User creation fields
+    create_user = fields.Boolean(
+        default=False,
+        help="Create user if not found. The login and name will defaults to the SAML "
+        "user matching attribute. Use the mapping attributes to change the value "
+        "used. If a deactivated user has a matching saml uid, activate it rather than"
+        "create a new one.",
+    )
+    create_user_template_id = fields.Many2one(
+        comodel_name="res.users",
+        # Template users, like base.default_user, are disabled by default so allow them
+        domain="[('active', 'in', (True, False))]",
+        default=lambda self: self.env.ref("base.default_user"),
+        help="When creating user, this user is used as a template",
+    )
+    create_user_reactivate = fields.Boolean(
+        "Reactivate when Creating Users",
+        default=False,
+        help="If a deactivated user has a matching SAML uid when trying to create the "
+        "user, and this is checked, then the user is reactivated. Otherwise,"
+        "access is denied.",
     )
 
     @api.model
@@ -256,9 +279,7 @@ class AuthSamlProvider(models.Model):
         }
         state.update(extra_state)
 
-        sig_alg = ds.SIG_RSA_SHA1
-        if self.sig_alg:
-            sig_alg = getattr(ds, self.sig_alg)
+        sig_alg = getattr(ds, self.sig_alg)
 
         saml_client = self._get_client_for_provider(url_root)
         reqid, info = saml_client.prepare_for_authenticate(
@@ -272,6 +293,7 @@ class AuthSamlProvider(models.Model):
         for key, value in info["headers"]:
             if key == "Location":
                 redirect_url = value
+                break
 
         self._store_outstanding_request(reqid)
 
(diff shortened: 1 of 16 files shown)


lorenzoallegrucci:
LGTM, @vincent-hatakeyama could someone merge it?

vincent-hatakeyama:
The OCA process requires two reviews before merging can be done.

You are welcomed to leave a review.

auth_saml/models/res_users.py
```diff
@@ -45,11 +45,47 @@ def _auth_saml_signin(self, provider: int, validation: dict, saml_response) -> s
...
         )
         user = user_saml.user_id
-        if len(user) != 1:
-            raise AccessDenied()
+        user_copy_defaults = {}
+        if (
+            not user.active
+            and self.env["auth.saml.provider"].browse(provider).create_user
+        ):
+            user.active = True
```
arnaudlayec: Was is the use-case for this feature? (activating a disabled user)
I think it can be a security issue. Let's say I don't way a user to login anymore, but I want to keep its detail in my database: I usually archive it. But I don't want her/him to be able to authenticate back thanks to the this auto-provisioning feature.
I think we should consider archived users like non-existing user.

auth_saml/models/auth_saml_provider.py
```diff
@@ -136,6 +137,21 @@ class AuthSamlProvider(models.Model):
         default=True,
         help="Whether metadata should be signed or not",
     )
+    # User creation fields
+    create_user = fields.Boolean(
+        default=False,
+        help="Create user if not found. The login and name will defaults to the SAML "
+        "user matching attribute. Use the mapping attributes to change the value "
+        "used. If a deactivated user has a matching saml uid, activate it rather than"
```
arnaudlayec, line 145: Same comment: if we remove re-activating archived users, the "help" should be updated (last sentence to remove).

arnaudlayec (changes requested):
Hello,
It looks very good to me, thanks for the interest of the initial PR.
I just don't understand the feature of re-activating an archived user, I think it could be a security issue, unless I'm missing something.

Do you plan to (back)port it, to v16.0 / v18.0? I could help on v16.0 at least

vincent-hatakeyama:
Concerning the port to 18, there was one (#823) but it now lacks the latest fixes made here. I plan on porting the feature in 18.0 if its merged.
I do not plan to backport it in 16.0.

auth_saml/models/res_users.py
vincent-hatakeyama: The use case:

The identity provider list users and is up to date. If a user is not in the company, the user is disabled here.
Eventually, an odoo admin can disable users, but it is not mandatory ; or maybe another system is in place to disable them once disabled in the identity provider. If there was a SCIM implementation in odoo, that would be done that way (and in that case, that feature would be useless).
After some time, the employee joins back the company and their user is reactivated in the Identity Provider. In that case, the odoo user should be reactivated. As the login is unique, it is probably impossible to create a new user.

If you have a use case where the user should be not be activated again, that behaviour could be an option.
In any cases, I should update the description of the feature to reflect that point.

arnaudlayec: Thanks for the explanation, I understand now.
I agree with your use-case, except about who (which tool) should re-activate the user. I think it's up to the tool or person who disabled it in the first place to re-activate it when needed, and also to warn the administrator about the synchronization in place. It also still think it's a security issue to automatically re-active archived users here, since we don't know who/what archived it in the 1st place.
I feel like this re-activation feature should/could be managed by a SCIM implementation or any other identity-synch solution. It could be managed by this SAML module, but maybe by the claims?

To explain my use-case, it was a bit opposite to your standard. It is a small-to-medium company of 50-employees which uses Microsoft 365 for emails and identity management. The HR management wants sometime to disable user in Odoo directly (not in external identity provider) for manual-allowed control on the ERP, independently of email/identity in M365 because ERP access is sensitive. I agree controlling access to the ERP should be done by managing groups in identity provider of who has/has not access it, but this is too complex for this small/medium company.
=> So this is a use-case where `active` field modified in Odoo takes precedence over the user status in external identity tool.

Another thought:
Syncing `active` field from external identity provider is a real choice, at the discretion of the administrator.
Not tested, but: if a Odoo user is **active** Odoo and **disabled** in identity provider, I think this is the identity provider who will block the authentication request, anyway of the `active` field in Odoo, right?
=> So I'm really wondering the need of updating `active` field in hard code in Odoo.

If we want to manage the update of `active` field in this SAML module, should it be supported through the claims? I haven't tested/though about it hardly enough but I guess it would imply:
* `active_test=False` some places in the code
* early update of `active` field if managed in the claims and if modified
* check `user.active` boolean before authenticating it, or raise **AccessDenied**

Sorry for such many words, I hope I'm not adding complexity.
I'm not an identity-management expert, so if your use-case seems like a standard in the identity industry, go for it. Indeed a README.md update seems needed in that case.

vincent-hatakeyama: Identity providers are meant to be the source of the users and their rights. Having it otherwise is not standard practices. There is still a missing piece in auth_saml (also missing from auth_oauth) to match groups in the identity providers and groups in the application. This could be included in a SCIM implementation instead of putting it in the modules.
This implementation is a workaround the lack of SCIM implementation to create the user. There is still the issue of giving the correct groups to users.

Disabling users can either be done manually, with a separate process (which is done in the company that uses the feature in this PR) or not done at all.

The simplest way to handle the various cases is to introduce a new boolean in a provider to indicate if deactivated user should be activated again. I’m going to make the default to not re-enable disabled users.

arnaudlayec: Thanks for listening, I also agree
This lets some area of improvements for future work:
* manage group synchronization for permissions
* and maybe support archiving/re-activation of users through claims

auth_saml/models/auth_saml_provider.py
vincent-hatakeyama: Moved to the help text of the reactivate option.

arnaudlayec (approved):
Ok after above discussion, thank you!

PR #699, opened 2024-10-09
[18.0][MIG] base_user_empty_password

chaule97:


simahawk:
@chaule97 can you please split the test that you added in the migration commit to its own commit that can be backported? Thanks

chaule97:
> @chaule97 can you please split the test that you added in the migration commit to its own commit that can be backported? Thanks

Thanks for suggesting, I have updated

simahawk (approved):

Issue #707, opened 2024-10-10
Add support to skip the logout confirmation in the auth provider


wluyima:
**Is your feature request related to a problem?**
For some implementations, the logout confirmation is not a desirable behavior.

**Additional context**
According to the [oauth2 logout specification](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout), by including the **id_token_hint** parameter in the logout URL, the logout confirmation is skipped by the resource provider.

**Describe the solution you'd like**
An admin should be able to configure a specific oauth provider in Odoo to skip the logout confirmation in the resource provider. It means we can store the value of the **id_token_hint** in the users table after successful authentication alongside the access_token so that when the user is logging out, it gets included in the logout URL.

wluyima:
@ap-wtioit, @sbidoul  our team wishes to work on this and issue a PR.

ap-wtioit:
sounds like a useful extension to me.

PR #708, opened 2024-10-11
[14.0][IMP] auth_oidc: add support to skip the logout confirmation in the auth provider

wluyima:
https://github.com/OCA/server-auth/issues/707

### DIFF
--- a/auth_oidc/controllers/main.py
+++ b/auth_oidc/controllers/main.py
@@ -72,6 +72,8 @@ class OpenIDLogout(Session):
                 params = parse_qs(components.query)
                 params["client_id"] = provider.client_id
                 params["post_logout_redirect_uri"] = redirect_url
+                if provider.skip_logout_confirmation and user.oauth_id_token:
+                    params["id_token_hint"] = user.oauth_id_token
                 logout_url = components._replace(query=url_encode(params)).geturl()
                 return super().logout(redirect=logout_url)
         # User has no account with any provider or no logout URL is configured for the provider
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -52,6 +52,12 @@ class AuthOauthProvider(models.Model):
         "in the client, should be the value of end_session_endpoint specified by "
         "the authorization provider.",
     )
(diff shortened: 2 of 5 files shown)


sbidoul:
@wluyima This sounds useful. I'm not convinced with storing the id token on the user, if only because it won't be compatible with `oauth_oauth_muti_token`. Would it possible to obtain the id token on demand?

wluyima:
I'm not sure how multi tokens are handled in the auth provider, the addon currently also stores the access_token in the user table, so multi tokens would affect it too. My guess is that when a user logs out, the auth provider ends all their sessions for that particular client implying that as long I have one of the id tokens it would still end all their sessions across all tokens but I will do some research on this.

I'd guess it should be possible to fetch the id_token using the token_endpoint because that's its purpose.

wluyima:
@sbidoul @ap-wtioit based on my last comment and the fact that this feature is disabled by default, can we merge this PR as it is? Support for multi tokens can always be implemented in a separate future issue, the assumption is that if an implementation is using multi token, then currently this feature is not for them therefore they should not turn it on.

wluyima:
@sbidoul and @ap-wtioit do you mind merging the PR if possible? Please see my last 2 comments especially the former.

sbidoul:
I'm really not comfortable with storing the id token on the user. Also if we really need to store it it should be declared as a private field to prevent leaks via export.

But I'd prefer if it would not be stored at all, if that is possible.

rbuisson:
Hi folks, any way to get this merged?

wluyima:
@sbidoul the addon stores the access token too on the user, am not suggesting that we should go ahead to store the id token too on the user, I just wanted to raise the fact that if doing so poses is a security concern then access token needs to be fixed too.

wluyima:
Any response to my last comment?

sbidoul:
@wluyima Ah sorry, I did not read your comment as expecting a reply from me.

Anyway I'm still not comfortable with this. So I'd rather not merge it.

Also, as mentioned in other PR, I'm generally not comfortable with including the logout flows in the base `auth_oidc` module for now, so better do this in separate modules for now.

PR #711, opened 2024-10-15
[18.0][MIG] user_log_view: Migration to 18.0

moncefarajdal:


kobros-tech (commented):
ok, but didn't need to change the xml id, it should remain unigue as it is a name and can not get affected by JS, xml tag change from <tree/> to <list/>

PR #713, opened 2024-10-16
[16.0][FIX] password_security: update password_write_date on copy

maneandrea:
Sometimes users are created from a template user via a `copy()`.
This has the issue that a password is passed via the `vals` of the copy
and therefore never seen by the `write()` function.

As a result, the `password_write_date` field is left to the value of the
template, which is either outdated or null.

A concrete bug that resulted from this is that newly created users were
asked to renew their password on their very first login.

---

This commit reapplies the same logic of the `write()` method to the
`copy()` method as well.

It also changes the unit test test_03_create_user_signup to create the
user at some time in the past so that
```python
assertNotEqual(password_write_date, created_user.password_write_date)
```
makes sense.

Finally it fixes the do_signup method to user the current user's
password otherwise the password_write_date will be overwritten even when
inputting invalid passwords

### DIFF
--- a/password_security/controllers/main.py
+++ b/password_security/controllers/main.py
@@ -17,7 +17,13 @@ _logger = logging.getLogger(__name__)
 class PasswordSecurityHome(AuthSignupHome):
     def do_signup(self, qcontext):
         password = qcontext.get("password")
-        user = request.env.user
+        # If 2FA is activated, request.env.user is not updated to the logged-in user
+        # at this point. In order to do _check_password on the correct user we
+        # search by login.
+        user = (
+            request.env.user.search([("login", "=", qcontext.get("login"))])
+            or request.env.user
+        )
         user._check_password(password)
         return super(PasswordSecurityHome, self).do_signup(qcontext)
 
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -18,7 +18,7 @@ class ResUsers(models.Model):
     _inherit = "res.users"
 
     password_write_date = fields.Datetime(
-        "Last password update", default=fields.Datetime.now, readonly=True
+        "Last password update", default=fields.Datetime.now, readonly=True, copy=False
     )
     password_history_ids = fields.One2many(
         string="Password History",
--- a/password_security/tests/test_signup.py
+++ b/password_security/tests/test_signup.py
@@ -4,6 +4,7 @@
 
 from unittest import mock
 
+from freezegun import freeze_time
 from requests.exceptions import HTTPError
 
 from odoo import http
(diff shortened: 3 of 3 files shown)


maneandrea:
@moylop260 or @luisg123v Can you review?

hparfr (approved):

password_security/controllers/main.py
```diff
@@ -17,7 +17,10 @@
 class PasswordSecurityHome(AuthSignupHome):
     def do_signup(self, qcontext):
         password = qcontext.get("password")
-        user = request.env.user
+        user = (
```
moylop260, line 23: the commit and the description mention the change in the copy

But this change is not mentioned

Also the title mention "...password_write_date on copy"

Could you elaborate how this search affects the purpose of this PR, please?

maneandrea: I am open to separate this in two PRs. Conceptually they are different fixes but in practice they are both needed.

The case is when a user has 2FA activated. If that is the case, they get to this method as the public user, so the check in `user._check_password(password)` does not detect the use of an old password. This makes the write method run which in turn updates `password_write_date`. Only after then, a second call to `_check_password rejects` the change.

However, since `password_write_date` is ~not~ now updated to today, the user is able to log in with the old (expired) password and use it for a long time after that.

BTW: I updated the PR description to match the commit message.

sebalix: Hello, could you add a comment above the search to explain what we are trying to fix? That will help the maintenance in the future as it's not obvious just by reading the code.

maneandrea: @sebalix Sorry for the delay. I added some explanatory comment

moylop260 (approved):
👍

password_security/models/res_users.py
```diff
@@ -32,6 +32,11 @@ def write(self, vals):
             vals["password_write_date"] = fields.Datetime.now()
         return super(ResUsers, self).write(vals)
 
+    def copy(self, vals):
+        if vals.get("password"):
+            vals["password_write_date"] = fields.Datetime.now()
+        return super(ResUsers, self).copy(vals)
```
remi-filament: Hi @maneandrea thank you for proposing a fix for this bug.
I found the same issue when signing up, the template user is copied with its initialization value of password_write_date.

I am however wondering about the way you propose to fix it, from my point of view, it would be better that this field "password_write_date" is not copied at all (adding a copy=False on field definition). If no value is provided then it will use default (= fields.Datetime.now()).

antonag32: I'm taking over for Andrea so all issues in the PR get corrected. I just changed the field definition so that it is not copied over. Also added a unit test that verifies the fix is working. Can you review again please?

remi-filament (approved):

remi-filament:
Thanks @antonag32, tested on Runboat, it works as expected !
Best Regards

antonag32:
Great, can a maintainer merge the change?

moylop260:
FYI fowardporting:
 - https://github.com/OCA/server-auth/pull/963
 - https://github.com/OCA/server-auth/pull/964

PR #627, opened 2024-03-20
[MIG] auth_oauth_autologin: Migration to 16.0

benwillig:

PR #717, opened 2024-10-18
[16.0][ADD] auth_oauth_autoredirect

vincent-hatakeyama:
The module adds the same feature that exists in the auth_saml module, allowing redirecting users automatically to an oauth provider.

Adapted from auth_saml code.


sbidoul:
Is it not similar to auth_oauth_autologin? https://github.com/OCA/server-auth/pull/627

vincent-hatakeyama:
The feature are quite similar.
The implementation differs. I’m surprised the other module uses Javascript. The autoredirect in auth_saml was done in 11.0; I wonder why it was not reused rather than implement it in another way in auth_oauth_autologin (module added in 13.0).
To me, this module is more easy to maintain as its the same as what is done in auth_saml.

sbidoul:
Initially (in v12/13), we did it with a backend redirect too (see for instance https://github.com/OCA/server-auth/commit/f25f5d45bac10752318585bcedeac1050292708c).

But user complained that when they received "deep" links to Odoo records by email and had to go through the authentication flow, they landed on the home page instead of the record.

This is because the object reference is in a URL fragment which is client side only and not available to the backend, so was lost in the redirection. 

So @benwillig  ended up redoing it in javascript to solve that problem (in https://github.com/OCA/server-auth/pull/511).

vincent-hatakeyama:
> But user complained that when they received "deep" links to Odoo records by email and had to go through the authentication flow, they landed on the home page instead of the record.

I just checked, that also happens with auth_saml. This probably needs to be added to the module too.

I think in 18.0, doing it with javascript will not be needed any more.

I feel that auth_oauth_autologin is better so I’m closing this PR. I’m going to test the port of the other module.

PR #720, opened 2024-10-28
[IMP] migrate the auth_jwt with its demo from odoo16 to odoo17

elheit:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

Check also the CI states.

elheit:
but the problem i'm getting is from installing the pyjwt package on the odoo test machine

sbidoul:
Note there is ongoing work for auth_jwt in #657

StefanRijnhart:
Module is migrated in https://github.com/OCA/server-auth/pull/751, which was a continuation of #657. Please check open PRs next time. Thank you for your effort!

PR #722, opened 2024-10-31
[14.0][IMP] impersonate_login

toita86:
MIG/ADD of https://github.com/OCA/server-auth/pull/637 

I have decided to adopt the approach from the original PR and adjust the tests, rather than 
backporting from v16 which would have required too many rollbacks to make it compatible with v14.


toita86:
@Kev-Roche @astirpe, can you please checkout my implementation for the `_create()` method? 
It is the only way i found to guarantee the correct assignment for `create_uid` and to get the tests passing.
If you have any concerns or suggestions for improvement, please let me know.

impersonate_login/models/model.py
```diff
@@ -0,0 +1,39 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+from odoo.http import request
+
+
+class BaseModel(models.AbstractModel):
+    _inherit = "base"
+
+    @api.model_create_multi
```
HekkiMelody: **issue**: This is an `api.model`, not a `model_create_multi` in base Odoo v14

impersonate_login/models/model.py
```diff
@@ -0,0 +1,39 @@
...
+        ):
+            impersonator_id = request.session.impersonate_from_uid
+            for rec in res:
+                self.env.cr.execute(
+                    """
+                    UPDATE "%s" SET create_uid = %%s WHERE id = %%s
+                """
+                    % self._table,
+                    (impersonator_id, rec.id),
+                )
```
HekkiMelody: **issue**: Prefer `%(varname)` syntax if possible please. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#33idioms

**question**: Is there a strong reason for running a query directly instead of doing it like in the original commit?

```python
            for rec in res:
                rec["create_uid"] = request.session.impersonate_from_uid
```

If we have to run the query directly, can we at least run a single query for all records in `res` instead of a separate query for each?

HekkiMelody (changes requested):
Code review, mostly looks ok, a couple of questions

impersonate_login/models/model.py
toita86: I should have used the preferred syntax now.

> **question**: Is there a strong reason for running a query directly instead of doing it like in the original commit?

I was not able to get it working in the way was done in the original commit because the assignment of `rec["create_uid"]` doesn't not happen for some reason.

I think has something to do with the method `_create` that in the following version of this module is changed with `_prepare_create_values` but I'm not sure.

amh-mw: > I was not able to get it working in the way was done in the original commit because the assignment of `rec["create_uid"]` doesn't not happen for some reason.

Normal write operations can't update `create_uid`. Excerpts from https://github.com/odoo/odoo/blob/14.0/odoo/models.py:

```python
LOG_ACCESS_COLUMNS = ['create_uid', 'create_date', 'write_uid', 'write_date']
```
```python
        bad_names = {'id', 'parent_path'}
        if self._log_access:
            # the superuser can set log_access fields while loading registry
            if not(self.env.uid == SUPERUSER_ID and not self.pool.ready):
                bad_names.update(LOG_ACCESS_COLUMNS)
```

HekkiMelody: ah, good to know, thank you so much @amh-mw

HekkiMelody (approved):
Code review, LGTM

impersonate_login/models/model.py
toita86: @amh-mw Thanks for the explanation!

francesco-ooops (approved):
Functional ok!

@Kev-Roche @astirpe what do you think?

francesco-ooops:
@OCA/tools-maintainers @simahawk good for merge?

PR #723, opened 2024-11-07
[15.0][MIG] base_user_show_email: Migration to 15.0

amh-mw:


.pre-commit-config.yaml
```diff
@@ -23,7 +23,7 @@ exclude: |
   # You don't usually want a bot to modify your legal texts
   (LICENSE.*|COPYING.*)
 default_language_version:
-  python: python3
+  python: python3.11
```
amh-mw, line 26: This was necessary for me to run pre-commit locally, but seems out of scope for this migration.

PR #724, opened 2024-11-08
[16.0] [MIG] auth_sms

NL66278:
Protect logins by the user having to also enter a code sent through sms.

At the moment hardcoded to use the Messagebird provider.


NL66278:
I had to make one correction in another module as this had a problem in the login template that was inherited in the enter sms code template: https://github.com/OCA/server-auth/pull/724/files#diff-564288b4f2312458e84b608a19806e84706dcd2f700bc83bf4b3763b68881786

PR #726, opened 2024-11-12
[17.0][IMP] auth_saml: only write value that changes

vincent-hatakeyama:
When using mapping, not writing the value systematically avoids getting security mail on login/email changes when there is no change. Also use SQL for blanking passwords avoids the security update mails.

### DIFF
--- a/auth_saml/models/res_users.py
+++ b/auth_saml/models/res_users.py
@@ -55,7 +55,20 @@ class ResUser(models.Model):
             user_saml.with_env(new_env).write({"saml_access_token": saml_response})
 
         if validation.get("mapped_attrs", {}):
-            user.write(validation.get("mapped_attrs", {}))
+            # Only write field that changes to avoid generating Security Update on users
+            # when login/email changes (from mail module)
+            vals = {}
+            for key, value in validation.get("mapped_attrs", {}).items():
+                # If the value or the field value is not a str,
+                # avoid comparison and write anyway
+                if (
+                    not isinstance(value, str)
+                    or not isinstance(user[key], str)
+                    or user[key] != value
+                ):
+                    vals[key] = value
+            if vals:
+                user.write(vals)
 
         return user.login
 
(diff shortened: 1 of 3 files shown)


vincent-hatakeyama:
@OCA/tools-maintainers Any chance for a review?

It needs to be ported in 16 and 18 and I’d rather have a review first.

thomaspaulb:
@vincent-hatakeyama If you can try to do review trades with other PR authors, that would be preferable, and then can ping us PSC's for a merge.

auth_saml/models/res_users.py
```diff
@@ -55,7 +55,14 @@ def _auth_saml_signin(self, provider: int, validation: dict, saml_response) -> s
             user_saml.with_env(new_env).write({"saml_access_token": saml_response})
 
         if validation.get("mapped_attrs", {}):
-            user.write(validation.get("mapped_attrs", {}))
+            # Only write field that changes to avoid generating Security Update on users
+            # when login/email changes (from mail module)
+            vals = {}
+            for key, value in validation.get("mapped_attrs", {}).items():
+                if not isinstance(value, str) or getattr(user, key) != value:
```
simahawk:
```suggestion
                if not isinstance(value, str) or user[key] != value:
```
is it granted that `value` will be compatible w/ the field type? Eg: what if you have a boolean field and you compare `False` with `0`?

vincent-hatakeyama: The mapping as implemented match attributes in the SAML response. I do not believe it can be a type different than string. In keycloak, that does not seem possible.

It’s possible to map one of the attribute to a boolean field. It might be strange to the users that any value is considered as True ("0", "1", "False", etc.) and only "" would put False in the field.
Other type of fields can also be written, but in most cases, that will fail. Writing a many2one with a string fails for example.

To make the code more robust, I’m going to only compare if the value is a string. That will avoid rewriting login/email and any string or selection field which is probably the most often written field when the feature is used.

simahawk: Thanks for the feedback. Can you please report this comment in the code so that we'll have clear where we come from? :)

vincent-hatakeyama: I’ve already adjusted code and added a comment.

https://github.com/OCA/server-auth/pull/726/files#diff-3f2fc558f32a584e17b5ab27d08e9fe0d111070b3230b58d1db2ddc24376309fR62

StefanRijnhart (approved):

thomaspaulb (approved):

PR #610, opened 2024-01-31
[17.0][MIG] password_security: Migration to 17.0

beagle-cloud:

PR #731, opened 2024-11-21
[17.0][MIG] password_security + use ir.config_parameters

alexis-via:
Replaces migration PR #610 

Migration to v17 + move config parameters from res.company to ir.config_parameter. That way, a user that has access to several companies cannot switch to the company with weakest password security params before changing its password, which was a kind of security hole. 

By the way, the password security param of auth_password_policy was already stored in ir.config_parameter (cf https://github.com/odoo/odoo/blob/17.0/addons/auth_password_policy/models/res_config_settings.py#L8), so it's coherent to do the same here.


amh-mw (changes requested):
Though I agree with the intent of the improvement, this is the second migration pull request in a row for password_security where improvements are being done in the migration pull request. I think it will be more coherent to migrate between versions as usual and then review the improvement in its own pull request.

alexis-via:
The move to ir.config_parameter is in a separate commit.

etobella:
@amh-mw the usual flow in this kind of changes is as @alexis-via suggested (at least is what I found in a lot of cases in the OCA)

etobella (approved):

alexis-via:
Let's merge this. Then I'll help to backport it.

password_security/migrations/17.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,27 @@
...
+def migrate(env, version):
+    env.cr.execute(
+        f"SELECT {openupgrade.get_legacy_name('password_expiration')}, "
+        f"{openupgrade.get_legacy_name('password_minimum')}, "
+        f"{openupgrade.get_legacy_name('password_history')}, "
+        f"{openupgrade.get_legacy_name('password_lower')}, "
+        f"{openupgrade.get_legacy_name('password_upper')}, "
+        f"{openupgrade.get_legacy_name('password_numeric')}, "
+        f"{openupgrade.get_legacy_name('password_special')} "
+        "FROM res_company ORDER BY id LIMIT 1"
```
amh-mw: This could degrade security for installations with more than one company. Perhaps remove `LIMIT 1` and use `MAX(field)` or `MIN(field)` as appropriate?

password_security/migrations/17.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,57 @@
...
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(
```
amh-mw, line 11: Why are company fields being renamed if `ir.config_parameter` is replacing them? Shouldn't pre-migration copy the fields into `ir.config_parameter`, then post-migration drop the fields?

password_security/models/res_users.py
```diff
@@ -32,16 +32,37 @@ def write(self, vals):
...
+            ),
+            "minimum_hours": int(
+                params.get_param("password_security.minimum_hours", default=60)
+            ),
+            "history": int(params.get_param("password_security.history", default=30)),
+            "lower": int(params.get_param("password_security.lower", default=1)),
+            "upper": int(params.get_param("password_security.upper", default=1)),
+            "numeric": int(params.get_param("password_security.numeric", default=1)),
+            "special": int(params.get_param("password_security.special", default=1)),
+        }
```
amh-mw, line 53: These defaults don't match previous versions. Change all default back to zero and get rid of this method entirely?

password_security/models/res_config_settings.py
```diff
@@ -6,20 +6,59 @@
...
+    # present in the database, Odoo displays the default value
+    # on the config page => Odoo displays 1 !
+    # So, when the users sets the value of 0 on the config page, he will see 1
+    # after saving the page !!!
+    # If the default value is 0 (like auth_password_policy.minlength in the
+    # module auth_password_policy of the official addons), there is no problem.
+    # So the solution to avoid this problem and have a non-null default value:
+    # 1) define the ir.config_parameter fields on res.config.settings with default=0
+    # 2) initialize the ir.config_parameter with a default value in the init script
+    # So the default value of the fields below are written in post_install.py
```
amh-mw, line 21: This explanation is unnecessary if all defaults are returned to zero.

amh-mw (changes requested):

antoniocanovas (approved):
Functional review OK !!

password_security/migrations/17.0.2.0.0/pre-migration.py
etobella: This is done this way in order to avoid that odoo removes the fields.

etobella:
@alexis-via Can you attend the comments please?

password_security/models/res_users.py
fkantelberg: The only default which isn't matching with version 16.0 is `minimum_hours`. There is was 24 as default on the fields of res.company.

The defaults in the res.config.settings are different than the actual ones.

IsabelAForgeFlow:
LGTM Functional + code

IsabelAForgeFlow (approved):
Functional review

amh-mw (approved):

antonag32:
I was going to do a forward port of this fix: https://github.com/OCA/server-auth/pull/713

But since the module has not been migrated yet it might be a good opportunity to include the commit in this PR from the get go.

password_security/models/res_config_settings.py
alexis-via: This is a very very important comment, because it explains a design decision which is not obvious at all. We should keep it.

password_security/migrations/17.0.2.0.0/post-migration.py
alexis-via: The security problem is not really in this migration script but in the design of this module which had per-company settings and nobody dared to change this for years !!!
I was very surprised to be the first developer to criticize the design error and propose to fix it.
Anyway, admins will be to checks the new settings after update. I feel that mixing settings from different companies is kind of strange, so my idea was to simplify take the settings from the first company. I just updated to code to take the settings from the first **active** company.

PR #733, opened 2024-12-09
[16.0][FIX] users_ldap_populate: get additional operational attributes

dhx:
When using users_ldap_populate together with the users_ldap_groups module to synchronize groups especially the memberOf attribute is typically not fetched from the LDAP server without specifying it. This commit fixes this by adding the ['*', '+'] attributes to the search_st call.

### DIFF
--- a/users_ldap_populate/models/users_ldap.py
+++ b/users_ldap_populate/models/users_ldap.py
@@ -125,7 +125,11 @@ class CompanyLDAP(models.Model):
         ldap_binddn = conf["ldap_binddn"] or ""
         conn.simple_bind_s(ldap_binddn, ldap_password)
         results = conn.search_st(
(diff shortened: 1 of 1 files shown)


dhx:
Hi, @yajo I see you've also commited to this module :) We're using this improvement for quite some time now - do you think it makes sense to merge it in this current form?

yajo:
Sorry I don't use or maintain this since long ago, and most LDAP things are out of my mind right now. I guess I'm not the best to review this particular PR. 🙄

dhx:
Hi @CarlosRoca13 @carolinafernandez-tecnativa as you've done previous migrations of this module does this change make sense to merge for you?

CarlosRoca13 (approved):
It's OK for me

pedrobaeza (approved):
OK, I don't this supposes any security problem, as it's just fetching more data.

/ocabot merge patch

PR #734, opened 2024-12-12
[16.0] [ADD] cross_connect_server

paradoxxxzero:
This module allows other odoo instances, where the `cross_connect_client` (https://github.com/OCA/server-auth/pull/735) module is
installed and configured, users to connect directly on this odoo instance.

(Needed to bump precommit flake8 version for annotation support)

PR #735, opened 2024-12-12
[16.0] [ADD] cross_connect_client

paradoxxxzero:
This module allows an odoo instance users to connect directly on another odoo instance
where the module `cross_connect_server` (https://github.com/OCA/server-auth/pull/734) is installed.


cross_connect_client/models/cross_connect_server.py
```diff
@@ -0,0 +1,180 @@
...
+from odoo.exceptions import UserError
+
+
+class CrossConnectServer(models.Model):
+    _name = "cross.connect.server"
+    _description = "Cross Connect Server"
+    _inherit = "server.env.mixin"
+
+    name = fields.Char(
+        required=True, compute="_compute_name", readonly=False, store=True
```
florian-dacosta: I would add a `help` to explain that this name will be the new "app" name.
Also, I wonder why it is computed from the `server_url`. 
I mean, it does not cost much to have a normal (not computed) required field, as it is quite important.
I don't see in which case we could want to have the new app with the server url name.

cross_connect_client/models/cross_connect_server.py
```diff
@@ -0,0 +1,180 @@
...
+                    }
+                )
+
+    def action_sync(self):
+        for record in self:
+            record._sync_groups()
+
+    def action_disable(self):
+        for record in self:
+            record.group_ids.unlink()
```
florian-dacosta: I don't like this button very much because I think it is easy to click without knowing the consequences (deletion of the group). Because clicking on `action_disable`, then `action_sync` is not without consequence since you would have to reconfigure the group on all users needing support.

I wonder if we could do something like `record.menu_id.unlink()` instead.
action_sync re-create the menu anyway (or if it does not if not group is created, we could force it.

Then, the `action_disable` would disable the menu but would still keep all user / group configuration.
What do you think ?

cross_connect_client/models/cross_connect_server.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @property
+    def _server_env_fields(self):
+        return {"api_key": {}}
```
florian-dacosta, line 173: I think we should override the `unlink` method to delete the groups and menu before deleting the cross connect server.
(Even without changing the `action_disable` method, I guess the `cross.connect.server` could be deleted while enabled)

cross_connect_client/models/cross_connect_server.py
```diff
@@ -0,0 +1,180 @@
...
+        for record in self:
+            if not record.group_ids:
+                if record.menu_id:
+                    if record.menu_id.action:
+                        record.menu_id.action.unlink()
+                    record.menu_id.unlink()
+                record.menu_id = False
+                continue
+
+            menu_groups = self.env.ref("base.group_system") | record.group_ids
```
florian-dacosta: `self.env.ref("base.group_system")` should be removed IMO because, the group does not come from server so it will display the new app menu but won't allow you to access it.

florian-dacosta (approved):
LGTM test + code review.
I'd added some not blocking remarks, but is good enough to be merged already IMO.

cross_connect_client/models/cross_connect_server.py
paradoxxxzero: Hostname but it’s just for minor convenience, might not be needed, remove the attrs in the view in this case.

cross_connect_client/models/cross_connect_server.py
paradoxxxzero: Agreed

cross_connect_client/models/cross_connect_server.py
paradoxxxzero: Maybe archiving menu entries in this case?

cross_connect_client/models/cross_connect_server.py
paradoxxxzero: Can’t we cascade these?

cross_connect_client/models/cross_connect_server.py
florian-dacosta: what about the removal of the compute method ?

cross_connect_client/models/cross_connect_server.py
florian-dacosta: Better indeed.

cross_connect_client/models/cross_connect_server.py
paradoxxxzero: ! meant also the attrs :sweat_smile:

cross_connect_client/models/cross_connect_server.py
florian-dacosta: Well why not, we'll just have to manage the active state in the `_compute_menu_id` then.
But ok for me, I 'd just prefer avoiding the groups deletion

florian-dacosta (approved):

sebastienbeau (approved):
LGTM (code review)

PR #736, opened 2024-12-12
[17.0]  auth_session_timeout: Fix pre-migrate script

Robrechtc:
The pre-migrate script should use cr instead of env as parameter.


sebalix (approved):
Thanks, indeed the `env` is usable only if `def migrate` is decorated with `@openupgrade.migrate` (but no need to introduce this dependency here)

PR #738, opened 2024-12-17
[16.0][FIX] auth_api_key: Use 401 for unauthorized user

anusriNPS:
As per documentation, using [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) as status code seems valid for a user who is not able to be authorized. However, [403](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) is used for scenario where user is forbidden to access a  resource. So, returning 401 instead of 403.

### DIFF
--- a/auth_api_key/models/ir_http.py
+++ b/auth_api_key/models/ir_http.py
@@ -5,8 +5,9 @@
 
 import logging
 
+from werkzeug.exceptions import Unauthorized
+
 from odoo import models
-from odoo.exceptions import AccessDenied
 from odoo.http import request
 
 _logger = logging.getLogger(__name__)
@@ -33,4 +34,4 @@ class IrHttp(models.AbstractModel):
                 request.auth_api_key_id = auth_api_key.id
                 return True
         _logger.error("Wrong HTTP_API_KEY, access denied")
-        raise AccessDenied()
+        raise Unauthorized()


HekkiMelody (approved):
This is the same change as https://github.com/OCA/server-auth/pull/629 , I think it makes sense here as it did there.

LGTM

HekkiMelody:
@OCA/tools-maintainers Hello, this is a tiny fix to be more compliant with expectations of a REST architecture, we've been using it in production for a few months now, can it be merged? Thanks

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
There is no declared maintainer of the module so I'm merging...

/ocabot merge patch

HekkiMelody:
Thank you @LoisRForgeFlow !

LoisRForgeFlow:
@aleuffre no problem! Could you take care of fwporting the fix to 17 and 18? ping me on the PR and we can quickly merge

simahawk:
@anusriNPS @aleuffre thanks for your contrib! 
Next time please, respect commit convention by removing the odoo version.
I see that in the fwd ports, the commit msg has been rewritten to match the odoo version: we shouldn't do it.
If you drop the version in the original commit, there's nothing to fix afterwards :wink:

HekkiMelody:
> @anusriNPS @aleuffre thanks for your contrib! Next time please, respect commit convention by removing the odoo version. I see that in the fwd ports, the commit msg has been rewritten to match the odoo version: we shouldn't do it. If you drop the version in the original commit, there's nothing to fix afterwards 😉

You're right, I missed it when reviewing, apologies

anusriNPS:
> @anusriNPS @aleuffre thanks for your contrib! Next time please, respect commit convention by removing the odoo version. I see that in the fwd ports, the commit msg has been rewritten to match the odoo version: we shouldn't do it. If you drop the version in the original commit, there's nothing to fix afterwards 😉

Sure, will take care to align with Odoo convention in upcoming commits.

PR #694, opened 2024-09-24
[17.0][MIG] auth_oauht_multi_token: Migration to 17.0

timnis:
Based on https://github.com/OCA/server-auth/pull/678 and thanks to [tb-flyt](https://github.com/tb-flyt)

Tried to follow instructions and hopefully I got it right 😃

PR #742, opened 2024-12-31
[17.0][MIG] auth_oauth_multi_token: Migration to 17.0

kobros-tech:
fixing git commits log history with keeping the changes the same for PR #694 this will enable me to migrate to 18.0


timnis:
Hi @CRogos could this PR merged?

CRogos (changes requested):
Could you have a look on your commits? Some of them should be squashed, and the last commit should be splitted into: 
[IMP] auth_oauth_multi_token: pre-commit auto fixes
[MIG] auth_oauth_multi_token: Migration to 17.0
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

image

Code LGTM, but havn't tested yet.

Do you use odoo.sh? The last time I've tested this module it was not compatible with the login-as function of odoo.sh

timnis:
Hi @kobros-tech Could you have a look on your commits?

kobros-tech:
@timnis 
@CRogos 

Hi all, I improved the PR more, and all contributions logs are mentioned.

To have all log preserved here, there was a conflict arised and I guess the first PR was not made correctly as I had to resolve a conflict in another module which is out of scope.

If I git rid of there commits and start on top of last merged commit in version 16.0 or to keep it as it is

kobros-tech:
@CRogos 
@timnis 

so far so good :)

CRogos:
@kobros-tech can you squash these commits as well: 
image

also some of your content in the migration commit should be in the "[IMP] auth_oauth_multi_token: pre-commit auto fixes" commit as it was done in the original PR (but without the deb file)

image

kobros-tech:
@CRogos 

please, make sure that the commit you mentioned that has pandoc file is not there, only the [MIG] commit which is clean as I guess.

I don't understand why should I squash commits done for 16.0 and not for 17.0, could you explain?

CRogos:
@kobros-tech 
Squashing commits is part of the migration guide: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0
and there is an extra document, which explains what to squash and how to do it: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

In https://github.com/OCA/server-auth/pull/694 the Commit history is according to the migration guide.
image
But this PR has the issue with the *.deb file, which is a big issue. We need to make sure that this file is not commited in any of the commits. (adding in one commit and removing it in another commit still blows up the repository size)

Nevertheless, the "[IMP] auth_oauth_multi_token: pre-commit auto fixes" commit of https://github.com/OCA/server-auth/pull/694 is correct beside of the *.deb file and should also exist in this PR.

kobros-tech:
@CRogos 

**I need you to look at the log of commits, because the commit you are talking about is squashed already** and pandoc file is not mentioned or tracked in the migration commit and you can review, thanks!

Screenshot from 2025-01-24 19-30-35
Screenshot from 2025-01-24 19-30-54

kobros-tech:
@CRogos 
If you still are suspecting, I strongly ask you for the commit id that contains pandoc file here and not in other PR as I worked on it and I am quite sure that pandoc file is not tracked at all in this PR branch

CRogos:
> @CRogos If you still are suspecting, I strongly ask you for the commit id that contains pandoc file here and not in other PR as I worked on it and I am quite sure that pandoc file is not tracked at all in this PR branch

There is no commit with pandoc file, but there is also no commit "[IMP] auth_oauth_multi_token: pre-commit auto fixes" which should be there.

image

kobros-tech:
> > @CRogos If you still are suspecting, I strongly ask you for the commit id that contains pandoc file here and not in other PR as I worked on it and I am quite sure that pandoc file is not tracked at all in this PR branch
> 
> There is no commit with pandoc file, but there is also no commit "[IMP] auth_oauth_multi_token: pre-commit auto fixes" which should be there.
> 
> image

thanks, now I understand what you mean.

In fact I am not used to make two commits for migration, I have a point of view that I add it in a single commit, I don't mind you if you see it necessary here.

kobros-tech:
@CRogos 
let me know if I have to divide the commit into two instad of one, so that we get the module merged and get satisfied.

CRogos:
Yes, that is what I'm trying to explain. You need to split the commit and adapt the commit message.

kobros-tech:
@CRogos 
two commits are there, clean and out of pandoc tracking

CRogos:
Better, but could you squash the yellow marked and correct the commit message of the last commit? 
image

auth_oauth_multi_token/views/res_users.xml
```diff
@@ -0,0 +1,34 @@
...
+            </field>
+            <xpath expr="//field[@name='oauth_provider_id']/.." position="after">
+                <group name="multi_token_info" string="Latest Tokens">
+                    <label for="oauth_access_token_ids" />
+                    <field
+                        name="oauth_access_token_ids"
+                        nolabel="1"
+                        options="{'no_create': True, 'no_open': True}"
+                    >
+                        <tree limit="10">
```
CRogos, line 20: This node should be changed to <list

CRogos (changes requested):

auth_oauth_multi_token/views/res_users.xml
kobros-tech: Is list tag introduced in version 18.0 or version 17.0?

CRogos: you are totally right... got confused, sorry.

CRogos (approved):

auth_oauth_multi_token/models/res_users.py
```diff
@@ -0,0 +1,83 @@
...
+    def _check_credentials(self, password, env):
+        """Override to check credentials against multi tokens."""
+        try:
+            return super()._check_credentials(password, env)
+        except exceptions.AccessDenied:
+            res = self.multi_token_model.sudo().search(
+                [("user_id", "=", self.env.uid), ("oauth_access_token", "=", password)]
+            )
+            if not res:
+                raise
```
CRogos, line 78: I think we should add an additional check like in the link below.

```
            passwd_allowed = env['interactive'] or not self.env.user._rpc_api_keys_only()
            if passwd_allowed and self.env.user.active:
```

https://github.com/odoo/odoo/blob/10550d9f20f7d08157b4c885d59efb65821ab651/addons/auth_oauth/models/res_users.py#L137-L152

CRogos (commented):
I've now also tested this PR and it is working fine. 

Nevertheless I think a found another issue in this module in the _check_credentials method. 

I also found out, why this module is not working with odoo.sh "login as". I was also able to change this module to make it compatible with odoo.sh by reusing "oauth_access_token" instead of adding "oauth_master_uuid". Neverthless I am not sure if refactoring this module or adding a new module "auth_oauth_multi_token_odoosh" is the right way, because the original solution is cleaner as the patched version?

https://github.com/c4a8-odoo/module-oca-server-auth/commit/d108748b8e2b541b772aace84bf94fbd02393549

auth_oauth_multi_token/models/res_users.py
kobros-tech: @CRogos 

I will add your commit right now the this PR
Any time if you can have access of time, I don't mind if you use and append to our wok.

It is a community and our prior interest is to get something effective.

thanks

CRogos: @sbidoul I saw your feedback on the v18 migration of this module and I would like to ask you how you think about this change?

https://github.com/OCA/server-auth/pull/742/commits/e6c6427043cd977f8d036063c29a69a9f6c5f1c5

kobros-tech: I will review it
@kobros-tech

dnplkndll: this is to allow admin to impersonate? I really like that to verify access. 

I was going to ask about making the provider a one2many too. 
I think would need to be in this module as you might have 2 tokens with 2 providers. the user case would be a multi-company / multi-website each with their own branded oauth login page to match the website / company. if you where to show on each website the sign in for that company. then it needs its only client id and redirect. 
if you have your B2B customers each on their own user no issue. but then the internal users cold not login in anything but one domain/company.

CRogos: No this is an addon, which i found at a similar case in another documentation (see the link). 
The picked commit after the migration makes it compatible with the Odoo.sh implementation which does the "login as".

dnplkndll (approved):

CRogos:
@kobros-tech could you rename the migration commit and squash the yellow marked commits?

image

kobros-tech:
@CRogos 
can you check the log now?

CRogos:
@pedrobaeza could you have a look and merge if you agree with the changes.

Regarding the changes related to odoo.sh, I also opened a ticket at Odoo, but they are not willing to change anything on their implementation.

auth_oauth_multi_token/models/res_users.py
```diff
@@ -24,60 +20,77 @@ def _generate_oauth_master_uuid(self):
...
+
     oauth_access_max_token = fields.Integer(
         string="Max Number of Simultaneous Connections", default=10, required=True
     )
-    oauth_master_uuid = fields.Char(
+
+    # use the oauth_access_token field as oauth_master_uuid
+    oauth_access_token = fields.Char(
         string="Master UUID",
-        copy=False,
```
pedrobaeza: Why are you removing these attributes?

CRogos: You are right. I thought False is the default, but it is not. @kobros-tech could you please readd the copy=False?

pedrobaeza:
I need more explanation of the latest commit, and I think it should be out of the migration itself in a separate PR, so we can merge as is the migration, and then discuss about that changes, with steps to reproduce the problem, etc, etc.

kobros-tech:
ok give me one hour and I will remove the commit

kobros-tech:
@pedrobaeza 
I removed the last commit

pedrobaeza (approved):
/ocabot migration auth_oauth_multi_token
/ocabot merge nobump

CRogos:
Added a new PR regarding the odoo.sh issue https://github.com/OCA/server-auth/pull/764

PR #743, opened 2024-12-31
[18.0][MIG] auth_oauth_multi_token: Migration to 18.0

kobros-tech:
Based on PR 742 which is just improvement of PR 694
#694 => #742 [17.0][MIG]


dnplkndll (approved):
looks good functional test, oidc aws cognito

kobros-tech:
@sbidoul could you review this PR and 17.0 one too?

auth_oauth_multi_token/tests/test_multi_token.py
```diff
@@ -38,10 +38,14 @@ def test_no_provider_no_access(self):
...
-                self.provider_google.id, validation, params
-            )
+        try:
+            with self.assertRaises(exceptions.AccessDenied):
+                self.user_model._auth_oauth_signin(
+                    self.provider_google.id, validation, params
+                )
+        # In version 18.0 AccessDenied exception is raising ValueError exception
+        except ValueError as e:
+            self.assertEqual(str(e), "Unknown token version")
```
sbidoul: This change to the test looks suspicious to me. It probably does not test what it was meant to test anymore.

kobros-tech: not at all, in version 17.0 
there is one single exception raised, the one you mean!

In version 18.0, the exception you mean is raising another exception which is the source one.

I handle it and then I trace the log message for the one you mean.

If you like to update the test case all right show me your opinion.

dnplkndll: could I suggest:
https://github.com/OCA/server-auth/pull/754/commits/a8955ff9d9523135312b6e33d75c9c998de880bf
prevent attempt to create versus refuse. or did we want to change the expects to be a new user?
'Exception: Unknown token version'
is thrown in the https://github.com/odoo/odoo/blob/18.0/addons/auth_oauth/models/res_users.py#L111 I believe.

kobros-tech:
@sbidoul 
you have one suggestion to review: 

[https://github.com/OCA/server-auth/commit/e6c6427043cd977f8d036063c29a69a9f6c5f1c5](https://github.com/OCA/server-auth/commit/e6c6427043cd977f8d036063c29a69a9f6c5f1c5)

kobros-tech:
forwarded to #754

PR #752, opened 2025-01-16
[18.0][MIG] auth_jwt: Migration to 18.0 

dnplkndll:

PR #753, opened 2025-01-17
[18.0][IMP] auth_jwt: allow more authorization options over aud 

dnplkndll:
@kobros-tech can you add some tests?

probably want to require one of the possible types too. so maybe a type selection [aud,scope,group] then the aud_text to use to parse the match value?

need to rebase after: https://github.com/OCA/server-auth/pull/752 merges

image
or
image


dnplkndll:
@sbidoul working aws cognito as a token provider, there are no aud. but we do have scopes and user groups. would it be useful to replace the aud with these new tests to validate a server to server account has proper scope or a user has a group? the tests are pretty sloppy ( any intersection of the sets) can refine if useful. another option might be a simple disable but really do need the scope test in our case. 
The next issue is we would have tokens that have sub and map user/partner via oath provider data? I guess that can be done in partner_id_strategy extension. 
https://github.com/OCA/server-auth/blob/035093daf0edc8ee30ce9b65c263aa8f1de5048a/auth_oidc/models/res_users.py#L72

sbidoul:
I'm ok to make `audience` optional. May I suggest doing that in an independent PR to facilitate review?

Then adding validation on additional claim sounds ok too. I would not override the meaning of the audience field, though. How about an `expected_claim` field containing a literal dictionary (to be parsed with ast.literal_eval) or a json field.

dnplkndll:
optional aud only
https://github.com/OCA/server-auth/pull/755

kobros-tech:
@sbidoul 

We have added test cases here and there is another PR for the optional aud, do you need any thing else?

@lmignon 
I like you to review if there is any observation, welcome to your opinion.

auth_jwt/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Auth JWT",
     "summary": """
         JWT bearer token authentication.""",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
```
lmignon, line 8: Version should not be modified. It will be automatically  bumped at merge.

lmignon (changes requested):
Thank you for the proposal @dnplkndll 
Can you improve the documentation to explain your new options. I personally find it difficult to understand the proposed changes and the problem they are trying to solve. I am therefore unable to give an informed opinion on what is being proposed.

PR #755, opened 2025-01-18
[18.0][IMP] auth_jwt drop audience required

dnplkndll:
support simple jwt token issuer like AWS cognito using a pool_id in the certificate path to segment the resource group

ref: https://stackoverflow.com/questions/53148711/why-doesnt-amazon-cognito-return-an-audience-field-in-its-access-tokens


kobros-tech:
@lmarion-source 
@sbidoul

lmignon (commented):
Thank you for the proposal @dnplkndll 

It would be useful to add to the documentation (or help attribute) when it is necessary to specify the target audience or not.

PR #761, opened 2025-02-11
[17.0][FIX] auth_saml: avoid redirecting when there is a SAML error

vincent-hatakeyama:


StefanRijnhart (approved):
Thanks! I see it's `saml_error` in https://github.com/OCA/server-auth/blob/17.0/auth_saml/controllers/main.py#L134 as well.

PR #762, opened 2025-02-12
[16.0][FIX] password_security: Fix expected singleton

moylop260:
Updating the password for more than one user raises expected singleton error

This fix supports multi case

### DIFF
--- a/password_security/models/res_users.py
+++ b/password_security/models/res_users.py
@@ -8,6 +8,7 @@ from datetime import datetime, timedelta
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
+from odoo.tools import groupby
 
 
 def delta_now(**kwargs):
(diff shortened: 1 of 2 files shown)


password_security/models/res_users.py
```diff
@@ -97,22 +98,23 @@ def _check_password(self, password):
...
-            "(?=.*?[A-Z]){" + str(company_id.password_upper) + ",}",
-            "(?=.*?\\d){" + str(company_id.password_numeric) + ",}",
-            r"(?=.*?[\W_]){" + str(company_id.password_special) + ",}",
-            ".{%d,}$" % int(minlength),
-        ]
-        if not re.search("".join(password_regex), password):
-            raise ValidationError(self.password_match_message())
+        users_by_company = groupby(self, lambda u: u.company_id.id)
+        for company_id, users in users_by_company:
+            company_id = users[0].company_id
```
luisg123v: This seems unnecessary because it's assigned in the for loop, maybe you just need to group by the record instead of the ID, which would produce a similar result as `_read_group` in Odoo >= v18.

password_security/models/res_users.py
```diff
@@ -97,22 +98,23 @@ def _check_password(self, password):
...
+            params = self.env["ir.config_parameter"].sudo()
+            minlength = params.get_param("auth_password_policy.minlength", default=0)
```
luisg123v: This could be outside the for loop

luisg123v (changes requested):

luisg123v:
Commit message:

"Updating the password for more than one user it raises": remove "it"

StefanRijnhart (approved):

pedrobaeza (approved):
/ocabot merge patch

antonag32:
Hello @pedrobaeza, can you issue the merge command again? I rebased the PR so that it uses the updated version of workflow files, it should not fail anymore.

moylop260:
FYI Foward-porting
 - https://github.com/OCA/server-auth/pull/961
 - https://github.com/OCA/server-auth/pull/962

PR #764, opened 2025-02-17
[17.0][IMP] auth_oauth_multi_token: make it compatible with odoo.sh "login as"

CRogos:
Odoo.sh provides an option to login-in as a specified user. 
When you install this OCA module auth_oauth_multi_token , the "connect" or "login as" of odoo.sh is not working anymore, but this is essential for the Odoo support.

image
image

This is caused by this change of the module: https://github.com/OCA/server-auth/blob/260f2629e42f8ea3c8ea7d322771e8a4db60e399/auth_oauth_multi_token/models/res_users.py#L80-L83

This causes, that the session_token (I assume generated by odoo.sh code) is not valid and the check here causes an automatic logout.
https://github.com/odoo/odoo/blob/7d3931bf36996cf5b51533e4a0a788ad8f3eb6c0/odoo/service/security.py#L11-L13C17

https://github.com/odoo/odoo/blob/7d3931bf36996cf5b51533e4a0a788ad8f3eb6c0/odoo/addons/base/models/res_users.py#L924-L941

My assumption is, that odoo.sh does not use the changes done by the _get_session_token_fields method, when creating the login session, but because the odoo.sh code is not public I cannot validate this assumption. 

Meanwhile I also received and answer from Odoo and they are not willing/able to improve their implementation. 

This PR bypasses the change of `def _get_session_token_fields` and therefore stays compatible with Odoo.sh


auth_oauth_multi_token/models/res_users.py
```diff
@@ -24,13 +20,15 @@ def _generate_oauth_master_uuid(self):
         readonly=True,
         groups="base.group_system",
     )
+
```
pedrobaeza: Don't add extra empty lines in the middle.

auth_oauth_multi_token/models/res_users.py
```diff
@@ -24,13 +20,15 @@ def _generate_oauth_master_uuid(self):
...
     oauth_access_max_token = fields.Integer(
         string="Max Number of Simultaneous Connections", default=10, required=True
     )
-    oauth_master_uuid = fields.Char(
+
+    # use the oauth_access_token field as oauth_master_uuid
+    oauth_access_token = fields.Char(
         string="Master UUID",
```
pedrobaeza: You only need to define the default, isn't it?

auth_oauth_multi_token/models/res_users.py
```diff
@@ -39,45 +37,62 @@ def _generate_oauth_master_uuid(self):
     def multi_token_model(self):
         return self.env["auth.oauth.multi.token"]
 
+    @api.model
+    def _generate_signup_values(self, provider, validation, params):
+        """Because access_token was replace in
```
pedrobaeza:
```suggestion
        """Because access_token was replaced in
```

pedrobaeza (commented):
I think you need to write migration script for moving the values in `oauth_master_uuid` to `oauth_access_token`, isn't it?

auth_oauth_multi_token/models/res_users.py
CRogos: The origninal definition is this:
`
oauth_access_token = fields.Char(string='OAuth Access Token', readonly=True, copy=False, prefetch=False)
`

pedrobaeza: The redefinition of labels cause a mix of terms across the system sometimes, so it's not the best option. And the required=True will provoke troubles everywhere.

CRogos: I am also not very happy for reusing this field. We could leave the string value as is and just add an comment that this field is used as Master UUID instead of access_token.
The required=True I just took from the oauth_master_uuid field, but I am not sure if this is needed to force the default creation? It shouldn't be empty or the same value as another user, to make sure that the access_token is not used by a different user.

CRogos:
> I think you need to write migration script for moving the values in `oauth_master_uuid` to `oauth_access_token`, isn't it?

I've added a migration script.

I've also added a test and found out, that not the oldest but the second oldest token was removed when reaching max token number.

gbrito (approved):

CRogos:
@pedrobaeza could you have a look on the last update?

pedrobaeza (approved):
OK, it seems better, but I would like to have the opinion of other old user like @sbidoul

CRogos:
@sbidoul could you have a look into this change?

CRogos:
How can we continue on this PR?

dreispt (approved):

PR #780, opened 2025-04-01
[MIG] auth_jwt_demo: Migration to 18.0

kobros-tech:
- [x] https://github.com/OCA/server-auth/pull/752


kobros-tech:
@sbidoul 

demo is there now, I had to ignore eslint for js test files, also warn message had to be ignored for CI test.

auth_jwt_demo/tests/test_auth_jwt_demo.py
```diff
@@ -1,13 +1,27 @@
...
+    def filter(self, record):
+        return (
+            'A route with auth="jwt" must not be used within a user session'
+            not in record.getMessage()
+        )
+
+
+# Add filter to the specific logger
+logger = logging.getLogger("odoo.addons.auth_jwt.models.ir_http")
+logger.addFilter(IgnoreSpecificWarningFilter())
```
sbidoul, line 22: We should not filter warnings like this.

If this warning is expect it we should capture it and test that it has been emitted with `self.assertLogs`

sbidoul (changes requested):
see previous comment

sbidoul:
Also, let's not change `auth_jwt_demo/tests/spa/js/oidc-client.js`. This file is an external lib. We should exclude it from the linter.

auth_jwt_demo/tests/test_auth_jwt_demo.py
kobros-tech: can you give me an example?

kobros-tech:
> Also, let's not change `auth_jwt_demo/tests/spa/js/oidc-client.js`. This file is an external lib. We should exclude it from the linter.

how can I exclude a file?

@sbidoul

TDu:
> > Also, let's not change `auth_jwt_demo/tests/spa/js/oidc-client.js`. This file is an external lib. We should exclude it from the linter.
> 
> how can I exclude a file?
> 
> @sbidoul

Probably somewhere in this file https://github.com/OCA/server-auth/blob/0771ec76cd9f593acb3f1260fe4ed374c4c00522/.pre-commit-config.yaml#L115

auth_jwt_demo/tests/test_auth_jwt_demo.py
kobros-tech: anyway, I am ignoring the warining log here as it causes test error in the run boat although it is intentional log from the base module.

I don't want to assert the log but I want to ingonre it from test wathcing errors.

dnplkndll:
@kobros-tech is this ready for review?

kobros-tech:
> @kobros-tech is this ready for review?

to me yes, but I am waiting for @sbidoul to tell his opinion.

sbidoul:
@kobros-tech I'm finishing this in #846. The log capture you made here is too broad as we expect the warning in exactly one test.

sbidoul:
Thanks for you help with this migration nevertheless!

PR #775, opened 2025-03-10
[17.0][FIX] auth_signup_verify_email: Use reCaptcha

Cl0ut1eR:
#774

Issue #774, opened 2025-03-10
[17.0] auth_signup_verify_email unable to use with reCAPTCHA


ecloutier-fcfq:
Using reCAPTCHA and auth_signup_verify_email at the same time results in the error ValueError: Invalid field 'recaptcha_token_response' on model 'res.users'

## Module
auth_signup_verify_email (Verify email at signup)

## Describe the bug
"Something went wrong, please try again later or contact us." when reCAPTCHA in the logs 

## To Reproduce
**Affected versions**: 17.0.1.0.0

1. install and setup auth_signup_verify_email
2. install and setup google_recaptcha
3. try to create an account

**Expected behavior**
The account creation should proceed if the recaptcha passes

**Additional context**

flowersofpepper:
We are also facing this issue. Do you know if anyone has solved this issue?

flowersofpepper:
There seems to be a fix to this issue by @Cl0ut13R. Can someone please confirm if it's working and merge it to the main branch. Thank you.

Cl0ut1eR:
Hi, yeah, I created the issue with my work account but solved it on my personal one. Sorry for the confusion! It's working great on my end—we've been using the fork for two weeks with no issues on our 15 instances.

sans-eng:
I'm also has the same issue

Cl0ut1eR:
> I'm also has the same issue

Feel free to use my fork (don't auto update with it just in case), I won't update it unless there is an app we use that is updated, but if you/anyone want me to update the fork to the latest commit just ask.

PR #783, opened 2025-04-03
[FIX] auth_signup_verify_email: Use reCaptcha

Cl0ut1eR:
#774 
(previous pr with same changes #775 )

Fixes an issue when using the reCaptcha addon together with auth_signup_verify_email.
Currently, `passwordless_signup() `bypasses the reCaptcha check in the original `web_auth_signup() `method.

The root cause is that `_verify_request_recaptcha_token() `is supposed to remove the token from the request, but the same request is then used directly for user creation, resulting in: `ValueError: Invalid field 'recaptcha_token_response' on model 'res.users'`

### DIFF
--- a/auth_signup_verify_email/__manifest__.py
+++ b/auth_signup_verify_email/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Verify email at signup",
     "summary": "Force uninvited users to use a good email for signup",
-    "version": "17.0.1.0.0",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Isn't `reCaptcha` an extra module?

Cl0ut1eR:
> Isn't `reCaptcha` an extra module?

Yes but the original implementation of the method _verify_request_recaptcha_token is from base.

PR #790, opened 2025-05-13
[18.0][IMP] auth_saml: only write value that changes

vincent-hatakeyama:
When using mapping, not writing the value systematically avoids getting security mail on login/email changes when there is no change. Also use SQL for blanking passwords avoids the security update mails.

Port of #726


vincent-hatakeyama:
17.0 version was merged in February, this is only a port of the same fix.

/ocabot merge patch

PR #800, opened 2025-06-03
[16][cross_connect_server]No invitation email for cross server created user

florian-dacosta:
These users are not meant to be able to log in the normal way, it is always a redirection from another Odoo with automatic login

@fmr @sebastienbeau


sebastienbeau (approved):
LGTM (code review)

paradoxxxzero (approved):

PR #801, opened 2025-06-03
[17.0][IMP] auth_oidc: Add support for OIDC discovery

david-banon-tecnativa:
Adds support for OIDC Discovery.
Will forward port to 18 once merged.

@Tecnativa @andreschenkels @sbidoul

### DIFF
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -7,7 +7,7 @@ import secrets
 
 import requests
 
-from odoo import fields, models, tools
+from odoo import _, api, exceptions, fields, models, tools
 
 try:
     from jose import jwt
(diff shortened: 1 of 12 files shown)


auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
         help="Additional parameters for the auth link. "
         "For example: {'prompt':'select_account'}"
     )
+    oidc_discovery_url = fields.Char(string="OIDC config URL")
```
sbidoul:
```suggestion
    oidc_discovery_url = fields.Char(string="OIDC Discovery Document URL")
```

To use the standard terminology.

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
...
+                            % {"scope": scope}
+                        )
+            if scope not in self.scopes_supported:
+                raise exceptions.ValidationError(
+                    _("Scope %(scope)s is not supported by provider") % {"scope": scope}
+                )
+
+    def action_get_oidc_configuration(self):
+        if not self.supports_oidc_discovery:
+            return
```
sbidoul: raise a UserError instead ?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
         help="Additional parameters for the auth link. "
         "For example: {'prompt':'select_account'}"
     )
+    oidc_discovery_url = fields.Char(string="OIDC config URL")
+    supports_oidc_discovery = fields.Boolean(default=False)
+    auth_endpoint = fields.Char(required=False)
```
sbidoul, line 55: Why making this field required=False, since it is mandatory in the discovery document?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
...
+            return
+        r = requests.get(self.oidc_discovery_url, timeout=10)
+        r.raise_for_status()
+        response = r.json()
+        try:
+            self.auth_endpoint = response["authorization_endpoint"]
+            self.validation_endpoint = response["userinfo_endpoint"]
+        except KeyError:
+            raise exceptions.ValidationError(
+                _("Incorrect response from issuer, check OIDC config URL")
```
sbidoul:
```suggestion
                _("Incorrect response from issuer, check OIDC Discovery Document URL")
```

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
...
+            return
+        for scope in self.scope.split(" "):
+            if ":" in scope:
+                while len(scope) > 1:
+                    scope = ":".join(scope.split(":")[0:-1])
+                    if scope + ":*" not in self.scopes_supported:
+                        raise exceptions.ValidationError(
+                            _("Scope %(scope)s is not supported by provider")
+                            % {"scope": scope}
+                        )
```
sbidoul: Could you add some unit tests for this logic?

sbidoul (commented):
I had wanted to do this for a long time. Thanks for this!

A few minor suggestion.

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
...
+            if scope not in self.scopes_supported:
+                raise exceptions.ValidationError(
+                    _("Scope %(scope)s is not supported by provider") % {"scope": scope}
+                )
+
+    def action_get_oidc_configuration(self):
+        if not self.supports_oidc_discovery:
```
sbidoul, line 89:
```suggestion
        # https://openid.net/specs/openid-connect-discovery-1_0.html
        if not self.supports_oidc_discovery:
```

To link to the standard.

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
...
+        response = r.json()
+        try:
+            self.auth_endpoint = response["authorization_endpoint"]
+            self.validation_endpoint = response["userinfo_endpoint"]
+        except KeyError:
+            raise exceptions.ValidationError(
+                _("Incorrect response from issuer, check OIDC config URL")
+            ) from KeyError
+        self.jwks_uri = response.get("jwks_uri", False)
+        self.scopes_supported = response.get("scopes_supported", False)
```
sbidoul, line 104: According to https://openid.net/specs/openid-connect-discovery-1_0.html, scopes_supported is a json array, so this should be converted to a string?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -50,6 +50,46 @@ class AuthOauthProvider(models.Model):
         help="Additional parameters for the auth link. "
         "For example: {'prompt':'select_account'}"
     )
+    oidc_discovery_url = fields.Char(string="OIDC config URL")
+    supports_oidc_discovery = fields.Boolean(default=False)
+    auth_endpoint = fields.Char(required=False)
+    scopes_supported = fields.Char()
+
+    @api.constrains("scope")
+    def constrains_available_scopes(self):
```
sbidoul: The spec says *Servers MAY choose not to advertise some supported scope values even when this parameter is used*, so this validation should cause a warning instead?

PR #739, opened 2024-12-18
[IMP][16.0] auth_saml: download the provider metadata

dutrieuc:
Reopening of #647 
Migration to 16 of https://github.com/OCA/server-auth/pull/602 feature

PR #804, opened 2025-06-10
[IMP][18.0] auth_saml: download the provider metadata

dutrieuc:
Forward port of  #739, (originally #602)

> On Office365, what you get when configuring an application for SAML authentication is the URL of the federation metadata document. This URL is stable, but the content of the document is not. I suspect some of the encryption keys can be updated / renewed over time. The result is that the configured provider in Odoo suddenly stops working, because the messages sent by the Office365 provider can no longer be validated by Odoo (because the federation document is out of date). Downloading the new version and updating the auth.saml.provider record fixes the issue.
> 
> This PR adds a new field to store the URL of the metadata document. When this field is set on a provider, you get a button next to it in the form view to download the document from the URL. The button will not update the document if it has not changed.
> 
> Additionally, when a SignatureError happens, we check if downloading the document again fixes the issue.
>


vincent-hatakeyama (changes requested):
> Additionally, when a SignatureError happens, we check if downloading the document again fixes the issue.

This should be an option on the provider. If the metadata URL is set, user might not realise that the module would download the metadata again. That has security implications, if the IDP uri is modified, it could allow changing the metadata.

dutrieuc:
@vincent-hatakeyama 
>  If the metadata URL is set, user might not realise that the module would download the metadata again. 

we could make it explicit with the small paragraph by the field without adding an extra field ? Beside warning the user I don't think another field would be very useful as they can already use static IDP metadata by copy pasting it in the proper field.

vincent-hatakeyama:
I liked the idea to be able to download the metadata manually by using the URI field and a button.

This is clear enough to avoid unexpected behaviour from users.

vincent-hatakeyama (approved):

dutrieuc:
@vincent-hatakeyama I guess it's possible still, they can remove the url after the refresh, but a little hacky

I squashed and rebased my commits on top of 18.0, let me know if I need to do something else to merge.

vincent-hatakeyama (approved):

vincent-hatakeyama:
Note that as with all OCA PR, it needs a second review before merge.

dutrieuc:
hey @StefanRijnhart, since you reviewed #739 could you review this as well ?

StefanRijnhart (approved):
Thanks!

/ocabot merge minor

PR #805, opened 2025-06-16
[17.0][FIX] auth_api_key: Use 401 for unauthorized user

anusriNPS:
Port forward fix from 16.0 [#738](https://github.com/OCA/server-auth/pull/738)

As per documentation, using [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) as status code seems valid for a user who is not able to be authorized. However, [403](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) is used for scenario where user is forbidden to access a resource. So, returning 401 instead of 403.

### DIFF
--- a/auth_api_key/models/ir_http.py
+++ b/auth_api_key/models/ir_http.py
@@ -5,8 +5,9 @@
 
 import logging
 
+from werkzeug.exceptions import Unauthorized
+
 from odoo import models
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Straight forward fwport

/ocabot merge patch

PR #817, opened 2025-07-08
[16.0][FIX] users_ldap_populate: Decode login information

pedrobaeza:
When iterating through the users from LDAP, if the login information is read in bytes, passing the value in bytes to the psycopg2 library doesn't work, failing with this error:

```
psycopg2.errors.UndefinedFunction: operator does not exist: text = bytea
LINE 1: SELECT id FROM res_users WHERE lower(login)='...
```

so we need to decode the result before to get a pure string.

@Tecnativa TT56992


pedrobaeza:
Please review @david-banon-tecnativa @victoralmau

david-banon-tecnativa (approved):

victoralmau (approved):

PR #821, opened 2025-07-11
[17.0][IMP] auth_oauth_autologin: Move all auto login logic to the backend

wluyima:
With the current implementation of the oauth auto login feature that is handled by both front and back end logic, there is an annoying issue where the user temporarily sees the Odoo login page before they get redirected to the IDP's login page.

This PR modifies the oauth auto login addon so that the logic is implemented on the backend to resolve the issue.


sbidoul:
See https://github.com/OCA/server-auth/pull/717#issuecomment-2606740341 for why this won't work in 17. In 18 it should be ok to make this a backend redirect again

wluyima:
We have been using this feature for a while and we have not experienced that issue but I will test and verify.

Issue #815, opened 2025-07-04
[auth_jwt] Using database.secret as cookie secret


jesusVMayor:
Hi, i'm starting to work with auth_jwt with cookies, and I'm seeing some things that i see as inconsistent.

`_get_jwt_cookie_secret` function is returning allways the database.secret parameter, and is used to call `_encode` and `_decode` in `ir.http` functions, that means that we don't have a way to use the validator secret_key field as secret key, and there will be allways only one secret key for all validator and always will be the database secret.

- Why not use the secret_key field of the validator to encode and decode?
- Another thing that im don't know if is a bug, _auth_method_jwt is always setting the cookie again, with the `cookie_max_age` parameter of the validator, why not let the client handle the expiration of the cookie?

sbidoul:
Hi,

> Why not use the secret_key field of the validator to encode and decode?

This is because the secret_key field of the validator is meant for another purpose (i.e. decoding the JWT token itself). Actually, it may not even be set if we use public key to validate the JWT tokens.

In which scenario is it a problem to sign the cookie with `database.secret`?

> Another thing that im don't know if is a bug, _auth_method_jwt is always setting the cookie again, with the cookie_max_age parameter of the validator, why not let the client handle the expiration of the cookie?

This is by design so the client stays authenticated as long as it is using the API. But what do you mean exactly by "_why not let the client handle the expiration of the cookie?_"

jesusVMayor:
I see a problem if there is an external auth server, the server issues a token, the token its received in odoo and then odoo starts to issue new tokens in the cookie with a different signature, and expiry time. In every response the cookie is renewed, with a new expire in the payload and the cookie based on cookie_max_age field. I see it as a security issue, the JWT can't be invalidated, if the token is compromised an attacker could be accessing the system without a way to remove the access, the only way will be changing the database.secret, and this will casue problems with CSRF token, mail links... 
If you want to use the JWT to manage sesions, its needed to implement it in a more complex way, using also a refresh token who could be invalidated for example.

We must differentiate the authorization server and the resource server.

- **If odoo is the authorization server and the resource server in different controllers:** The token should be consistent, have a expiry time who requires a new authentication after the expiry time, or implement a refresh token saved in the user who could be invalidated.

[...]

PR #822, opened 2025-07-14
[16.0][IMP]auth_jwt: Add the possibility to not renew the cookie in the responses

jesusVMayor:
Fix the problems exposed in the issue https://github.com/OCA/server-auth/issues/815
- Allows to configure if the cookie is renewed or not
- If the cookie is renewed allow only to use secret as signature type.
- Use the validator secret to enconde.

### DIFF
--- a/auth_jwt/models/auth_jwt_validator.py
+++ b/auth_jwt/models/auth_jwt_validator.py
@@ -27,6 +27,12 @@ from ..exceptions import (
 _logger = logging.getLogger(__name__)
 
 AUTHORIZATION_RE = re.compile(r"^Bearer ([^ ]+)$")
+SECRET_ALGORITHM_SELECTION = [
+    # https://pyjwt.readthedocs.io/en/stable/algorithms.html
+    ("HS256", "HS256 - HMAC using SHA-256 hash algorithm"),
+    ("HS384", "HS384 - HMAC using SHA-384 hash algorithm"),
+    ("HS512", "HS512 - HMAC using SHA-512 hash algorithm"),
+]
 
 
 class AuthJwtValidator(models.Model):
@@ -39,12 +45,7 @@ class AuthJwtValidator(models.Model):
     )
     secret_key = fields.Char()
     secret_algorithm = fields.Selection(
-        [
-            # https://pyjwt.readthedocs.io/en/stable/algorithms.html
-            ("HS256", "HS256 - HMAC using SHA-256 hash algorithm"),
-            ("HS384", "HS384 - HMAC using SHA-384 hash algorithm"),
-            ("HS512", "HS512 - HMAC using SHA-512 hash algorithm"),
-        ],
+        SECRET_ALGORITHM_SELECTION,
         default="HS256",
     )
     public_key_jwk_uri = fields.Char()
@@ -96,6 +97,17 @@ class AuthJwtValidator(models.Model):
     )
     cookie_secure = fields.Boolean(
         default=True, help="Set to false only for development without https."
+    )
+    renew_cookie_on_response = fields.Boolean(
+        help="Renew the cookie in every response to stay the client "
+        "authenticatedas long it use the API. Don't mark unless you have a "
+        "way to invalidate sessions",
+        default=True,
+    )
+    renew_cookie_secret = fields.Char()
+    renew_cookie_algorithm = fields.Selection(
+        SECRET_ALGORITHM_SELECTION,
+        default="HS256",
     )
 
     _sql_constraints = [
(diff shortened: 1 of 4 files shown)


sbidoul:
Hi, I had not noticed #815 so far. Let's first discuss there.

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -97,11 +96,25 @@ class AuthJwtValidator(models.Model):
     cookie_secure = fields.Boolean(
         default=True, help="Set to false only for development without https."
     )
+    renew_cookie_on_response = fields.Boolean(
+        help="Renew the cookie in every response", default=True
```
sbidoul: Expand the help text a bit to explain the risk of setting to True?

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -97,11 +96,25 @@ class AuthJwtValidator(models.Model):
...
+    @api.constrains("signature_type", "renew_cookie_on_response")
+    def _check_signature_type_with_cookie_renew(self):
+        for rec in self:
+            if rec.renew_cookie_on_response and rec.signature_type != "secret":
+                raise ValidationError(
+                    _(
+                        "The sginature type should be secret if the cookie is "
+                        "renewed on responses"
+                    )
+                )
```
sbidoul: I don't think we should error in this case. Creating a http-only secure cookie once the client has been authenticated with the bearer token is a valid use case to me. If anything it makes the life of front-end developers easier (and more secure) as they don't need to worry with securely storing the JWT token, as the browser handles the cookie securely for them. I do agree with the caveat regarding session invalidation.

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -292,11 +305,7 @@ def unlink(self):
         return super().unlink()
 
     def _get_jwt_cookie_secret(self):
-        secret = self.env["ir.config_parameter"].sudo().get_param("database.secret")
-        if not secret:
-            _logger.error("database.secret system parameter is not set.")
-            raise ConfigurationError()
-        return secret
+        return self.secret_key
```
sbidoul: If the database secret is not adequate for your use case we could a new secret dedicated to signing/encrypting the cookie on the jwt validator record?

jesusVMayor: Modified encode and decode functions to use the fields renew_cookie_secret and renew_cookie_algorithm

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -97,6 +98,17 @@ class AuthJwtValidator(models.Model):
     cookie_secure = fields.Boolean(
         default=True, help="Set to false only for development without https."
     )
+    renew_cookie_on_response = fields.Boolean(
+        help="Renew the cookie in every response to stay the client "
+        "authenticatedas long it use the API. Don't mark unless you have a "
+        "way to invalidate sessions",
+        default=True,
+    )
+    renew_cookie_secret = fields.Char()
```
sbidoul, line 107:
```suggestion
    cookie_secret = fields.Char()
```

This secret is not specific to whether the cookie lifetime is renewed or not.

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -97,6 +98,17 @@ class AuthJwtValidator(models.Model):
...
+    renew_cookie_algorithm = fields.Selection(
+        SECRET_ALGORITHM_SELECTION,
+        default="HS256",
+    )
```
sbidoul, line 111: I'm not convinced we should propose an algorithm selection here.
For the JWT / bearer token, the algorithm must be agreed with the clients so an algorithm selection makes sense.
But here it's the server which protect the cookie for itself (the cookie is opaque to the clients), and we can let the server select the best algorithm.

sbidoul:
Thanks for your work on this. Would you split this PR in two that we can discuss independently? As the cookie renewal and the cookie secret are two independent topics, it would be easier to review.

auth_jwt/models/auth_jwt_validator.py
jesusVMayor: I see a relation with the renewal, if is not set, we dont want to change it. 
For example, i have  a domain.com who creates a cookie with the jwt for the domain, and i have odoo in odoo.domain.com, so the cookie will be shared between domain and subdomains. I want the validator to check that cookie against the secret/public key, not against that new secret, the cookie should only be checked against that secret if odoo creates its own cookie when the renew is checked.

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -291,13 +317,6 @@ def unlink(self):
         self._unregister_auth_method()
         return super().unlink()
 
-    def _get_jwt_cookie_secret(self):
-        secret = self.env["ir.config_parameter"].sudo().get_param("database.secret")
-        if not secret:
-            _logger.error("database.secret system parameter is not set.")
-            raise ConfigurationError()
-        return secret
```
sbidoul, line 299: I prefer keeping this method and return the validator `cookie_secret` if set else the database secret.

auth_jwt/models/ir_http.py
```diff
@@ -91,15 +91,14 @@ def _auth_method_jwt(cls, validator_name=None):
                 raise list(exceptions.values())[0]
             raise UnauthorizedCompositeJwtError(exceptions)
 
-        if validator.cookie_enabled:
+        if validator.cookie_enabled and validator.renew_cookie_on_response:
```
sbidoul, line 94: This change does not seem necessary.

auth_jwt/models/ir_http.py
```diff
@@ -65,7 +65,7 @@ def _get_jwt_payload(cls, validator):
                 raise
             token = cls._get_cookie_token(validator.cookie_name)
             assert token
-            return validator._decode(token, secret=validator._get_jwt_cookie_secret())
+            return validator._decode(token, cookie_secret=True)
```
sbidoul, line 68: This change is not necessary.

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -163,26 +175,40 @@ def _get_key(self, kid):
         jwks_client = PyJWKClient(self.public_key_jwk_uri, cache_keys=False)
         return jwks_client.get_signing_key(kid).key
 
-    def _encode(self, payload, secret, expire):
+    def _encode(self, payload, expire, secret=False):
```
sbidoul, line 178: I would not change this method. It is not necessary to change it if all the logic to get the secret is in `_get_jwt_cookie_secret()`.

auth_jwt/models/auth_jwt_validator.py
sbidoul: Yes that is your use case, but the two changes are logically independent. Deciding whether renewing the cookie automatically and how to sign it are two independent topics. Reasoning about it that way should make the change simpler while still supporting your use case I think.

jesusVMayor:
https://github.com/OCA/server-auth/pull/824
https://github.com/OCA/server-auth/pull/825

auth_jwt/models/ir_http.py
jesusVMayor: i don't understand why, we need it to not renew the cookie.

sbidoul:
Thanks for splitting the PR!

PR #823, opened 2025-07-14
[18.0][MIG] auth saml create user

tm-smile:
Migration of module auth_saml_create_user to 18.0 from 11.0


vincent-hatakeyama:
This module is rendered useless by #695 which does the same feature. The PR just needs to be reviewed then ported to 18.0.

PR #824, opened 2025-07-17
[16.0][IMP]Allow to configure cookie secret in the validator

jesusVMayor:


sbidoul (commented):
This one looks good to me. 

Could you add a test? A simple unit test on _get_jwt_cookie_secret is probably sufficient. If you want to go further, some other integration tests are in `auth_jwt_demo`.

sbidoul:
Sorry for the delay, I had lost track of this.

PR #825, opened 2025-07-17
[16.0][IMP]auth_jwt: Allow to not renew the cookie in the responses

jesusVMayor:


auth_jwt/models/auth_jwt_validator.py
```diff
@@ -97,6 +97,12 @@ class AuthJwtValidator(models.Model):
     cookie_secure = fields.Boolean(
         default=True, help="Set to false only for development without https."
     )
+    renew_cookie_on_response = fields.Boolean(
+        help="Renew the cookie in every response to stay the client "
+        "authenticatedas long it use the API. Don't mark unless you have a "
+        "way to invalidate sessions",
+        default=False,
```
sbidoul:
```suggestion
        default=True,
```

for backward compatibility

auth_jwt/models/ir_http.py
```diff
@@ -65,7 +65,10 @@ def _get_jwt_payload(cls, validator):
                 raise
             token = cls._get_cookie_token(validator.cookie_name)
             assert token
-            return validator._decode(token, secret=validator._get_jwt_cookie_secret())
+            secret = None
+            if validator.renew_cookie_on_response:
+                secret = validator._get_jwt_cookie_secret()
```
sbidoul, line 70: I don't understand this part. If we don't check that the cookie is signed with the expected secret, a malicious client could send a forged cookie, right?

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -97,6 +97,12 @@ class AuthJwtValidator(models.Model):
     cookie_secure = fields.Boolean(
         default=True, help="Set to false only for development without https."
     )
+    renew_cookie_on_response = fields.Boolean(
+        help="Renew the cookie in every response to stay the client "
+        "authenticatedas long it use the API. Don't mark unless you have a "
+        "way to invalidate sessions",
```
sbidoul:
```suggestion
        help=(
            "Renew the cookie in every response to keep the client "
            "authenticated as long as it uses the API. Don't mark unless you have a "
            "way to invalidate sessions"
        ),
```

fixing a couple of typos

auth_jwt/models/ir_http.py
```diff
@@ -91,7 +94,7 @@ def _auth_method_jwt(cls, validator_name=None):
                 raise list(exceptions.values())[0]
             raise UnauthorizedCompositeJwtError(exceptions)
 
-        if validator.cookie_enabled:
+        if validator.cookie_enabled and validator.renew_cookie_on_response:
```
sbidoul, line 97: This completely disable the cookie setting mechanism is `renew_cookie_on_response` is not set, so that's probably not what you intended?

auth_jwt/models/ir_http.py
jesusVMayor: No, if renew_cookie_on_response is False, _decode will be called with secret=None, so the funcion will check the cookie with the secret or public key configured in the validator.

auth_jwt/models/ir_http.py
jesusVMayor: yes, is the idea.  We need a flow where odoo only checks the cookie/header and validate against the secret/public key configured in the validator, we dont need the cookie to be setted again in the response

sbidoul: I don't get it, sorry. What will then set the cookie in the first place?

jesusVMayor: Another service in the same domain, the cookie can be shared, or odoo itself in a public controller who checks the credentials and sets the cookie.

Our goal is to implement a flow in the module where it only checks the token and give access to the endpoint. 
This can be done with the token in the header, but is a problem with cookies, because odoo set it again with his own secret and also extends the expire time.

sbidoul:
I'm not comfortable with this one, because what it does does not match the name of the new option.

jesusVMayor:
How about issue_cookie? Or external_cookie, reversing the boolean logic?

PR #827, opened 2025-08-04
Fix typo in Italian translation for "Log in with Microsoft"

filippolmt:


sbidoul:
Normally, all translations should be submitted via https://translation.odoo-community.org to avoid merge conflict.
This is small, so merging right way, but next time, please do it over there.

PR #829, opened 2025-08-08
[16][cross_connect_server] Make external user as inactive to avoid pollution in users and tables across Odoo

florian-dacosta:
@sebastienbeau @paradoxxxzero 
Do you see any reason to have this kind of users active ?

It is an issue, because, with users from multiple instance we could (we will) have users with the same name.
It then pollutes the partner and user database.


paradoxxxzero (approved):
LGTM

florian-dacosta:
Setting users as inactive seems finally to invasive. 
It requires to set active test = False on some fields and also it breaks the mail notification (no mail sent to unactive followers).
Misc other features may beak because of this, it seems finally safer to keep users active and workaround the partner pollution...

PR #830, opened 2025-08-13
[18.0][MIG] auth_ldap_attribute_sync: Migration to 18.0

ap-wtioit:


ap-wtioit:
Tests are failing because Odoo core auth_ldap still has `'python': ['ldap'],` declared as external dependency:
https://github.com/odoo/odoo/blob/18.0/addons/auth_ldap/__manifest__.py

PR #831, opened 2025-08-18
installable > true

Ruben-Techstate:


flotho (changes requested):
Thanks for this PR.
This is probably your first PR for OCA and we would like to thank you for this.
Some changes must be applied as mentionnend in the guiidelines https://github.com/OCA/server-auth?tab=contributing-ov-file
basically Make PR for each module, change your pr names, and also motivate the reason of the change.

PR #832, opened 2025-08-18
[16][IMP] auth_oidc: add `end_session_endpoint`

danypr92:
Improved `auth_oidc` module with the logout mechanism implemented in version 14.0 in commit 9b6f5eed7932bfcabfe5ac4887119f8f94a10939 and adapt to the 16.0 logout flow.

Co-authored-by: @konykon

### DIFF
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -49,6 +49,7 @@ class AuthOauthProvider(models.Model):
     auth_link_params = fields.Char(
         help="Additional parameters for the auth link. For example: {'prompt':'select_account'}"
     )
+    end_session_endpoint = fields.Char(string="End Session URL")
 
     @tools.ormcache("self.jwks_uri", "kid")
     def _get_keys(self, kid):
--- a/auth_oidc/views/auth_oauth_provider.xml
+++ b/auth_oidc/views/auth_oauth_provider.xml
@@ -18,6 +18,7 @@
             <field name="validation_endpoint" position="after">
                 <field name="token_endpoint" />
                 <field name="jwks_uri" />
+                <field name="end_session_endpoint" />
             </field>
             <field name="auth_endpoint" position="after">
                 <field name="auth_link_params" />


danypr92:
Hi @sbidoul, can you review these changes?

konykon:
Hi @sbidoul could you review this PR?

sbidoul:
Hi,

Thanks for the contrib and your patience.

I think adding the `end_session_endpoint` field in `auth_oidc` makes sense as it is part of the spec, and it could be populated from the discovery document in #801.

What to do with it, I'm less sure as I'm under the impression that there are many way to handle logout. Should it close the Odoo session? should it also logout from the provider? For this app only or should it do a complete SSO logout? What if we use `auth_oauth_multi_token`? etc... 

So there are so many possible approaches that I think it is better to explore them in separate modules that can depend on auth_oidc. So at this stage I think it is best to explore different approaches outside of auth_oidc (except adding the `end_session_endpoint` field as I mentioned).

danypr92:
Hi @sbidoul, thanks for your response!

We answer inline with your suggestions:

> I think adding the end_session_endpoint field in auth_oidc makes sense as it is part of the spec, and it could be populated from the discovery document in https://github.com/OCA/server-auth/pull/801.

We can keep the field and add it to the discovery flow, but we are thinking if it makes sense to have a field in this module that does not work without installing an additional "auth_oidc_logout" module... Maybe if we separate the logout flow in another module, the end_session endpoint field has to be in that module. In the other hand, the logout session can be complex, but we think that the base flow (end the Odoo session and the provider session) should be in this module. Then we can generate another module to cover the complexities in that flow.

> What to do with it, I'm less sure as I'm under the impression that there are many way to handle logout. Should it close the Odoo session? should it also logout from the provider? For this app only or should it do a complete SSO logout? What if we use auth_oauth_multi_token? etc...

We can investigate the OIDC end session flow and purpose a new module that cover the multi_token logout flow and that allow to choose if you want to end the provider session or only the Odoo session. But we want to separate this approach and discussion of this PR, because this PR only migrates the functionalities from old versions.

sbidoul:
I'm not worried with having the logout endpoint url in the base module. It allows different logout flows to be implement in different modules while using a common base.

> the base flow (end the Odoo session and the provider session) should be in this module

The thing is I'm not even sure that that is the base flow.

danypr92:
Ok, we can keep the logout endpoint url in this module and then discuss in a different issue the implementations of the logout channel to be implemented in another module.

danypr92:
Done!

The discovery flow is not implemented in 16.0, #801 is an implementation on 17.0. Maybe we can port it also, but in a separate PR.

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -49,6 +49,12 @@ class AuthOauthProvider(models.Model):
     auth_link_params = fields.Char(
         help="Additional parameters for the auth link. For example: {'prompt':'select_account'}"
     )
+    end_session_endpoint = fields.Char(
+        string="End Session URL",
+        help="If set, the user is logged out in the authorization provider upon logout "
+        "in the client, should be the value of end_session_endpoint specified by "
+        "the authorization provider.",
```
sbidoul: I'd remove the help text, as what this does precisely is not defined in this module.

danypr92: Done!

sbidoul:
Thanks!

/ocabot merge minor

sbidoul:
Forward ports in #856 and #857

PR #833, opened 2025-08-21
[18.0][MIG] auth_ldaps: Migration to 18.0

BT-pkosnik:


auth_ldaps/models/res_company_ldap.py
```diff
@@ -0,0 +1,55 @@
...
+import ldap
+
+from odoo import fields, models
+from odoo.tools.misc import str2bool
+
+_logger = logging.getLogger(__name__)
+
+
+class CompanyLDAP(models.Model):
+    _inherit = "res.company.ldap"
```
BT-vstoev, line 17: The original model has a method `def test_ldap_connection(self):` that is used from a button on the form view.
Inside that method the `conf` is hardcoded and because of that it is missing the new keys that are introduced in this PR "is_ssl" and "skip_cert_validation". Because of that actually with this PR the button is broken (because it calls `_connect` with the hardcoded `conf` and `if conf["is_ssl"]` throws an error).
As a solution I would propose to add context to the button in the form, and if that context is present inside the `_connect` call (meaning it's a call from the button) - fill the missing conf values from self.

BT-vstoev (changes requested):
Hello and thanks for the work!
I performed a Test on a local Odoo 19 instance and found out a functional problem.
Please check my comment about the "Test connection" button on the form view. It didn't work for me. I left a comment with my understanding of the issue.

lorenzomorandini:
Hi, is there a plan to finish and merge this?

PR #836, opened 2025-09-01
[ADD] auth_saml: Support for unsolicited SAML requests

bringsvor:
Add configurable feature to allow unsolicited SAML requests. This is for example useful to allow IdP to initiate.

### DIFF
--- a/auth_saml/models/auth_saml_provider.py
+++ b/auth_saml/models/auth_saml_provider.py
@@ -140,8 +140,7 @@ class AuthSamlProvider(models.Model):
     allow_saml_unsolicited_req = fields.Boolean(
         compute="_compute_allow_saml_unsolicited",
         string="Allow Unsolicited Requests",
-        help="Allow IdP-initiated authentication requests without prior "
-        "AuthnRequest from SP",
+        help="Allow IdP-initiated authentication requests"        
     )
 
     @api.model
--- a/auth_saml/models/auth_saml_provider.py
+++ b/auth_saml/models/auth_saml_provider.py
@@ -140,7 +140,7 @@ class AuthSamlProvider(models.Model):
     allow_saml_unsolicited_req = fields.Boolean(
         compute="_compute_allow_saml_unsolicited",
         string="Allow Unsolicited Requests",
-        help="Allow IdP-initiated authentication requests without prior AuthnRequest from SP"
+        help="Allow IdP-initiated authentication requests without prior AuthnRequest from SP",
     )
 
     @api.model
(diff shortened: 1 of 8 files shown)


auth_saml/models/auth_saml_provider.py
```diff
@@ -137,6 +137,12 @@ class AuthSamlProvider(models.Model):
         help="Whether metadata should be signed or not",
     )
 
+    allow_saml_unsolicited_req = fields.Boolean(
+        compute="_compute_allow_saml_unsolicited",
+        string="Allow Unsolicited Requests",
+        help="Allow IdP-initiated authentication requests",
+    )
+
```
vincent-hatakeyama, line 145: Why make this a company dependent value and not a provider dependant one?

If it is company dependent, it should only be in the configuration panel.

auth_saml/models/auth_saml_provider.py
```diff
@@ -219,7 +225,7 @@ def _get_config_for_provider(self, base_url: str = None) -> Saml2Config:
                             (acs_url, saml2.BINDING_HTTP_POST),
                         ],
                     },
-                    "allow_unsolicited": False,
+                    "allow_unsolicited": self.allow_saml_unsolicited_req,
```
vincent-hatakeyama, line 228: I’m not confident the company is correct at this place, I believe it might be using the superuser company instead of any company.

auth_saml/tests/test_unsolicited_requests.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (C) 2010-2016, 2022 XCG Consulting <http://odoo.consulting>
```
vincent-hatakeyama, line 1:
```suggestion
# Copyright (C) 2010-2016, 2022 XCG SAS <https://orbeet.io/>
```

No need to keep the copyright unless you copied something from another test.
You should add your name or your company’s.

auth_saml/views/res_config_settings.xml
```diff
@@ -13,6 +13,13 @@
                     id="module_auth_saml"
                 >
                     <field name="allow_saml_uid_and_internal_password" />
+                    <!-- Allow IdP-initiated authentication without prior AuthnRequest from SP.
+                         Security Note: Only enable if your IdP requires unsolicited requests
+                         and you trust the IdP to properly validate user sessions. -->
```
vincent-hatakeyama, line 18: This should not be a comment, it should be displayed in the configuration page.

vincent-hatakeyama (changes requested):
Thanks for the pull request.

A couple of additional review elements:

- The pull request lacks the version tag (“[18.0]”). It should be [IMP] rather than [ADD].
- Squash/fixup your commits into a single one
- Commit title should include [IMP] rather than [FEAT] (see https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message)

I also have a question about testing this: is there a way to test this feature with keycloak?

PR #837, opened 2025-09-02
[18.0][IMP]auth_oidc: verify self-signed certificates

ChristophAbenthungCibex:
If the connection between odoo and an oauth provider uses self-signed certificates, a ssl error is thrown because the self-signed certificated cannot be verified.
Even if the user credentials are correct, the user will not get logged in. 

Following error gets thrown:
requests.exceptions.SSLError: HTTPSConnectionPool(host='***', port=443): Max retries exceeded with url: /realms/***/protocol/openid-connect/token (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1000)')))

Steps to reproduce:
1. Setup a oauth provider (for example keycloak)
2. Setup odoo the oauth provider in odoo (used OpenID Connect (authorization code flow) as Auth Flow for my tests)
3. Logout
4. Try to login with the oauth provider
5. Type in the correct credentials
6. The oauth provider will redirect to Odoo
7. Odoo shows an error message and the user is not logged in


auth_oidc/models/auth_oauth_provider.py
```diff
@@ -46,10 +46,21 @@ class AuthOauthProvider(models.Model):
         string="Token URL", help="Required for OpenID Connect authorization code flow."
     )
     jwks_uri = fields.Char(string="JWKS URL", help="Required for OpenID Connect.")
+    self_signed = fields.Boolean(
+        string="Self-signed", help="Defines if the used certificate is self-signed."
```
ap-wtioit: In the backend i think it would be better to give the admin a hint what this does, like:
> Disable certificate checks for server to server token requests when using self signed certificates.

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -46,10 +46,21 @@ class AuthOauthProvider(models.Model):
...
+    )
+    self_signed_verify = fields.Char(
+        string="Self-signed verify path",
+        help="Path to the self-signed certificate for the verification process."
+        "Empty value disables the verification.",
```
ap-wtioit: There is a space missing in the multi line string between the two sentences:
image

ap-wtioit (changes requested):
Looks good (just the missing space is a requirement for the PR to be correct from my point of view)

This is a useful addition for checking with local HTTPS were one cannot really get official certificates.

auth_oidc/models/auth_oauth_provider.py
ChristophAbenthungCibex: Thank you for the suggestion. Changed the help text. :ok_hand:

auth_oidc/models/auth_oauth_provider.py
ChristophAbenthungCibex: Added the whitespace. Thanks

ap-wtioit (approved):

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -46,10 +46,23 @@ class AuthOauthProvider(models.Model):
         string="Token URL", help="Required for OpenID Connect authorization code flow."
     )
     jwks_uri = fields.Char(string="JWKS URL", help="Required for OpenID Connect.")
+    self_signed = fields.Boolean(
+        string="Self-signed",
+        help="Disable certificate checks for server to server token requests "
+        "when using self signed certificates.",
+    )
```
sbidoul, line 52: I think this does not make it sufficiently clear that this enables an insecure option. Could we name this field, say, "insecure" like curl does?

auth_oidc/models/auth_oauth_provider.py
```diff
@@ -46,10 +46,23 @@ class AuthOauthProvider(models.Model):
...
+    self_signed_verify = fields.Char(
+        string="Self-signed verify path",
+        help="Path to the self-signed certificate for the verification process. "
+        "Empty value disables the verification.",
+    )
```
sbidoul, line 56: I would name this field `ca_bundle`, but I have a strong preference to let that be configured at the system level, or with the `REQUESTS_CA_BUNDLE` environment variable.

sbidoul (commented):
Thanks. A couple a comment after a quick look.

auth_oidc/models/auth_oauth_provider.py
ap-wtioit: For me the benefit here would be that this is only for one OAuth server connection while REQUESTS_CA_BUNDLE is for all requests from Odoo to other servers.

CRogos: Instead of the attributes **self_signed** and **self_signed_verify**, how about only have **ca_bundle**, where you can add your valid self signed certificates (automatically used, when set), and a **disable_certificate_check** which bypasses the check. (same as self_signed=True + self_signed_verify="")

I think this would be more clear and less risky for false configurations.

auth_oidc/models/auth_oauth_provider.py
CRogos: I agree with sbidoul. If we add this kind of feature we should add a big warning in the descripiton.md and add a reference here.

ChristophAbenthungCibex: Thanks for the feedback. I renamed the fields and added a security note. Is that good enough?

auth_oidc/models/auth_oauth_provider.py
ChristophAbenthungCibex: Thanks, I renamed the field as suggested by @CRogos . I agree that this way is less risky.

CRogos (commented):
Implementation LGTM.
I am not sure if the merge commit is a problem, or if it would be better to rebase your changes.

Could you also add some tests for your changes?

ChristophAbenthungCibex:
> Implementation LGTM. I am not sure if the merge commit is a problem, or if it would be better to rebase your changes.
> 
> Could you also add some tests for your changes?

I rebased it. 

I'll check if I can add tests.

auth_oidc/models/res_users.py
```diff
@@ -38,6 +43,7 @@ def _auth_oauth_get_tokens_auth_code_flow(self, oauth_provider, params):
             ),
             auth=auth,
             timeout=10,
+            verifiy=verify,
```
ap-wtioit: typo here should be `verify=verify` (from test errors)

ChristophAbenthungCibex: oh yea missed that. Thx

PR #844, opened 2025-10-02
[19.0][MIG] auth_jwt: Migration to 19.0

pablo-cort-s73:
## Migration to v19.0

- Update the module version from `18.0.1.0.0` to `19.0.1.0.0`.

## Technical changes
- Change of the version field in the manifest to ‘19.0.1.0.0’.
- Replaced the SQL constraint `_sql_constraints` with the new syntax `models.Constraint(‘UNIQUE(name)’, ...)`.
- Correction of attribute calls in the request context (e.g. request.env.uid instead of request.uid).
- Updated the use of translation functions, changing from `_()` to `self.env._()` with modernised parameter formatting.
- Removed a pylint comment for the jwt dependency.
- In the test file, the search for an existing partner was replaced by the explicit creation of a test partner.

## Changes to exceptions
- Improved the format of error messages to use the new translation system with placeholders.


auth_jwt/models/auth_jwt_validator.py
```diff
@@ -246,7 +245,14 @@ def _get_and_check_partner_id(self, payload):
...
         res = super()._register_hook()
-        self.search([])._register_auth_method()
+        BATCH_SIZE = 100
+        offset = 0
+        while True:
+            validators = self.search([], limit=BATCH_SIZE, offset=offset)
+            if not validators:
+                break
+            validators._register_auth_method()
+            offset += BATCH_SIZE
```
sbidoul: This is not worth the added complexitty. In this case we know that there will never be thousands of validators so `search([])` is fine. Let's silence the pylint warning instead `# pylint: disable=...` on the search line.

sbidoul (changes requested):
Thanks for this migration! Looks good, just a minor comment.

I mark it as change requested, just because it is important to not merge this until we are sure that the tests in `auth_jwt_demo` pass too (this was forgotten in 17 and 18). Would you like to migrate it too, starting from #846? This migration can be done in this same PR since the modules are so closely related.

pablo-cort-s73:
> Thanks for this migration! Looks good, just a minor comment.
> 
> I mark it as change requested, just because it is important to not merge this until we are sure that the tests in `auth_jwt_demo` pass too (this was forgotten in 17 and 18). Would you like to migrate it too, starting from #846? This migration can be done in this same PR since the modules are so closely related.

Hello @sbidoul , I have migrated the demo module, but it seems that it does not load the demo data in the tests, since in the first test where it checks if the validator exists, it no longer finds it.
I have tried to replicate it locally and have managed to do so by manually loading the demo data from the Odoo settings, but not by initialising the auth_jwt_demo module. 
Do you know why it is not loading in the OCA tests?

Best regards,

sbidoul:
Indeed Odoo 19 does not load demo data by default. The general thinking is that this is a good thing and that tests should work independently of demo data. It makes sense to me that tests should not depend on demo data of other modules, but I kind of think depending on demo data of the module itself is okay.

I'm not sure what to do with that here yet. I'll need to look closer.

Ah, I see you have tried to make the linter happy on the js lib. But since that one is copied from elsewhere, it's better to keep it untouched and exclude it in `.pre-commit-config.yaml` like I did in https://github.com/OCA/server-auth/pull/846/commits/9762fdff57c14846a4ca072ccce2c5c825323d1a.

pablo-cort-s73:
Hello @sbidoul, have you seen or do you know how we can launch the demo data for the tests?
Can the demo data be launched from the test setup in this case?

auth_jwt/models/auth_jwt_validator.py
```diff
@@ -0,0 +1,316 @@
...
+    cookie_path = fields.Char(default="/")
+    cookie_max_age = fields.Integer(
+        default=86400 * 365,
+        help="Number of seconds until the cookie expires (Max-Age).",
+    )
+    cookie_secure = fields.Boolean(
+        default=True, help="Set to false only for development without https."
+    )
+
+    models.Constraint("UNIQUE(name)", "JWT validator names must be unique !")
```
ferran-S73:
```suggestion
    _name_uniq = models.Constraint("UNIQUE(name)", "JWT validator names must be unique !")
```

thib-d:
### Problem
When using Keycloak as an OAuth/OIDC provider via `Migration to 19.0 #844`, Odoo users can authenticate successfully, but logging out from Odoo does **not** log the user out from the Identity Provider (Keycloak).

As a result, when the user clicks "Log out" in Odoo and then goes back to the login page, clicking the SSO button logs them back in immediately because the Keycloak session is still active.

This is a common expectation for OIDC-based SSO: Odoo should support **RP-Initiated Logout** (OIDC end_session_endpoint), or at least offer a way to redirect to the IdP logout endpoint.

### Environment
- Odoo: 19.0 (Docker image `odoo:19.0`)
- Module: `auth_oauth`
- Identity Provider: Keycloak (OIDC)
- Provider endpoints (example):
  - authorization_endpoint: `.../protocol/openid-connect/auth`
  - token_endpoint: `.../protocol/openid-connect/token`
  - userinfo_endpoint: `.../protocol/openid-connect/userinfo`
  - end_session_endpoint: `.../protocol/openid-connect/logout`

### Steps to reproduce
1. Configure OAuth provider pointing to Keycloak.
2. Login to Odoo using the SSO button.
3. Click "Log out" in Odoo.
4. Go back to the Odoo login page and click SSO again.

### Actual behavior
User is logged in again immediately because Keycloak session was not terminated.

### Expected behavior
On Odoo logout, user should also be logged out of Keycloak (or Odoo should redirect to IdP logout endpoint).

### Additional notes / suggestion
- Odoo currently handles OAuth login but does not implement OIDC RP-initiated logout.
- Possible improvements:
  1. Add optional configuration fields on the OAuth provider:
     - `end_session_endpoint`
     - `post_logout_redirect_uri`
     - optional `id_token_hint` support if available
  2. On `/web/session/logout`, redirect to `end_session_endpoint` when the user logged in via an OIDC provider.
- At minimum, providing a supported way to configure a logout redirect without patching Odoo would help production deployments.

Thanks!

pablo-cort-s73:
@sbidoul can you review, pls?
thanks

pablo-cort-s73:
@pedrobaeza i have resolved this PR, but i don't have response, can you review or remove the "stale tag" pls?

pedrobaeza:
Removed the label but @sbidoul should approve it.

pablo-cort-s73:
Great, thank you @pedrobaeza 
Yes, I mentioned it to him; I suppose he’ll look into it when he can.

Reyes4711-S73 (approved):
LGTM

Alexgars73 (approved):

PR #848, opened 2025-10-10
[18.0][FIX] auth_saml: do not crash if no users

len-foss:
If self is empty, the syntax is invalid and will crash. Since there is no `ensure_one`, this case should normally not cause any problems.


vincent-hatakeyama:
Thank you for your PR.

Minor detail : the name of the target branch/odoo version should be in bracket in the PR title.

len-foss:
@vincent-hatakeyama no problems, updated :-)

PR #858, opened 2025-10-31
Draft: [18.0][IMP] auth_oidc: add groups' handling [NOT READY]

OdyX:
This allows groups' handling via a token's attributes as passed by an OIDC IdP (tested with both Keycloak and Gitlab)

This is the 18.0-version of these pulls by @26houseRobot and @hbrunn, with some minor fixes
* https://github.com/OCA/server-auth/pull/682
* https://github.com/OCA/server-auth/pull/372
* https://github.com/OCA/server-auth/pull/350, then

It's exactly the same commits (+ rebase on top of 18.0 ) as https://github.com/OCA/server-auth/pull/682

EDIT: Eh. Turns out it doesn't work straightforwardly on 18. Will update this description as I fix this.

### DIFF
--- a/auth_oidc/__manifest__.py
+++ b/auth_oidc/__manifest__.py
@@ -16,6 +16,10 @@
     "summary": "Allow users to login through OpenID Connect Provider",
     "external_dependencies": {"python": ["python-jose"]},
     "depends": ["auth_oauth"],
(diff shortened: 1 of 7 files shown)


OdyX:
I'll reopen this when the tests pass. Sorry for the noise.

PR #869, opened 2025-11-26
[FIX] [18.0] impersonate_login: session_info should use self.env.user

PieterPaulussen:


StefanRijnhart (commented):
Thanks! Can you explain into some detail what issue this fixes and how it can be reproduced?

PR #900, opened 2026-02-20
[16.0][MIG] auth_keycloak

hbrunn:
this only contains the part of the module that pertains to user creation, everything else is handled already by `auth_oidc`.

I added a commit that allows the admin to have keycloak send a password reset mail on user creation.


Stephanvandenberg (approved):
We are using this feature quite extensively and it has worked flawlessly. Would be good to have it as part of the standard server-auth repo.

PR #905, opened 2026-02-25
[19.0][MIG] auth_signup_verify_email: Migration to 19.0

F4LC0NN:
# auth_signup_verify_email: Migration to 19.0
* Fixed signup xml template for the login field
* Fixed translations symbol in main.py to adapt to version 19.0 specifications
* Fixed cursor invocation from env in main.py according to version 19.0 specifications


F4LC0NN (commented):
Auto Review OK

kwoychesko:
Thanks for the migration, @F4LC0NN! One thing worth flagging on the view — I think it's a behavioral regression in 19.0:

`only_passwords` was **removed from `auth_signup` core in 19.0**. In 18.0 the `auth_signup.signup` and `auth_signup.reset_password` templates set it (`<t t-set="only_passwords" .../>`), but those `t-set`s are gone in 19.0 and the variable isn't defined anywhere now. So the `t-if="only_passwords"` added here resolves against an undefined name, which QWeb renders as falsy (no error) — meaning the password / confirm-password fields end up **always hidden**.

Since `auth_signup.fields` is also rendered (with a token) on the **reset-password page** and on **invited/token signup**, this hides the password inputs there too: a user following a password-reset link or an invitation link gets a form with no password fields and can't set a password. The two tests don't catch it because neither renders the token/password-fields state.

A fix that preserves the passwordless-for-uninvited behavior without relying on the removed flag — gate the fields call on a valid token instead:

```xml
<template id="signup_fields" inherit_id="auth_signup.signup">
    <!-- show password fields only for invited/token signups; uninvited = passwordless -->
    <xpath expr="//t[@t-call='auth_signup.fields']" position="attributes">
        <attribute name="t-if">token and not invalid_token</attribute>
    </xpath>
    <xpath expr="//input[@name='login']" position="attributes">
        <attribute name="type">email</attribute>
    </xpath>
</template>
```

This leaves `auth_signup.reset_password`'s own (already token-gated) call to `auth_signup.fields` untouched, so password reset and invited signup keep working. Verified on 19.0.

PR #909, opened 2026-03-13
auth_oidc: models: res_users: Send cleint secret as well

okawo80085:
Otherwise most OIDC providers will fail the token endpoint call

This is a bug fix btw

### DIFF
--- a/auth_oidc/models/auth_oauth_provider.py
+++ b/auth_oidc/models/auth_oauth_provider.py
@@ -38,6 +38,9 @@ class AuthOauthProvider(models.Model):
     client_secret = fields.Char(
         help="Used in OpenID Connect authorization code flow for confidential clients.",
     )
+    client_secret_post = fields.Boolean(
+        help="Use Client Secret in authorization post requests"
+    )
     code_verifier = fields.Char(
         default=lambda self: secrets.token_urlsafe(32), help="Used for PKCE."
     )
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -27,16 +27,21 @@ class ResUsers(models.Model):
         auth = None
         if oauth_provider.client_secret:
             auth = (oauth_provider.client_id, oauth_provider.client_secret)
+
+        post_data = dict(
+            client_id=oauth_provider.client_id,
+            grant_type="authorization_code",
+            code=code,
+            code_verifier=oauth_provider.code_verifier,  # PKCE
+            redirect_uri=request.httprequest.url_root + "auth_oauth/signin",
+        )
+
+        if oauth_provider.client_secret_post:
+            post_data["client_secret"] = oauth_provider.client_secret
+
         response = requests.post(
             oauth_provider.token_endpoint,
-            data=dict(
-                client_id=oauth_provider.client_id,
-                client_secret=oauth_provider.client_secret,
-                grant_type="authorization_code",
-                code=code,
-                code_verifier=oauth_provider.code_verifier,  # PKCE
-                redirect_uri=request.httprequest.url_root + "auth_oauth/signin",
-            ),
+            data=post_data,
             auth=auth,
             timeout=10,
         )
--- a/auth_oidc/models/res_users.py
+++ b/auth_oidc/models/res_users.py
@@ -31,6 +31,7 @@ class ResUsers(models.Model):
             oauth_provider.token_endpoint,
             data=dict(
                 client_id=oauth_provider.client_id,
+                client_secret=oauth_provider.client_secret,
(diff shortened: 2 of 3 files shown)


okawo80085:
Oh yea and it's also needed in 19.0 and any other version that wants to use OIDC providers that verify client secrets when processing token calls

sbidoul:
Not sure it is a bug. Possibly a missing feature.

Section 9. here seems relevant: https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication
It says the default authentication method is `client_secret_basic`, which we use here.
What you want seems to be `client_secret_post`.

Actually I'm not sure it is ok to send the client secret in the body unconditionally.

https://www.rfc-editor.org/rfc/rfc6749.txt section 2.3.1 says *Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED*.

So yeah, I think this is not obvious at all and requires more investigation.

Do you have more evidence to share?

okawo80085:
Well, with PocketID (the OIDC provider im running) it's not optional, so if not unconditionally, it should at least be possible to enable it using a config option, if you show me the docs or an example on how to access options from odoo.conf i'll add in like 5 minutes

okawo80085:
Or wait, better yet, i'll add an option for it in the Oauth provider edit page

okawo80085:
As for the bug itself, i can show you the logs of what happens if i dont provide the `client_secret` field in the post requests

okawo80085:
Here are the logs btw
```
odoo2-1  | 2026-03-13 17:20:59,970 7 INFO fwdb2 werkzeug: 172.18.0.5 - - [13/Mar/2026 17:20:59] "GET /web/login?redirect=/odoo? HTTP/1.1" 200 - 6 0.003 0.008
odoo2-1  | 2026-03-13 17:21:04,383 7 ERROR fwdb2 odoo.addons.auth_oauth.controllers.main: Exception during request handling 
odoo2-1  | Traceback (most recent call last):
odoo2-1  |   File "/usr/lib/python3/dist-packages/odoo/addons/auth_oauth/controllers/main.py", line 139, in signin
odoo2-1  |     _, login, key = request.env['res.users'].with_user(SUPERUSER_ID).auth_oauth(provider, kw)
odoo2-1  |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
odoo2-1  |   File "/mnt/extra-addons/auth_oidc/models/res_users.py", line 56, in auth_oauth
odoo2-1  |     access_token, id_token = self._auth_oauth_get_tokens_auth_code_flow(
odoo2-1  |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
odoo2-1  |   File "/mnt/extra-addons/auth_oidc/models/res_users.py", line 43, in _auth_oauth_get_tokens_auth_code_flow
odoo2-1  |     response.raise_for_status()
odoo2-1  |   File "/usr/lib/python3/dist-packages/requests/models.py", line 1021, in raise_for_status
odoo2-1  |     raise HTTPError(http_error_msg, response=self)
odoo2-1  | requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://id.<mydomain>.com.ua/api/oidc/token
odoo2-1  | 2026-03-13 17:21:04,388 7 INFO fwdb2 werkzeug: 172.18.0.5 - - [13/Mar/2026 17:21:04] "GET /auth_oauth/signin?code=K0LjDBc1E63gW5E6GTld542E6ciCcQK0&state={"d":+"fwdb2",+"p":+4,+"r":+"http%253A%252F%252Fodoo2.s2.<mydomain>.com.ua%252Fodoo%253F"}&iss=https://id.<mydomain>.com.ua HTTP/1.1" 303 - 2 0.002 0.227
```

okawo80085:
If i add the `client_secret` field, the request doesn't fail anymore

okawo80085:
Now it's not enabled by default, but the user can still enable it if their OIDC provider requires it

okawo80085:
Forgot to disable by auto format and it changed more code then i needed to touch, so repushed

okawo80085:
So, @sbidoul , can this be merged now or are more changes needed?

sbidoul:
@okawo80085 looking at this again, could you try *removing both* client_id and client_secret from the POST data? I wonder if the bug is not that we send client_id alone. Normally the basic auth without client_id nor client_secret  in the POST data should be sufficient, and according to the spec, identity providers must support basic auth.

okawo80085:
Huh, will try your solution later today

PR #910, opened 2026-03-13
[19.0][MIG] password_security: Migration from 18.0 to 19.0

JustineBlanchet:
## Summary

Migration of the `password_security` module from version **18.0** to **19.0**.

## Changes

- Updated version: `18.0.1.0.0` → `19.0.1.0.0`
- No breaking API changes between Odoo 18 and 19 for this module
- All dependencies confirmed available in Odoo 19.0:
  - `auth_password_policy` ✓
  - `auth_password_policy_signup` ✓
  - `auth_totp` ✓
- `ensure_db` import path unchanged (`web.controllers.home`) ✓
- `auth_totp.controllers.home.Home` class name unchanged ✓
- Updated docstrings to follow OCA/Omydoo conventions (French for comments, English for technical terms)

## Test plan

- [ ] Install module on Odoo 19.0
- [ ] Verify password policy settings appear in General Settings
- [ ] Verify password complexity rules are enforced on login/signup
- [ ] Verify password expiration redirects to reset page
- [ ] Verify 2FA flow with expired password (`test_totp`)
- [ ] Verify password history prevents reuse
- [ ] Run full test suite: `test_res_users`, `test_login`, `test_signup`, `test_reset_password`, `test_change_password`, `test_password_history`, `test_totp`

## Migration notes

This is a straightforward port — no Odoo 19 API changes affect this module. The codebase is compatible as-is with minor version bump and comment updates per OCA conventions.

Migrated from: https://github.com/OCA/server-auth/tree/18.0/password_security

Made with [Cursor](https://cursor.com)


Reyes4711-S73 (approved):
LGTM
Functional review OK

sbiosca-s73 (approved):

hailangvn (changes requested):
Thanks for the work. Please use `git format-patch` as in [instruction](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0#technical-method-to-migrate-a-module-from-180-to-190-branch) to preserve commit history. Or you can use [oca-port](https://pypi.org/project/oca-port/) to do migration.

sergioM-S73:
I've been testing this branch in Odoo 19 and I encountered a traceback when a user tries to reset their password or sign up:
image

In Odoo 19, the auth_signup base controller added a new parameter do_login=True to the do_signup method. 
https://github.com/odoo/odoo/blob/894281f2a3e313fc239529572b5cc8c06a3511f7/addons/auth_signup/controllers/main.py#L165

kwoychesko:
Thanks @JustineBlanchet for taking this migration on, and @sergioM-S73 for testing it and catching the `do_signup` regression. We were migrating `password_security` to 19.0 in parallel on our own fork and ran into one more issue worth flagging here.

It's a second, subtler bug in the password-expiry flow, separate from the `do_signup(…, do_login=…)` one already reported above.

**Symptom:** when a user with an expired password logs in, `password_security` correctly bounces them to the reset-password page — but in a real browser that page shows **"Invalid signup token"**. It only reproduces from a real browser (one that sends a `tz` cookie) when the user already has a timezone set; header-less requests and the current test suite never hit it, which is why it stays green.

**Root cause:** in 19.0 `res.users._login` became an instance method running on the request env (in 18.0 it was a `@classmethod` with its own `cls.pool.cursor()`). Its timezone check —

```python
if tz in pytz.all_timezones and (not user.tz or not user.login_date):
```

now reads `login_date` into the **request env cache** *before* `_update_last_login()` writes the new one, and that stale value is never invalidated. The expiry bounce then calls `partner._get_signup_url()` → `_generate_signup_token()`, which signs the **stale** `login_date` into the token; validation later compares it to the fresh `login_date` and rejects it.

**Fix:** drop the stale cache before generating the token, in both `web_login` (`controllers/main.py`) and `web_totp` (`controllers/home.py`):

```python
request.session.logout(keep_db=True)
request.env.user.invalidate_recordset(["login_date", "log_ids"])
redirect = request.env.user.partner_id._get_signup_url()
```

**Reproducing it in a test** needs a `tz` cookie *and* a prior `res.users.log` SQL-backdated so the stale and fresh login dates differ by more than a second (the shared `HttpCase` cursor otherwise collapses the truncated-to-second timestamps); then assert the bounced `/web/reset_password` page doesn't contain "Invalid signup token".

Our fork's 19.0 branch has both this fix and the `do_signup` `*args/**kwargs` forward, plus a regression test for the expiry/`tz`-cookie case — please feel free to cherry-pick anything useful:
https://github.com/bpmi/server-auth/tree/19.0-mig-password_security

etobella (changes requested):
You did not follow the migration process...

moylop260:
FYI the migration process: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

moylop260:
Please, review the following PR merging all the PR and fixing the review:
 - https://github.com/OCA/server-auth/pull/980

Issue #902, opened 2026-02-24
Rights issues with impersonate_login


florian-dacosta:
When impersonating a user, we may have some right issues.
From my experiences, it is quite common when using wizards.

To reproduce one of those case, on a version 18 (but the bug is cross version), you can try to impersonate a user and then create an invoice from a sale order.
You-ll have this kind of error : 

<img width="1004" height="308" alt="Image" src="https://github.com/user-attachments/assets/1bbcc4ce-c82f-4efb-92b5-ecbf74bf1b3c" />

This is because of this kind of domain in `ir.rule`

<img width="1920" height="179" alt="Image" src="https://github.com/user-attachments/assets/f38287da-5083-40f9-901e-f3628f8a1a8c" />

The problem beeing that the module does set the create_uid on the impersonating user instead of the user beeing impersonated.

This is one example of the issue, but there are others.
In my opinion, it is a big issue, because the module's purpose is to debug and/or test what other users can do. And if it is not meant to do real actions on the production database, it is on preprod/dev environement.
When you have this kind of bug, it may be hard to know it it comes from the impersonate module or if it is a real bug with the user.

I am not sure on the best solution. I think it would be better to have a create/write_uid consistent to the current user to avoid such problems, so that we should stop forcing the impersonating user as create_uid.
We can already see on the chatter if something is created by an impersonating user.
I guess we could loose some traceability, for models without chatter or in case of write. But this impersonating feature should be reserved for managers/administrators so it is acceptable IMO.

Another idea could be to change the logic of create_uid, only for  wizard, which should solve some of the problems.
Or do something differently depending if we are on a production or developement environement, but it would add some constraint and complexity on the module.

Any idea ? @sebastienbeau @Kev-Roche @astirpe @tishmen @toita86

sebastienbeau:
I initially push Kevin for having the right uid in the create_uid/write_uid, in order to track who create/write the object.
After thinking twice and due to the issue raise, I think we should remove this hack.

Most of the time the write_uid is useless (as it's only the last user who did the write). And having the right create_uid is cool but is not so critical. What really matter is to have the right information in the mail.message (even if we can only have this information for model tracked).

So I am ok to remove it from the based module.

Note: If some people want to keep it, we can do it in an extra module.

PR #913, opened 2026-03-18
[18.0][FIX] impersonate_login access issues

Kev-Roche:
Related to issue : https://github.com/OCA/server-auth/issues/902
Add an inheritable function to keep real user on create / write on selected models. In this case, transient and abstract models. 

cc @florian-dacosta

### DIFF
--- a/impersonate_login/models/model.py
+++ b/impersonate_login/models/model.py
@@ -9,8 +9,17 @@ from odoo.http import request
 class BaseModel(models.AbstractModel):
     _inherit = "base"
 
+    def _keep_real_user_on_create_write(self):
+        # Avoid overriding the create_uid and write_uid
+        # when the model is abstract or transient
+        if self._abstract or self._transient:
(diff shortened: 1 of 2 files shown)


florian-dacosta (approved):
Code review LGTM
It should fixes most of the errors linked to the issue. if we find more error of this type on not transient models, we can decide if we need to do something more then.
At least the hook gives the possibility to manage it with real user in your own project if you feel the need.
Thanks

florian-dacosta:
@sebastienbeau can you review please ?

PaulGoubert (approved):

hbrunn:
for the record: we'll need something similar for attachments in the mailing wizard, because there attachments are created without res_id, and then the check is if [create_uid == env.uid](https://github.com/odoo/odoo/blob/18.0/odoo/addons/base/models/ir_attachment.py#L484), which fails for the same reason

PR #917, opened 2026-03-27
[IMP] auth_oidc: enforce JSON response for token endpoint (code flow)

kobros-tech:
Ensure the token endpoint always returns a JSON response when using
`_auth_oauth_get_tokens_auth_code_flow`.

Some OAuth2 providers (e.g., GitHub) return a URL-encoded or plain text
response by default if the "Accept: application/json" header is not set.
This leads to parsing errors when calling `response.json()`.

This commit enforces:
headers={"Accept": "application/json"}

when performing the token exchange request.

This method is shared by both OAuth2 Authorization Code flow and OpenID
Connect flows, so enforcing JSON ensures consistent and reliable behavior
across all providers.

Without this header, the response format may vary depending on the provider,
causing failures in access_token/id_token extraction.

This change improves compatibility with non-OIDC providers and prevents
unexpected decoding errors.


kobros-tech:
@sbidoul 

Hi, I appreciate your time, but I am sorry here is a new update to enable GitHub authentication.

sbidoul:
Sorry I don't understand the purpose of this PR. auth_oidc implements two oidc flows which are standardized, and section 3.1.3.3 of https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse says *In addition to the response parameters specified by OAuth 2.0, the following parameters MUST be included in the response: id_token*. 

So a token response without id_token is non compliant and in that case you probably need to create a module that implements another flow, assuming the standard Odoo auth_oauth module is not sufficient. But that is not oidc.

sbidoul (changes requested):
As explained

kobros-tech:
> Sorry I don't understand the purpose of this PR. auth_oidc implements two oidc flows which are standardized, and section 3.1.3.3 of https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse says _In addition to the response parameters specified by OAuth 2.0, the following parameters MUST be included in the response: id_token_.
> 
> So a token response without id_token is non compliant and in that case you probably need to create a module that implements another flow, assuming the standard Odoo auth_oauth module is not sufficient. But that is not oidc.

Thank you for the feedback. I completely agree that per Section 3.1.3.3 of the OpenID Connect Core 1.0 specification, an id_token is a mandatory requirement for a compliant OIDC Token Response.
My intention with this PR was not to diverge from the OIDC standard, but to provide a 'hybrid' capability within auth_oidc. This would allow the module to remain the primary authentication handler for environments that use strict OIDC (like Keycloak or Cognito) while gracefully falling back to the base Odoo auth_oauth logic for providers that only support OAuth 2.0 (like GitHub).
However, I understand the importance of keeping the OCA modules focused on their specific standards. Based on your suggestion, I will look into creating a separate module that complements auth_oidc rather than modifying it. My goal is to ensure both modules can work in harmony, allowing users to choose the appropriate flow for each provider without function overrides causing conflicts.
I appreciate the guidance and will pivot towards a modular approach.

pedrobaeza:
@kobros-tech why `auth_oauth` is not enough for doing what you want to do?

kobros-tech:
> @kobros-tech why `auth_oauth` is not enough for doing what you want to do?

Hi @pedrobaeza, that is the golden question. The short answer is: auth_oauth is hardcoded for the Implicit Flow, which many modern providers (like GitHub) have deprecated for security reasons.
Here are the three technical blockers in the standard auth_oauth that my PR/module addresses:

    Authorization Code Flow vs. Implicit Flow: Standard auth_oauth expects the access_token to be present in the URL fragment (Implicit Flow). Modern providers now require the Authorization Code Flow, where they return a code that must be exchanged for a token via a server-to-server POST request. Standard Odoo has no logic to perform this POST exchange.
    Missing Client Secret: Because auth_oauth was built for the Implicit Flow, the auth.oauth.provider model in Odoo doesn't even include a field for the Client Secret. Without a secret, Odoo cannot complete the handshake with providers like GitHub.
    Hardcoded Response Types: In the list_providers controller of the base module, Odoo hardcodes response_type=token. Most secure providers today (including GitHub) strictly require response_type=code [1].

The Purpose:
By bringing this into auth_oidc (or a dedicated code-flow module), we leverage the existing infrastructure for the Authorization Code Flow that is already present here. Without this, a user simply cannot log in with GitHub using the standard Odoo 18 auth_oauth module as it stands today.
I believe providing a secure 'Code Flow' fallback for non-OIDC providers is the missing piece for Odoo's authentication ecosystem.

kobros-tech:
@sbidoul
@pedrobaeza

I waited to share this until I had something fully valuable. This contribution is:

Fully clean and tested
Clear and reliable, thanks to the Mock classes that replicate real-world behavior
Verified with real authentication flows

I’m confident this improves the module’s stability and usability.

https://github.com/OCA/server-auth/pull/918

PR #922, opened 2026-04-01
[18.0][IMP] auth_oidc_environment: extra fields

qgroulard:
Needs https://github.com/OCA/server-env/pull/264

I believe more fields make sense to be configurable by environment.

Typically when you use different SSO for test and production.

### DIFF
--- a/auth_oidc_environment/__manifest__.py
+++ b/auth_oidc_environment/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "Auth Oidc Environment",
     "summary": """
         This module allows to use server env for OIDC configuration""",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
     "license": "AGPL-3",
     "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/server-auth",
--- a/auth_oidc_environment/migrations/18.0.1.1.0/post-migration.py
+++ b/auth_oidc_environment/migrations/18.0.1.1.0/post-migration.py
@@ -1,0 +1,11 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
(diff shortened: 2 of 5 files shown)


sbidoul (commented):
Does this need a migration or will it preserve existing values when this is deployed without setting the new env options?

qgroulard:
> Does this need a migration or will it preserve existing values when this is deployed without setting the new env options?

AFAIK no migration script is needed, this does nothing while no env var is set.

sbidoul:
I had the impression that server_environment removes the colums from the database and gets the data from a json field, but I don't know how it works when you add already-used fields to the server env mixin.

qgroulard:
I'll try to have a look but if this is the case we might have problems here as well: https://github.com/OCA/server-env/commit/3574e86d154df23078bc7f77b8817b44bc78cc7c

qgroulard:
Good catch, the previous values are indeed ignored.

I have opened this PR: https://github.com/OCA/server-env/pull/264

auth_oidc_environment/__manifest__.py
```diff
@@ -5,13 +5,13 @@
...
         This module allows to use server env for OIDC configuration""",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.1.0",
     "license": "AGPL-3",
     "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/server-auth",
     "depends": [
         "auth_oidc",
-        "server_environment",
+        "server_environment>18.0.1.0.6",
```
qgroulard, line 14: @sbidoul pre-commit doesn't like that. What would be the right syntax ?

sbidoul: Ah yes, Odoo would not support that either.

So the trick is to do that with a new section in pyproject.toml:

```toml
[tool.whool]
depends_override = { "server_environment": "odoo-addon-server_environment ==18.0.*, >18.0.1.0.6" }
```

PR #928, opened 2026-04-17
[18.0][ADD] auth_user_role

astirpe:


CasVissers-360ERP (approved):
Functional review

auth_user_role/models/res_users.py
```diff
@@ -0,0 +1,107 @@
...
+                self.env["ir.config_parameter"]
+                .sudo()
+                .get_param("auth_user_role.strict_sync", "True")
+                == "True"
+            )
+
+        roles_to_add = self._get_mapped_roles(identity_payload)
+
+        existing_lines = self.role_line_ids
+        existing_role_ids = set(existing_lines.mapped("role_id").ids)
```
CasVissers-360ERP, line 58: @astirpe In a full sync I would expect also res.groups to be removed since we want a full sync with roles setup for that user.

astirpe: @CasVissers-360ERP it's already the case, see this line:

```python
if strict_sync:
    self.set_groups_from_roles(force=True)
```

Since "force=True" is passed, Odoo performs the following:

- Calculates all groups implied by the user's current active roles.
- Wipes any groups currently on the user that are not in that calculated list.
- Ensures only the groups defined by the SAML-mapped roles remain.

This scenario is already covered by test_13_strict_sync_removes_native_groups().

CasVissers-360ERP: @astirpe what if the user has no roles?

astirpe: @CasVissers-360ERP In case a user ends up with no roles, here is how set_groups_from_roles() behaves:

The following statement is bypassed, because force is true:

```python
for user in self:
    if not user.role_line_ids and not force:
        continue
```

https://github.com/OCA/server-backend/blob/18.0/base_user_role/models/user.py#L92-L93

In this case:

groups_to_add will be an empty list;

groups_to_remove = list(set(user.groups_id.ids) - set([])). This means every single group the user currently has is flagged for removal;

Odoo executes the (3, ID) commands for all groups_to_remove, so that the user is stripped of all permissions. The user might be able to log in, but they will see a blank screen because they don't even have the basic base.group_user

CasVissers-360ERP: @astirpe My bad, thnx for the explanation!

PR #929, opened 2026-04-17
[18.0][ADD] auth_user_role_saml

astirpe:
Depends on https://github.com/OCA/server-auth/pull/928


CasVissers-360ERP (approved):
Functional review

astirpe:
@CasVissers-360ERP 
In commit https://github.com/OCA/server-auth/pull/929/changes/42596d11aa300c3d107fd6191d11023c1547d49e I added an extra safety check. If strict_sync is active, but the database has zero auth.user.role.mapping records configured, applying a strict sync would wipe all roles from every user who logs in via SAML. The last commit avoids it.
Do you agree?

CasVissers-360ERP:
@astirpe I agree.

PR #930, opened 2026-04-22
change to base.group_system for view permissions

NETechAlly:
Seems like a logical change to make the OAUTH Tokens information visible to administrative users in the base.group.system as opposed to invisible="id != uid".

Right now a full system admin cannot set or set the token limit or clear them. This change allows system admins to manage tokens limits and tokens for users.


NETechAlly:
Pretty basic change. It seems like maximum tokens and the ability to clear tokens is more of an administrative task than a task for a user to do on their own account. Thanks for your work on this!

PR #934, opened 2026-04-23
[18.0][FIX] vault: Error when reencrypt is pressed

CarlosRoca13:
cc @Tecnativa TT62141

ping @victoralmau @pedrobaeza

### DIFF
--- a/vault/static/src/backend/controller.esm.js
+++ b/vault/static/src/backend/controller.esm.js
@@ -165,13 +165,9 @@ patch(FormController.prototype, {
                     rec.iv
                 );
                 if (val === null) {
+                    const fixed_text = _t("of entry");
                     problems.push(
-                        _.str.sprintf(
-                            _t("%s '%s' of entry '%s'"),
-                            type,
-                            rec.name,
-                            rec.entry_name
-                        )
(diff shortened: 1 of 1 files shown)


vault/static/src/backend/controller.esm.js
```diff
@@ -159,25 +159,20 @@ patch(FormController.prototype, {
...
                 );
                 if (val === null) {
                     problems.push(
-                        _.str.sprintf(
-                            _t("%s '%s' of entry '%s'"),
-                            type,
-                            rec.name,
-                            rec.entry_name
-                        )
+                        _t(`${type} '${rec.name}' of entry '${rec.entry_name}'`)
```
pedrobaeza: I don't think you should translate the whole string, as it won't match anything in the translations. Please translate only the fixed texts.

pedrobaeza (changes requested):

vault/static/src/backend/controller.esm.js
CarlosRoca13: Done

pedrobaeza (approved):

victoralmau (approved):

PR #935, opened 2026-04-28
[19.0][MIG] auth_admin_passkey

andrel-exo:


legalsylvain (changes requested):
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/#migration

You can also take a look on the project https://github.com/OCA/odoo-module-migrator/ to make easier migration

PR #936, opened 2026-04-28
[19.0][MIG] auth_admin_passkey

andrel-exo:


auth_admin_passkey/models/res_users.py
```diff
@@ -61,7 +60,7 @@ def _check_credentials(self, credential, env):
 
         except exceptions.AccessDenied:
             # Just be sure that parent methods aren't wrong
-            users = self.with_user(SUPERUSER_ID).search([("id", "=", self._uid)])
+            users = self.with_user(SUPERUSER_ID).browse(self.env.uid).exists()
```
legalsylvain: why  `.exists()` ?

legalsylvain (commented):
a minor question. 
thanks for porting this module !

auth_admin_passkey/models/res_users.py
andrel-exo: I added it to keep the existence check explicit after replacing `search()` with `browse()`, since `browse()` alone may return a non-existing record. But it changes the previous behavior compared to `search()`, so I’ll remove it and only replace `self._uid` with `self.env.uid`.

legalsylvain (approved):

atchuthan (approved):
Functional & Tested in local 👍

skanndar (changes requested):
Thanks for the migration — the port itself is faithful to 18.0. But I tested it on a clean 19.0 database with the `odoo.conf` documented in `readme/CONFIGURE.md` and none of the boolean keys work any more, including one that fails **open**. This is not something the migration introduced: 19.0 changed how the config file is read.

**Root cause.** Up to 18.0, `configmanager._parse_config()` cast `'True'`/`'False'` to booleans for *every* key found in the `[options]` section. Since 19.0, `_load_file_options()` only parses the options it knows about and stores the rest as-is:

```python
option = self.options_index.get(name)
if not option:
    ...
    "unknown option %r in the config file at %s, option stored as-is, without parsing"
    self._file_options[name] = value
    continue
```

All the `auth_admin_passkey_*` keys are unknown options, so they now arrive as **strings** — and `"False"` is truthy. Odoo even logs a warning for each of them at startup.

**Setup**: official `odoo:19` image (build `19.0.20260723`), fresh database, this PR's branch, and an `odoo.conf` with the documented keys:

```ini
auth_admin_passkey_password = masterkey
auth_admin_passkey_send_to_user = False
auth_admin_passkey_ignore_totp = False
auth_admin_passkey_password_sha512_encrypted = False
```

Every one of them is read as `'False'` (`str`, truthy). Three observable consequences:

**1. The passkey itself stops working.** `password_sha512_encrypted` is truthy, so the typed password is sha512-hashed and compared against the plain `masterkey` from the config: `_check_credentials()` raises `AccessDenied`. With the documented configuration, the module's main feature is dead.

**2. `send_to_user = False` is ignored.** `_send_email_passkey()` still creates the `mail.mail` for the user (verified on the real method, one message created for the impersonated user).

**3. `ignore_totp = False` *enables* the 2FA bypass.** For a user with TOTP enabled:

| `request.session["ignore_totp"]` | `_mfa_url()` |
|---|---|
| not set | `/web/login/totp` |
| `False` (real boolean) | `/web/login/totp` |
| `'False'` (what 19.0 gives) | `None` → **2FA skipped** |

So an administrator who writes `auth_admin_passkey_ignore_totp = False` to make sure the bypass stays off gets exactly the opposite. That is the one I would not let through: the other two fail closed or are merely annoying, this one fails open.

I am reporting it here rather than through a private channel because the module is not in the `19.0` branch yet, so no one is running it on 19.0 — but it is worth fixing before the merge for that reason.

**Suggested fix** — normalise the falsy strings where the config is read, e.g.:

```python
_FALSY_CONFIG_VALUES = ("", "false", "none", "0")


def _get_config(key, default=False):
    value = config.get(key, default)
    if isinstance(value, str) and value.strip().lower() in _FALSY_CONFIG_VALUES:
        return False
    return value
```

and use it for the four keys. I have this running on two 19.0 databases (passkey login works, no notification when `send_to_user = False`, 2FA respected when `ignore_totp = False`, and the user's own password keeps working). Happy to push it here as a commit, with a test covering the three cases, if you prefer.

Worth keeping in mind for other migrations too: any module reading its own boolean keys from `odoo.conf` with `config.get()` has the same problem on 19.0.

SiscoIngeos:
LGTM

andrel-exo:
Thanks for the detailed report, the root cause is correct and I could reproduce it on 19.0. Fixed in c5c3325.

I used Odoo's own `str2bool` instead of a custom list of falsy strings: it exists in both 18.0 and 19.0, accepts booleans and strings, and takes a default for unparsable values. Applied to the three boolean keys only (`send_to_user`, `password_sha512_encrypted`, `ignore_totp`), `auth_admin_passkey_password` stays a raw string.

Added four tests that assign the raw strings the module gets from the config file, instead of Python booleans as the existing tests do. They fail without the fix and pass with it, verified on a real 19.0 instance with the keys set in `odoo.conf`.

@legalsylvain @atchuthan sorry for the noise, this needs a new look since you already approved.

skanndar (approved):
Re-tested `c5c3325` on a clean 19.0 database with the same setup as my report — official `odoo:19` image, real `odoo.conf` (not values injected into `config` at runtime, since the file-parsing path is what changed in 19.0), `auth_totp` installed. All three issues are gone:

| With `= False` in `odoo.conf` | before | now |
|---|---|---|
| `password_sha512_encrypted` → passkey login | `AccessDenied` | works |
| `send_to_user` → email to the impersonated user | sent | not sent (and the sysadmin one still goes out) |
| `ignore_totp` → `_mfa_url()` for a TOTP user | `None`, 2FA skipped | `/web/login/totp`, 2FA enforced |

For the third one I kept the controls from the report: no session flag and a real `False` boolean both give `/web/login/totp`, and now the value coming from the config file does too.

`str2bool` is the better call — thanks for not taking my falsy-list as-is. I also checked the edges: an unparsable value falls back to the given default (`ignore_totp` → `False`, so 2FA stays on; `send_to_user` → `True`, so the notification stays on — both fail the safe way), and an absent key still works since `str2bool` accepts the boolean default itself. Applying it only to the three boolean keys and leaving `sysadmin_email` / `sysadmin_lang` / `password` alone is right.

Tests 12-15 cover exactly the three cases plus the positive one; the module suite is green here (20 tests).

LGTM 👍

PR #941, opened 2026-05-18
[17.0][ADD] auth_oauth_link_by_email: add new module to allow linking OAuth accounts by email

zamberjo:
When existing Odoo users (created before OAuth/OIDC was configured) try to log in via an OAuth provider, the system fails to find them because their `oauth_uid` is not set, falling through to signup and failing.

This module adds an auto-link mechanism: on first OAuth login, if no user is found by `oauth_uid`, it attempts to match an existing active user whose `login` equals the **email** claim from the token, linking them automatically by writing the `oauth_uid` and `oauth_provider_id`.

Subsequent logins use the standard `oauth_uid` lookup with no extra queries.

**Usage**: install `auth_oauth_link_by_email` — no configuration needed.


sbidoul:
Hi! auth_oidc has a mechanism to control token fields mapping (see `token_map` field). Is that not sufficient for your use case?

zamberjo:
Hi! The issue is that existing users have an empty (`False`) **oauth_uid**, so [`_auth_oauth_signin` ](https://github.com/odoo/odoo/blob/17.0/addons/auth_oauth/models/res_users.py#L101)cannot find them during the lookup:

```python
self.search([("oauth_uid", "=", oauth_uid), ("oauth_provider_id", "=", provider)])
```

Even if **token_map** is configured with something like `email:user_id`, existing users still won't have that mapped value stored in **oauth_uid**

sbidoul:
Ah, I see. Then you probably need to create a new module that depends on `auth_oauth` and/or `auth_signup` as this feature is not releated to OpenID Connect but more to the generic oauth signup feature..

zamberjo:
> Ah, I see. Then you probably need to create a new module that depends on `auth_oauth` and/or `auth_signup` as this feature is not releated to OpenID Connect but more to the generic oauth signup feature..

Done.

sersanchus (approved):

maurochip (approved):

PR #945, opened 2026-06-04
[FIX] auth_api_key: Read API key from headers

SirPyTech:
Without this change, the API key is not being read from the headers.

~~It seems very strange to me that this is not working, but in my case it isn't, please let me know if I'm missing something.~~
In the end, the problem was elsewhere, but I'll leave this here because it smells to have `headers = request.httprequest.environ`: the `request`'s `headers` should be assigned to the `headers` variable.


francesco-ooops:
@simahawk wdyt?

simahawk (approved):
make sense

ivantodorovich (approved):

PR #949, opened 2026-06-11
[19.0][IMP] auth_api_key: generate API keys automatically

yankinmax:
Add server side random API key generation to `auth_api_key`, with a form button for empty keys and a revealable password widget for easier key management.


auth_api_key/views/auth_api_key.xml
```diff
@@ -21,7 +21,18 @@
...
                     <group name="config" colspan="4" col="4">
                         <field name="user_id" colspan="4" />
-                        <field name="key" colspan="4" password="True" />
+                        <label for="key" />
+                        <div class="o_row" colspan="3">
+                            <!-- The password widget masks the key and provides a reveal button. -->
+                            <field name="key" widget="password" />
+                            <button
+                                name="generate_random_key"
+                                string="Generate"
```
Ricardoalso:
```suggestion
                                string="Generate Random Token"
```

Ricardoalso (commented):
Nice feature, Thanks :pray: 

I’m wondering whether removing `required=True `will break any behavior built on top of this. It’s worth documenting in the commit and PR description

Ricardoalso:
> Nice feature, Thanks 🙏
> 
> I’m wondering whether removing `required=True `will break any behavior built on top of this. It’s worth documenting in the commit and PR description

Oh, ok I’ve just seen https://github.com/OCA/server-auth/pull/950/. The removal of `required=True ` relates to the other PR and could be moved there to avoid any confusion.

auth_api_key/models/auth_api_key.py
```diff
@@ -12,7 +14,6 @@ class AuthApiKey(models.Model):
 
     name = fields.Char(required=True)
     key = fields.Char(
-        required=True,
```
ivantodorovich: Please make sure this is ok security-wise

Like, we don't want to suddently allow authentications from people sending an empty string for API-KEY in the http header

yankinmax: This will be guarded with an ORM constraint

ivantodorovich: 🤔 maybe I wasn't clear.
I meant this: https://github.com/OCA/server-auth/pull/950#issuecomment-4691402872

yankinmax: I think this implementation is better than removing `required` attribute.
Or I still don't understand what changes you propose for this PR.
Thank you for the review.

ivantodorovich (approved):

Ricardoalso (approved):

ivantodorovich:
Just rebased to solve conflicts after merging https://github.com/OCA/server-auth/pull/950

@etobella is it ok to you? 🙏🏻

etobella (approved):

simahawk (approved):

PR #950, opened 2026-06-11
[19.0][IMP] auth_api_key: neutralize API keys

yankinmax:
Clear stored `auth_api_key` secrets during database neutralization.

### DIFF
--- a/auth_api_key/README.rst
+++ b/auth_api_key/README.rst
@@ -70,6 +70,8 @@ The api key menu is available into Settings > Technical in debug mode.
 By default, when you create an API key, the key is saved into the
 database.
 
+When a database is neutralized, stored API key values are cleared.
+
 If you want to manage them via serve environment settings use
 auth_api_key_server_env.
 
--- a/auth_api_key/data/neutralize.sql
+++ b/auth_api_key/data/neutralize.sql
@@ -1,0 +1,3 @@
+-- remove API keys
+UPDATE auth_api_key
+   SET key = NULL;
--- a/auth_api_key/models/auth_api_key.py
+++ b/auth_api_key/models/auth_api_key.py
@@ -12,7 +12,6 @@ class AuthApiKey(models.Model):
 
     name = fields.Char(required=True)
     key = fields.Char(
-        required=True,
         help="""The API key. Enter a dummy value in this field if it is
         obtained from the server environment configuration.""",
     )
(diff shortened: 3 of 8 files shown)


test-requirements.txt
```diff
@@ -1 +1,2 @@
 responses
+odoo-addon-auth_api_key @ git+https://github.com/OCA/server-auth.git@refs/pull/949/head#subdirectory=auth_api_key
```
ivantodorovich: Why is it a dependency?

ivantodorovich (commented):
LG overall

test-requirements.txt
yankinmax: Because in the https://github.com/OCA/server-auth/pull/949/changes#r3396650656 we're removing the required attribute to set keys as NULL: neutralize database.
Without a dependency we'll fail the tests here.
We're security protected not having empty keys, because we've an ORM constraint while still able to set NULL on the database level.

TDu:
It is already a feature in 18 https://github.com/OCA/server-auth/blob/18.0/auth_api_key/data/neutralize.sql should we not forward port ?

yankinmax:
> It is already a feature in 18 https://github.com/OCA/server-auth/blob/18.0/auth_api_key/data/neutralize.sql should we not forward port ?

It's not exactly what we think about. @ivantodorovich what do you think?

vvrossem:
> It is already a feature in 18 https://github.com/OCA/server-auth/blob/18.0/auth_api_key/data/neutralize.sql should we not forward port ?

:thinking: this would benefit #949: `key` would remain `required`

test-requirements.txt
Ricardoalso: That's why I was asking if you don't want to move it into this PR https://github.com/OCA/server-auth/pull/949#issuecomment-4681352646

ivantodorovich:
For me setting to `"neutralized"` is a security risk.

It means if you have services protected by this key, now people can easily authenticate to your non-production instances by using this "neutralized" API-KEY, potentially leaking private data.

Unless authentication is somehow prevented if the key is exactly `"neuralized"`, but I don't think that's implemented.


For the record, I see also the same posible risk with this proposal -- if we set the key to `NULL`, will external people be allowed to authenticate if they provide an empty `API-KEY` header?

IMO, if we make the key nullable, then we must change this line: https://github.com/OCA/server-auth/blob/2769e9f7776e9ae2d2197297089addf9171d4d48/auth_api_key/models/auth_api_key.py#L43

```diff
-        for api_key in self.search([], limit=None):
+        for api_key in self.search([("key", "!=", False)], limit=None):
```

test-requirements.txt
yankinmax: No other dependency. The `required` removal and constraint is here.

ivantodorovich (approved):
Thanks!

Ricardoalso (approved):

ivantodorovich:
@lmignon @etobella can we get your opinion on this? 🙏🏻

etobella (approved):

lmignon (approved):

PR #956, opened 2026-06-18
[19.0][FIX] auth_saml: fix no users

vincent-hatakeyama:
This is a forward port of #848. The fix is already in the branch, this PR avoid logging if there is no user and adds the tests.


vincent-hatakeyama:
Included in the migration merge request.

PR #966, opened 2026-07-01
[18.0][ADD] base_group_subcontractor: Extra user type group for restricted access to backend

pedrobaeza:
This module adds a new isolated user type group called "Subcontractors", that simulates to be an internal user, giving access to the backend, but only granting the minimum permissions for the web client to work, and only exposing the Discuss menu.

The rest of the needed permissions for that subcontractors should be given in extra customizations.

Use Cases / Context
===================

A usual need inside a company using Odoo is to have external people (subcontractors) that need to access Odoo to do certain operations that are not allowed (or they are not comfortable) in the portal, but giving them an internal system user is too much, as there some visible information (like contacts or calendar events) that shouldn't be shown by these users, even by legal reasons (GDPR, etc).

Known issues / Roadmap
======================

- Any extra installed module that performs operations on the web client (example: base_tier_validation), will require glue modules that adds the permissions for the corresponding models.

@Tecnativa TT63310


christian-ramos-tecnativa (approved):
Functional and code review OK!

victoralmau (approved):
Functional review OK

PR #972, opened 2026-07-16
[18.0][FIX] password_security: sudo password history read for reuse

LauraCForgeFlow:
The `res.users.pass.history` record rules only let a user read their own history. So when someone set another user's password (an admin resetting it), `_check_password_history` read an empty history and the "cannot reuse recent passwords" check was silently skipped. Reading the history with `sudo()` fixes this, the check sees the target's real history regardless of who triggers the change.


LoisRForgeFlow (approved):
Makes sense

JoanSForgeFlow (approved):

PR #973, opened 2026-07-22
[18.0][FIX] vault: generate keys on first usage instead of failing

jans23:
_ensure_keys() threw 'Failed to import keys from database' for a first-time user with no key pair yet (e.g. opening the vault module and pressing New then Save). It now generates a new key pair when none exist in the database, matching the fallback already present in _initialize_keys(). A genuine import failure with existing keys still raises the error. Added a QUnit test covering the first-usage flow.

### DIFF
--- a/vault/static/src/backend/vault.esm.js
+++ b/vault/static/src/backend/vault.esm.js
@@ -119,9 +119,12 @@ const vaultService = {
                 const store = await this._get_object_store();
                 store.clear();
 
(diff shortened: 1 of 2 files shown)


CarlosRoca13 (approved):
It seems to be ok! Thanks

pedrobaeza (approved):
/ocabot merge patch

PR #974, opened 2026-07-22
[17.0][FIX] users_ldap_populate: do not deactivate portal users

cvinh:
Before this commit, the portal users (who are not in LDAP directory) are archived


cvinh:
tests fail with auth_jwt not with this module

PR #975, opened 2026-07-22
[18.0][IMP] vault: add User security group to control module access

jans23:
Introduce a new visible User security group under a dedicated "Vault" category. The whole module (menus, model access and record rules) is now gated behind this group instead of base.group_user. The group is implied by base.group_user so every internal user keeps access by default; removing the group from a user denies access to the vault without touching their keys.


cvinh (approved):
Yes of course

cvinh:
> The group is implied by base.group_user so every internal user keeps access by default;

This is not a good idea... but I think you force pushed to default user

cvinh:
Maybe you can update readme to describe the feature

jans23:
@cvinh I addressed your feedback. Thank you for that.

jans23:
@CarlosRoca13 could you review this, please?

fkantelberg (changes requested):
This might be a breaking change for existing databases. If you include the fix in an already created database the users which had vaults before can't access them because the group isn't set on the migration. I would advise some migration code to set the newly created group to all existing internal users.

This isn't a security improvement because the rights control the access it only blocks the features from users. I don't fully understand the concept. You add it to the defaults but want to explicitly remove the rights from users. Afterwards when they already created their private vaults? I have some concerns about this workflow because the person who removes the access doesn't see if the user has a vault which he uses exclusivly.

jans23:
@fkantelberg Migration code added. This PR adds a way to manage which users can use Vault. Most other apps in Odoo have such option too. Same as not all users have access to--say--the Sales module but the administrator can control this. Of course if an access is removed later on the administrator should be aware of the consequences which is the case for all other modules too.

fkantelberg:
@jans23 The comparison with sales is off because the access model isn't relying on the groups and in sales the groups are the only possible access. If you want to compare why not with Calendar? Every one gets access but you only see those private events you are assigned to. Or Discuss. A basic feature with channels (vaults) to discuss (share secrets).

The reasoning with the admin isn't convincing because in all other cases the admin (or the person who manages access) can see which documents a person has access to. For vault and especially private vault those aren't shown anywhere inside of the UI except for the specific user. How do you want to address it?

If I'm honest I would leave the vault open to all internal users because it gives the all internal users an option to store secrets securely and otherwise people tend to use more unsecure methods to achieve it. I don't see how this is an improvement it looks more like a restriction without benefits.

PR #978, opened 2026-07-29
[MIG] password_security: Migration to 19.0

SiscoIngeos:

PR #855, opened 2025-10-27
[19.0][MIG] password_security: Migration to 19.0

bizzappdev:

PR #980, opened 2026-08-14
[MIG] password_security: Migration to 19.0 

moylop260:
* res.users.groups_id renamed to group_ids

Odoo 19.0 refactored the groups relation on res.users: groups_id is gone,
replaced by group_ids (explicitly assigned groups) and all_group_ids
(including implied ones).

Without this, test_user_with_admin_rights_can_create_users fails with
AttributeError: 'res.users' object has no attribute 'groups_id'

* Opener takes the HttpCase instead of a cursor

In Odoo 19.0 odoo.tests.common.Opener.__init__ receives the HttpCase
instance (it reads http_case.cr and asserts http_case.opener is itself),
so passing self.env.cr no longer works.

Without this, the 15 post_install HttpCase tests fail with
AttributeError: 'psycopg2.extensions.cursor' object has no attribute 'cr'

* forward the extra do_signup arguments

In 19.0 auth_signup's do_signup() gained a do_login parameter, and
web_auth_reset_password() now calls it as do_signup(qcontext, do_login=False):
https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/addons/auth_signup/controllers/main.py#L97
https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/addons/auth_signup/controllers/main.py#L165

The override kept the old two-argument signature, so resetting a password or
signing up through a token failed with:

    PasswordSecurityHome.do_signup() got an unexpected keyword argument 'do_login'

Forwarding *args/**kwargs keeps the override working across signature changes.

Reported by @sergioM-S73 in
https://github.com/OCA/server-auth/pull/910#issuecomment-4439151259

The existing tests only exercised the token-less "send instructions" branch of
/web/reset_password, which never reaches that call. The added regression test
posts a real signup token, which is the path that broke.

* invalidate the stale login_date before the reset bounce

When a user whose password has expired logs in, the module kicks them out to
the reset password page, but in a real browser that page shows "Invalid signup
token".

In 19.0 res.users._login() became an instance method running on the request env
(in 18.0 it was a classmethod with its own cursor):
https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/odoo/addons/base/models/res_users.py#L760

When the browser sends a tz cookie and the user already has a timezone,
"not user.tz or not user.login_date" reaches login_date and loads it into the
env cache, before _update_last_login() creates the new res.users.log:
https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/odoo/addons/base/models/res_users.py#L771-L775

login_date is a related field on log_ids.create_date and _update_last_login()
only creates a log record, so nothing refreshes that cached value. The expiry
bounce then calls _get_signup_url() -> _generate_signup_token(), which signs
login_date into the payload, so the token carries the stale date and is later
rejected:
https://github.com/odoo/odoo/blob/13e9e827dc21103052e8607d16f23c07adeb5634/addons/auth_signup/controllers/main.py#L149

Reported by @kwoychesko in
https://github.com/OCA/server-auth/pull/910#issuecomment-4728065429

The regression test needs both a tz cookie and a backdated res.users.log:
create_date is truncated to the second, so on the shared test cursor the stale
and the fresh login_date would otherwise be identical. Header-less requests
never reach that branch, which is why the rest of the suite stays green.

Related to
- [x] https://github.com/OCA/server-auth/pull/855
- [x] https://github.com/OCA/server-auth/pull/910
- [x] https://github.com/OCA/server-auth/pull/978


moylop260:
@Chandresh-SerpentCS @dreispt @astirpe 

WDYT?

Reyes4711-S73 (approved):
LGTM

nhomar:
+1

PR #983, opened 2026-08-21
[18.0][ADD] vault_portal: portal access to shared vaults

cvinh:
Extends vault (OCA/server-auth) to let portal contacts view and, where explicitly granted per contact, edit or add password field values from their own portal account.

Access stays strictly per individual, end-to-end encrypted exactly like internal users: each contact holds their own key pair, and the vault's master key is re-wrapped for them individually. Two clients never share any key material.

Read-only by default at every layer (model access, ir.rule and vault.abstract.check_access(), unchanged from "vault"): the vault itself and entry names always stay read-only from the portal; only the value of a vault.field can be written, gated by
vault.right.perm_write/perm_create per contact.

An optional MFA policy (Settings > Vault) can require two-factor authentication before a portal contact is granted write access, or any access at all; disabling 2FA afterwards applies the same policy in reverse.

Assisted-by: Claude Sonnet 5


cvinh:
@jans23 it would supercool if you review this 99% AI coded module... we deployed it in test and it looks like ok...
About reviewing js I'm not expert...

jans23:
My colleague would be better positioned for reviewing. @KDRdev could you review it please?

cvinh:
still WIP

cvinh:
ready for review
@jans23 maybe you can do a functional review on runboat ?

vault_portal/static/src/frontend/portal_vault.esm.js
```diff
@@ -0,0 +1,579 @@
...
+ * the current portal contact. Plain vanilla JS, mirroring vault's own
+ * static/src/frontend/inbox.esm.js. The private key is kept in a
+ * page-local variable only (not cached in IndexedDB): the master
+ * password is asked again on every visit.
+ */
+
+import {rpc} from "@web/core/network/rpc";
+import utils from "@vault/common/utils.esm";
+
+const root = document.getElementById("o_portal_vault_root");
```
fkantelberg: It should use the proper JS framework of Odoo instead of plain JS.

vault_portal/static/src/frontend/portal_vault.esm.js
```diff
@@ -0,0 +1,579 @@
...
+    }
+
+    /**
+     * Generate a new key pair, wrap the private key with a key derived
+     * from the chosen master password, and store it server-side.
+     * Mirrors vault's own backend/vault.esm.js _export_to_database(),
+     * so the stored res.users.key record has the exact same shape and
+     * remains readable by the standard "vault" backend logic too.
+     */
+    async function createKeys(password) {
```
fkantelberg: Having this in 2 places can lead to problems in the future (e.g. if only one appearence is adjusted). Is the vault service not working?

vault_portal/static/src/frontend/portal_vault.esm.js
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        const iv = utils.generate_iv_base64();
+        const encrypted = await utils.sym_encrypt(masterKey, value, iv);
+
+        // /vault/replace only writes existing records; creation goes
+        // through the standard generic ORM call instead.
+        let newId = null;
+        try {
+            const result = await rpc("/web/dataset/call_kw", {
```
fkantelberg: It should go through controllers instead of this.

fkantelberg (changes requested):
I left only some comments. The same applies for most of the JS. Please use the framework and controllers instead of doing everything with plain JS when it's not needed. Also a lot of the important functions like encryption/decryption etc. is duplicated and can lead to problems in the future. I haven't tested it yet and from glancing over it the key management might be a problem or atleast not fully supported (invalidation, new password etc.)

cvinh:
> I left only some comments. The same applies for most of the JS. Please use the framework and controllers instead of doing everything with plain JS when it's not needed. Also a lot of the important functions like encryption/decryption etc. is duplicated and can lead to problems in the future. I haven't tested it yet and from glancing over it the key management might be a problem or atleast not fully supported (invalidation, new password etc.)

Thanks for reviewing this 99% AI dev... I was unable to review js/controller because of my very bad level in it...
I hope there is not too many garbage now...

cvinh:
tests fail because of cross_connect_server/tests/test_cross_connect_server.py

vault_portal/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "depends": ["vault", "portal", "auth_totp"],
+    "data": [
+        "security/ir.model.access.csv",
+        "security/ir_rule.xml",
+        "views/portal_templates.xml",
+        "views/res_config_settings_views.xml",
+    ],
+    "assets": {
+        "web.assets_frontend": [
+            "vault/static/src/backend/vault.esm.js",
```
fkantelberg, line 23: `vault_utils` are missing.

vault_portal/controllers/portal.py
```diff
@@ -0,0 +1,169 @@
...
+    @route(
+        ["/my/vaults/<int:vault_id>/fields/<int:field_id>"],
+        type="json",
+        auth="user",
+    )
+    def portal_vault_write_field(self, vault_id, field_id, value, iv, **kwargs):
+        vault = self._get_vault_or_404(vault_id)
+        field = self._get_vault_record(vault, "vault.field", field_id, _("Field"))
+        field.write({"value": value, "iv": iv})
+        return self._field_props(field)
```
fkantelberg, line 180: Do you also want to support files? I don't see it anywhere and not even listed in the roadmap.

fkantelberg (commented):
Tell me if it's working again and you need a proper review + tests. It goes into a better direction now.

vault_portal/controllers/portal.py
cvinh: Done

vault_portal/__manifest__.py
cvinh: Done

PR #984, opened 2026-08-22
[18.0][FIX] vault: mark inbox route as website to fix KeyError: 'website'

jans23:
The /vault/inbox/<token> route renders the vault.inbox template which calls web.login_layout. When the website module is installed it overrides web.login_layout to render website.layout, which requires the 'website' value in the render context. That value is only injected when the request is flagged as a frontend request (request.is_frontend), which in turn is derived from the route's website=True flag. Without it, rendering the shared-secret page crashed with KeyError: 'website'.

### DIFF
--- a/vault/controllers/main.py
+++ b/vault/controllers/main.py
@@ -11,7 +11,7 @@ _logger = logging.getLogger(__name__)
 
 
 class Controller(http.Controller):
-    @http.route("/vault/inbox/<string:token>", type="http", auth="public")
+    @http.route("/vault/inbox/<string:token>", type="http", auth="public", website=True)
     def vault_inbox(self, token):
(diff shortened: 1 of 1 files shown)


vault/tests/test_controller.py
```diff
@@ -50,6 +50,12 @@ def setUpClass(cls):
             }
         )
 
+    def test_vault_inbox_is_website_route(self):
```
pedrobaeza: Please don't add tests that don't check flows although the AI suggest them.

jans23: I removed the test as requested.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #985, opened 2026-08-22
[18.0][FIX] vault: avoid singleton error on batched vault.right create

jans23:
A batched create() results in several records and the guard accessed the non-stored related field allowed_share on the whole set, triggering ensure_one() -> 'Expected singleton: vault.right(...)'.


jans23:
@CarlosRoca13 could you review this fix, please?

CarlosRoca13 (approved):

jans23:
@cvinh could you review, please?

pedrobaeza (approved):
/ocabot merge patch

PR #986, opened 2026-08-22
[19.0][FIX] password_security: avoid blocking module installation

### LINKED ISSUE #865: Default password_security value sbreak all test + some apps installs in demo mode.
## Module
password_security

## Describe the bug
There is currently a post_install script that sets all the defaults. 
With the defaults at something else then 0, It breaks the install of all other apps in demo mode that require to create a user.
More than that, it also breaks all tests executed on the databases since none of the passwords sets in the tests respect the password policies.

## To Reproduce
**Affected versions**: 17

1. Just install the app
2. Install a standard app that creates an account (in my case test_l10n_be_hr_payroll_account)
3. The installation cannot proceed because of the password verifications...

**Expected behavior**
You should really simply leave the defaults at 0, and let people configure that app on their production database !

**Additional context**
Everything is said.

luisg123v:
Users created from data files can have passwords that do not comply with the configured policy, which causes modules that define such users to fail during installation, even demo data.

To fix this, the policy check is skipped while loading data files.

Note that Tests are unaffected because users created there are through the ORM, so the policy is still covered.

Closes #865


luisg123v:
Hi @moylop260,

Could you review, please?

CC @isaako34, @desdelinux

moylop260 (approved):
👍

moylop260 (approved):

moylop260:
Approve since that I have a similar issue using this same way xd
 - https://github.com/Vauxoo/partner-contact/pull/14

luisg123v:
@pedrobaeza what do you think?

pedrobaeza:
The PR has conflicts.

luisg123v:
Done  @pedrobaeza, conflicts solved.

pedrobaeza (approved):
/ocabot merge patch

PR #987, opened 2026-08-24
[18.0][FIX] vault: inbox share form token handling and optional file

jans23:
The frontend inbox share form showed 'Invalid token' for valid links and wrongly required a file even when a name and secret were provided.

### DIFF
--- a/vault/controllers/main.py
+++ b/vault/controllers/main.py
@@ -19,12 +19,16 @@ class Controller(http.Controller):
         user = request.env["res.users"].sudo().find_user_of_inbox(token)
         if len(inbox) == 1 and inbox.accesses > 0:
             ctx.update({"name": inbox.name, "public": inbox.user_id.active_key.public})
-        elif len(inbox) == 0 and len(user) == 1:
+        elif len(user) == 1:
             ctx["public"] = user.active_key.public
-
-        # A valid token would mean we found a public key
-        if not ctx.get("public"):
+        else:
+            # The token doesn't match any inbox or user
             ctx["error"] = _("Invalid token")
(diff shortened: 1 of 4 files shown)


vault/views/templates.xml
```diff
@@ -64,7 +64,6 @@
                         placeholder="Secret"
                         id="secret_file"
                         name="secret_file"
-                        required="required"
```
fkantelberg, line 67: True this shouldn't be required here but neither should secret be required. It should be made sure that either one of those must be present. Without the required here it's not visually obvious what is missing except for the error message on submit.

- Entering secret but not secret_file => allowed
- Entering secret_file but not secret => allowed
- Entering secret and secret_file => allowed
- Entering neither secret nor secret_file => denied

vault/controllers/main.py
```diff
@@ -17,14 +17,18 @@ def vault_inbox(self, token):
         # Find the right token
         inbox = request.env["vault.inbox"].sudo().find_inbox(token)
         user = request.env["res.users"].sudo().find_user_of_inbox(token)
-        if len(inbox) == 1 and inbox.accesses > 0:
+        if len(inbox) == 1:
```
fkantelberg: I don't know about this change here. I think without access it should look like the link is invalid. To we want to be verbose here?

fkantelberg (changes requested):

vault/views/templates.xml
jans23: I added an appropriate error message shown to the user.

vault/controllers/main.py
jans23: restored as suggested.

PR #989, opened 2026-08-24
[18.0][FIX] vault: do not present exhausted inboxes as writable

jans23:
An inbox that can no longer be written to (access counter at 0 or expired) still showed its writable inbox link, and submitting to it silently did nothing while the form reported 'Successfully stored'. This is misleading for received/delivered secrets which are created with accesses = 0.
    
- Add a computed 'writable' flag (accesses left and not expired) and hide the inbox link in the list and form views when it is not writable.
- Make the controller report 'This link is no longer active' when store_in_inbox performs no write instead of a false success message.

This depends on #987

### DIFF
--- a/vault/views/vault_inbox_views.xml
+++ b/vault/views/vault_inbox_views.xml
@@ -3,7 +3,8 @@
     <record id="view_vault_inbox_tree" model="ir.ui.view">
         <field name="model">vault.inbox</field>
         <field name="arch" type="xml">
-            <list create="false">
+            <list create="false" decoration-muted="not writable">
+                <field name="writable" column_invisible="1" />
                 <field name="name" />
                 <field name="inbox_link" widget="url" />
             </list>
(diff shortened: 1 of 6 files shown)


vault/views/vault_inbox_views.xml
```diff
@@ -19,8 +20,13 @@
...
                     <field name="key" invisible="1" />
                     <field name="filename" invisible="1" />
+                    <field name="writable" invisible="1" />
                     <group>
-                        <field name="inbox_link" widget="url" />
+                        <field
+                            name="inbox_link"
+                            widget="url"
+                            invisible="not writable"
+                        />
```
fkantelberg: You can increase the access again or set a different expiration to make it writable again. Highlighting it that it's not writable is good but maybe better done with a ribbon.

```xml
<widget name="web_ribbon" title="Readonly" bg_color="text-bg-danger" invisible="not writable"/>
```

For the list view it might be better to use a decoration.

fkantelberg (changes requested):
This includes additional changes from another PR. Still some notes.

vault/views/vault_inbox_views.xml
jans23: changed accordingly.

PR #992, opened 2026-08-24
[18.0][FIX] vault_share: make share token consistent between UI and DB

jans23:
The server stored a different token than the one the form used to build the share link, which then always resulted in 'Invalid token'. Generate the token client-side via a string default and persist that exact value using a view readonly field with force_save, so the UI and DB always agree.

### DIFF
--- a/vault_share/models/vault_share.py
+++ b/vault_share/models/vault_share.py
@@ -22,7 +22,9 @@ class VaultShare(models.Model):
         store=False,
         help="Using this link and pin people can access the secret.",
     )
-    token = fields.Char(readonly=True, required=True, default=lambda self: uuid4())
+    token = fields.Char(
+        readonly=True, required=True, copy=False, default=lambda self: str(uuid4())
+    )
     secret = fields.Char()
     secret_file = fields.Char()
     filename = fields.Char()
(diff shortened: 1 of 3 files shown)


fkantelberg (approved):

vault_share/models/vault_share.py
```diff
@@ -22,7 +22,7 @@ class VaultShare(models.Model):
         store=False,
         help="Using this link and pin people can access the secret.",
     )
-    token = fields.Char(readonly=True, required=True, default=lambda self: uuid4())
+    token = fields.Char(required=True, copy=False, default=lambda self: str(uuid4()))
```
pedrobaeza: This should be still readonly.

vault_share/models/vault_share.py
```diff
@@ -73,6 +75,10 @@ def get(self, token, ip=None):
 
     @api.model_create_multi
     def create(self, vals_list):
+        vals_list = [
```
pedrobaeza: This safeguard shouldn't be applied, as the default assures the value.

vault_share/views/vault_share_views.xml
```diff
@@ -16,7 +16,7 @@
             <form>
                 <sheet>
                     <group>
-                        <field name="token" invisible="1" />
+                        <field name="token" invisible="1" readonly="1" force_save="1" />
```
pedrobaeza: In this version, you can just remove the field AFAIK.

vault_share/models/vault_share.py
jans23: Changed as suggested.

vault_share/models/vault_share.py
jans23: Changed as suggested.

vault_share/views/vault_share_views.xml
jans23: Changed as suggested.

pedrobaeza (approved):
/ocabot merge patch

PR #996, opened 2026-09-01
[19.0][MIG] base_user_show_email: Migration to 19.0

jcadhoc:
Supersedes #867 (same branch, plus one fix commit), which the stale bot closed
after 4 months without review. Migration work is @augusto-weiss'; this PR only
adds a fix on top and reopens the discussion.

## Migration

Standard migration of `base_user_show_email` to 19.0: manifest version bump and
removal of the `res_users_view.xml` form-view tweaks that core now covers.

## Fix on top

On the login page the renamed label and the core user switcher button render
glued together as `LoginChoose a user`.

Core renders the label as:

```xml
<label for="login" class="form-label d-flex justify-content-between">Email</label>
```

and the user switcher portals its "Choose a user" button straight into that
element — `addons/web/static/src/core/user_switch/user_switch.xml`,
`t-portal="'label.form-label'"`. This module replaces the whole label to rename
it to "Login" and dropped `d-flex justify-content-between` along the way, so the
portalled button loses its spacing. The password label right below, untouched by
this module, keeps rendering correctly.

The switcher only appears once the browser has stored logins for the database,
so the glitch is only visible to returning users.

Same fix for 18.0, where the module is already merged: #995

## Test plan

1. Install `base_user_show_email`.
2. Log in once so the user switcher stores the login, then log out.
3. Reload `/web/login`: the "Choose a user" button should sit at the right end
   of the label row, aligned the same way as "Reset password" is on the
   password label.

Internal ref: https://www.adhoc.inc/odoo/project.task/71192


base_user_show_email/README.rst
```diff
@@ -38,11 +38,6 @@ confusing.
 The goal of this module is to untangle (a bit) these two fields. Here
 are the changes:
 
--  In the res.users form view:
-
-   -  Display a "Login" label instead of "Email" on the login field
-   -  Show the currently invisible email field and its label
-
 -  In the res.users tree view:
```
kalpeshgajera-uncannycs, line 41: `README.rst` is auto-generated and should not be edited manually. 

The source of truth is `readme/DESCRIPTION.md` (which still contains these lines). As soon as the OCA bot or `pre-commit` runs, `README.rst` will be regenerated and these manual edits will be overwritten.
Please update `readme/DESCRIPTION.md` directly, then run `pre-commit run oca-gen-addon-readme -a` to regenerate `README.rst` (this will also update the branch badges from `18.0` to `19.0`).

kalpeshgajera-uncannycs (changes requested):

PR #1001, opened 2026-09-07
[17.0][FIX] password_security: Fix user login check in test_02_signup_user_success

Andrii9090-tecnativa:
This fix adapts the test to check that the user is logged in without relying on a specific redirect path, because in some cases, for example when the `portal` module installed, the redirect URL changes to `/my`.

@pedrobaeza @adasatorres-tecnativa can you review please

@Tecnativa


pedrobaeza (approved):

adasatorres-tecnativa (approved):
LGTM, code review

PR #1007, opened 2026-09-15
[18.0][FIX] auth_signup_verify_email: verify captcha on signup

### LINKED ISSUE #878: auth_signup_verify_email and recaptcha
**Describe the bug**
When the Google ReCaptcha feature (standard Odoo website configuration) is enabled, the passwordless signup flow provided by auth_signup_verify_email fails.

The controller passwordless_signup passes the entire values dictionary (which contains the form data, including recaptcha_token_response) directly to sudo_users.signup. Because recaptcha_token_response is not a valid field on the res.users model, the Odoo ORM raises a ValueError, resulting in a 500 Internal Server Error.

**To Reproduce**
Affected versions: Odoo 18 (possibly all others)

Install auth_signup_verify_email.

Enable "reCAPTCHA" in General Settings (requires valid Site Key and Secret Key).

Go to the website signup page.

Attempt to sign up using the flow provided by this module.

Check the server logs.

**Expected behavior** 
The signup process should complete successfully. The module should either filter out the recaptcha_token_response from the values dictionary before passing it to the signup method, or handle the ReCaptcha verification separately before user creation.

**Log / Traceback**

Python

File "/opt/odoo/parts/oca/server-auth/auth_signup_verify_email/controllers/main.py", line 56, in passwordless_signup
    sudo_users.signup(values, qcontext.get("token"))
  File "/opt/odoo/parts/odoo/odoo/addons/auth_signup/models/res_users.py", line 101, in signup
    self._signup_create_user(values)
  File "/opt/odoo/parts/odoo/odoo/addons/website/models/res_users.py", line 62, in _signup_create_user
    new_user = super(ResUsers, self)._signup_create_user(values)
  File "/opt/odoo/parts/odoo/odoo/addons/auth_signup/models/res_users.py", line 117, in _signup_create_user
    return self._create_user_from_template(values)
  ...
  File "/opt/odoo/parts/odoo/odoo/addons/auth_signup/models/res_users.py", line 156, in _create_user_from_template
    return super()._create_user_from_template(values)
  ...
ValueError: Invalid field 'recaptcha_token_response' on model 'res.users'
Additional context The issue is located in controllers/main.py. The values dictionary captures all params from the request. When ReCaptcha is active, the frontend sends recaptcha_token_response.

szalatyzuzanna:
Port of #933 (17.0) to 18.0, with a stronger test. Original fix by @Cl0ut1eR, kept as the commit author.

### Problem

`auth_signup` verifies the captcha token in `web_auth_signup()`, before `do_signup()` creates anything. This module overrides that route and, when a login arrives with no password, answers it with `passwordless_signup()` without calling `super()`, so that verification is never reached.

The result is that with `google_recaptcha` or `website_cf_turnstile` installed and configured, the widget is rendered on the signup page but the token is never checked on the server for this flow. We are getting automated signups in production because of it, and they are not harmless: each one triggers an activation email from our server to an address the bot chose, which is what damages the sending reputation of the domain.

There is a second effect, reported in #878: the captcha token stays in `request.params` and reaches `res.users.signup()`, where it raises an invalid field error for legitimate users.

### Fix

Call `ir.http._verify_request_recaptcha_token("signup")` at the start of the passwordless flow, so the existing error handling renders the message.

**No new dependency is added.** This answers the question raised in #783 ("Isn't reCaptcha an extra module?"), which was never followed up: `_verify_request_recaptcha_token` is defined in `base` (`ir.http`) and returns `True` when no captcha module is installed, so a database without one behaves exactly as it does today. `google_recaptcha` and `website_cf_turnstile` override it, and both pop their own token from `request.params` — which is also what fixes #878.

### About the test

`test_failed_recaptcha` patches the method on the registry class (`type(self.env["ir.http"])`) rather than on `base`'s `IrHttp`. This matters: when a captcha module is installed, its own override sits in front of the one in `base`, so patching `base` leaves the real implementation running and the test does not exercise the rejection at all. Patching `base` was verified to fail on a database with `website_cf_turnstile` installed — precisely the setup the fix is meant to protect.

The test also asserts that no user is created, which is the property that actually matters here, rather than only that an error is rendered.

### Related

#774, #783, #878, #901, #933


jelenapoblet (approved):
The issue and the proposed fix make sense to me.

oliverg09 (approved):
LGTM!

jelenapoblet:
@pedrobaeza I think this is a good fix, If you're ok with it, I would merge it

pedrobaeza (commented):
OK, but the problem may be using the new method `passwordless_signup` that replaces the whole logic of `web_auth_signup` instead of reusing it and only injecting the needed bits. Isn't possible to inject the verification inside the existing method and calling super afterwards for doing the rest of the standard setup?

And about the test, being so synthetic that is mocking the direct method, I don't think it's adding value.

szalatyzuzanna:
@pedrobaeza Thanks for the review.

I tried it: reusing `web_auth_signup()` and only overriding `_signup_with_values()` to skip the session authentication makes `test_good_email` fail. The user ends up on the login page with a "Wrong login/password" error instead of the "check your email" screen, since the account is created without a password and the standard flow logs in right
after. So it's a behaviour change, not just a super call. Worth doing, but as this is a security fix I'd rather leave the refactor to its own PR. Tell me if you'd rather have it all here.

On the test, fair point. It's only a regression guard so the call can't be dropped again. Happy to remove it if you think it's not worth it.

pedrobaeza (commented):
OK, I was talking about the method `web_auth_signup`, if you see it not possible, I don't insist. About the regression test, I think it's not adding value in this case. A real regression tests would be a complete signup flow and only mocking the captcha failure.

szalatyzuzanna:
You're right. Testing the real verification would need a captcha module as a dependency, which this PR avoids. Happy to drop the test if you prefer, it would just leave the `raise` line uncovered.

jelenapoblet:
@sbidoul there's an issue with the tests here, and it appears to affect the entire v18 branch of this repository too

